PR #116, opened 2017-03-14
i2c: Add DevReg.

maruel:
- Make endianness explicit when accessing an I²C device as memory mapped
  registers. This is more general and will work with devices having special
  needs.
- Add support for 32 bits and 64 bits.
- Add support for generic struct.
- Increase Register() correctness enforcement.
- Increase coverage to 100%


maruel:
Actually it cannot be committed as-is, because it hardcode the register address to be 8 bits and I have a use case of 16 bits, so I need to *at least* change the name to be DevReg8 or something similar.

conn/i2c/i2c.go
```diff
@@ -83,39 +85,123 @@ func (d *Dev) Write(b []byte) (int, error) {
...
+
+// ReadRegRaw writes the register number to the I²C bus, then reads data
+// unmarshalled.
+func (d *DevReg) ReadRegRaw(reg byte, b []byte) error {
 	return d.Tx([]byte{reg}, b)
 }
 
-// Shortcuts
+// ReadStruct writes the register number to the I²C bus, then reads data
+// marshalled.
```
tve: It would be nice to have more comments here. What is the intent? Size of struct = number of bytes read. Mention that binary.Read is used to read into struct and that it applies the byte ordering.

Perhaps an example could really help too!

conn/i2c/i2c.go
```diff
@@ -83,39 +85,123 @@ func (d *Dev) Write(b []byte) (int, error) {
...
+//
+// It uses reflection underneath.
+//
+// For slice of bytes, use ReadRegRaw(), for integers, use ReadRegUnitNN().
+func (d *DevReg) ReadRegStruct(reg byte, b interface{}) error {
```
tve: Why is this limited to a struct? binary.Read will happily read into other data structures... Or is it in order not to reimplement the binary.dataSize() function?

conn/i2c/i2c.go
```diff
@@ -83,39 +85,123 @@ func (d *Dev) Write(b []byte) (int, error) {
...
+// WriteRegUint64 writes a 64 bit register.
+func (d *DevReg) WriteRegUint64(reg byte, v uint64) error {
+	if d.Order == nil {
+		return errors.New("don't know if big or little endian")
+	}
+	var a [9]byte
+	a[0] = reg
+	d.Order.PutUint64(a[1:], v)
+	return d.Tx(a[:], nil)
 }
```
tve: WriteRegStruct could be implemented and could be useful to init a bunch of regs. But perhaps not worth it right now...

conn/i2c/i2c_test.go
```diff
@@ -36,12 +39,305 @@ func ExampleDev() {
...
+	}
+	if v, err := d.ReadRegUint64(34); err == nil || v != 0 {
+		t.Fatalf("%v, %v", v, err)
+	}
+}
+
+func TestDevReg_ReadRegUint(t *testing.T) {
+	r := []byte{0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f, 0x10}
+	b := &fakeBus{r: r}
+	d := DevReg{Dev: Dev{b, 12}, Order: binary.LittleEndian}
```
tve: Wouldn't cost anything to use BigEndian here for a change...

conn/i2c/i2c.go
maruel: Done.

conn/i2c/i2c_test.go
maruel: Done.

conn/i2c/i2c.go
maruel: Made is accept slices of int, slices of struct, struct with arrays, struct with structs, etc.

conn/i2c/i2c.go
maruel: Added an example for DevReg8.ReadRegStruct and DevReg8.WriteRegStruct.

maruel:
Added full support for I²C devices with a 16 bits address space. Since it doesn't make sense to switch between 8 bits and 16 bits address spaces, they are separate structs. This ensures compile time correctness. Please review.

PR #121, opened 2017-03-27
onewire: refactor factory alike i2c and spi

maruel:
- onewire: Redo the factory to be like i2c
- onewire: increase test coverage to 97%
- onewiretest: Add Playback.Q()
- onewiretest: increase test coverage to 100%


maruel:
I'll commit right away because there's increasing visitors on the project and this is a breaking change and I use the occasion to make it more coherent with the rest. Please still comment once you have time.

PR #123, opened 2017-04-01
spi: Update spi.Conn to separate app and device configuration.

maruel:
- Allow both app and device drive set max bus speed.
- Make DevParams callable only exactly once.
- sysfs: make SPI implement io.Reader.


maruel:
I'm going to commit; the fact that a SPI can only be configured exactly once is a design I wanted for a very long time. I think this PR is a good trade off vs the other design options.

PR #128, opened 2017-04-19
rsync compress

jpillora:
Go binaries are highly compressible. This greatly speeds transfers on slow networks.

### DIFF
--- a/cmd/push.sh
+++ b/cmd/push.sh
@@ -24,7 +24,7 @@ for i; do
   time GOOS=linux GOARCH=arm go build .
   echo ""
   echo -n "- Copying:  "
-  time rsync -v "$NAME" "$HOST:${NAME}"
+  time rsync -v --compress "$NAME" "$HOST:${NAME}"
   echo ""
   rm "$NAME"
   cd ..


maruel:
The common case for `push.sh` is for both to be on the same local network so in general bandwidth is not too much of a problem. On the other hand single core systems like BeagleBone and CHIP may be significantly slowed down by the compression.

Here's a few knobs that can impact performance:

- with/without `Compression yes` in `~/.ssh/config`
- with/without `rsync --compress`
- hot/warm/cold push
- RPi3 over ethernet (fast case) vs CHIP/BeagleBone over wifi (slow case, still pretty fast)

For cold cache:

- BeagleBone over ethernet, I saw a 1.2s->800ms reduction with `--compress`
- RPi3 over ethernet, I saw 1.1s->650ms
- CHIP over wifi, I saw 1.9s->1.45s
- CHIP over wifi with `Compression yes` in `.ssh/config`, I get 1.45s, with both 1.4s

For hot cache:
- CHIP, it doesn't have much effect at ~1.25s

My main concern was having it to be slower but it doesn't seem to be the case so lgtm.

Note that I have a vague desire to convert this script to Go so that Windows users could leverage it too but that's outside the scope of this change.

```
#!/bin/bash
set -eu
NAME=periph-smoketest.out
go build -o $NAME ./periph-smoketest
HOST=$1
for i in {0..5}; do
  time rsync -v --compress "$NAME" "$HOST:${NAME}"
  # Comment out for hot; we'd need to rebuild with a difference to test for warm
  ssh $HOST "rm $NAME" > /dev/null
done
rm $NAME
```

jpillora:
Cheers! In general, I find network IO vastly outweighs gzip IO. The only time I'd avoid it would be for files already compressed.

deadprogram:
I have intentions of adding this capability to Gort, so I am interested in this code. :) 

My own previous attempts at this use scp instead of rsync, not quite as slick:
https://github.com/hybridgroup/gobot-workshop-2017/blob/master/airlock/edison/runner.sh
and
https://github.com/hybridgroup/gobot-workshop-2017/blob/master/airlock/edison/runner.cmd

On related note, part of the intention for adding to Gort is to allow Windows devs to do the same thing. So we're all on the same wavelength.

maruel:
[OT]
I think [runner.cmd](https://github.com/hybridgroup/gobot-workshop-2017/blob/master/airlock/edison/runner.cmd) has a bug: it doesn't specify [setlocal](https://technet.microsoft.com/library/cc772046.aspx) so the variables leak through after the process run. I.e. "GOARCH and GOOS" remain set after running the script. I don't have access to a Windows box right now so I can't confirm.

deadprogram:
I doubt anyone tested that, so you are very likely correct.

PR #131, opened 2017-04-23
spi: Refactor Conn and ConnCloser into Conn, Port and PortCloser

maruel:
This clarifies the difference on ownership and makes it clear that DevParams()
is essentially a 'connection factory'.

Doing this caught inconsistencies in the unit tests. Made the structs in spitest
enforce the fact that DevParams() can only be called once.

### DIFF
--- a/conn/spi/spi.go
+++ b/conn/spi/spi.go
@@ -103,6 +103,21 @@ type Packet struct {
 // io.Reader.
 type Conn interface {
 	conn.Conn
+	// TxPackets does multiple operations over the SPI connection.
+	//
+	// The maximum number of bytes can be limited depending on the driver. Query
+	// conn.Limits.MaxTxSize() can be used to determine the limit.
+	//
+	// If the last packet has KeepCS:true, the behavior is undefined. The CS line
+	// will likely not stay asserted. This is a driver limitation.
+	TxPackets(p []Packet) error
+}
+
+// Port is the interface to be provided to device drivers.
+//
+// The device driver, that is the driver for the peripheral connected over
+// this port, calls DevParams() to retrieve a configured connection as Conn.
+type Port interface {
 	// DevParams sets the communication parameters of the connection for use by a
 	// device.
 	//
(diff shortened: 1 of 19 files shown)


maruel:
Hi I'd like to know your impressions on this, as I feel this would clarify the separation of "spi connection before it is configured" and "configured spi connection" as an atomic operation by calling `DevParams()`.

The device drivers now essentially accept a "connection factory" `spi.ConnConfig`. I think it's makes everything cleaner and more maintainable. wdyt?

conn/spi/spi.go
```diff
@@ -114,23 +128,15 @@ type Conn interface {
...
-	// TxPackets does multiple operations over the SPI bus.
-	//
-	// The maximum number of bytes can be limited depending on the driver. Query
-	// conn.Limits.MaxTxSize() can be used to determine the limit.
-	//
-	// If the last packet has KeepCS:true, the behavior is undefined. The CS line
-	// will likely not stay asserted. This is a driver limitation.
-	TxPackets(p []Packet) error
+	DevParams(maxHz int64, mode Mode, bits int) (Conn, error)
 }
```
tve: Have you considered calling `ConnConfig` a `Bus`, and calling `DevParams` either `Connect` or `Device`? You are opening the bus and you connect to a specific device on the bus by proving the params. Mhh, The CnnConfig already has the chip-select built-in, why? I'm probably missing something :-).

conn/spi/spi.go
```diff
@@ -103,6 +103,20 @@ type Packet struct {
...
+	// conn.Limits.MaxTxSize() can be used to determine the limit.
+	//
+	// If the last packet has KeepCS:true, the behavior is undefined. The CS line
+	// will likely not stay asserted. This is a driver limitation.
+	TxPackets(p []Packet) error
+}
+
+// ConnConfig is the interface to be provided to device drivers.
+//
+// The driver calls DevParams() to retrieve a configured connection.
```
maruel: By "driver", I meant "device driver", but really, it is the "consumer" of the connection, the slave end. It'd be good to find a word that carries the right semantic and could then be used thoroughly in the code base.

conn/spi/spi.go
maruel: I did.

SPI without CS is a bus, SPI with CS is a point-to-point connection.

I implemented a version with a Bus but it felt more clunky to use in practice... It was annoying for the app and the app had to eventually specify a CS anyway, it's not the job of the device driver to chose CS0 or CS1.

I agree DevParams is probably ill-named but it still carries the important meaning about the fact that it is about the communication parameters for the device. Maybe CommParams :/

tve: :+1:

conn/spi/spi.go
tve: How do you like "peripheral driver"?

maruel: I'd like to make this less ambiguous, I send PR #133 to experiment with this.

maruel:
I had an epiphany and realized that `Port` was the right term to use for an unconfigured `Conn`.

So this leads to `ì2c.Bus` and `i2c.BusCloser` but to `spi.PortCloser`, `spi.Port` then `spi.Conn`. As noted above, this wouldn't preclude from `spi.Bus` but from trying it out, it didn't give any benefit while making app code harder to use.

PR #132, opened 2017-04-23
Add driver for Bosch BMP180 temp/pressure sensor

### LINKED ISSUE #130: Driver for Bosch BMP180 (i2c temperature+pressure sensor)
Hi,

thanks a lot for this project! I've just switched over from https://github.com/kidoman/embd and like it very much. I have a couple of [BMP180](https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BMP180-DS000-121.pdf) temperature and pressure sensors lying around, and I'm just implementing a driver for periph. I'll do a PR later this weekend.

fd0:
This PR adds a driver for the Bosch BMP180 temperature/pressure sensor. At the moment the driver works and reads sensible data.

Closes #130


devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
```
quinte17: is here 2017 preferable?

cmd/bmp180/main.go
```diff
@@ -0,0 +1,119 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
```
quinte17: is here 2017 preferable?

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+	d.c.AC3 = int16(cal[4])<<8 | int16(cal[5])
+	d.c.AC4 = uint16(cal[6])<<8 | uint16(cal[7])
+	d.c.AC5 = uint16(cal[8])<<8 | uint16(cal[9])
+	d.c.AC6 = uint16(cal[10])<<8 | uint16(cal[11])
+	d.c.B1 = int16(cal[12])<<8 | int16(cal[13])
+	d.c.B2 = int16(cal[14])<<8 | int16(cal[15])
+	d.c.MB = int16(cal[16])<<8 | int16(cal[17])
+	d.c.MC = int16(cal[18])<<8 | int16(cal[19])
+	d.c.MD = int16(cal[20])<<8 | int16(cal[21])
+
```
quinte17: please consider something like
`convert(cal[:], &d.c)`
where
```
import "encoding/binary"
import "bytes"

func convert(b []byte, data interface{}) error {
	buf := bytes.NewReader(b)
	return binary.Read(buf, binary.LittleEndian, data)
}
```

cmd/bmp180/main.go
```diff
@@ -0,0 +1,119 @@
...
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// bmp180 reads the current temperature and pressure from a BMP180.
```
maruel: /me start thinking about more coherency for package docstring
https://godoc.org/periph.io/x/periph/devices#pkg-subdirectories

At the very least, it shall start with: `Package bmp180 ...`

Include a Reference section with a link to the datasheet.
https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BMP180-DS000-121.pdf

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+	env.Pressure = devices.KPascal(pressure)
+	return nil
+}
+
+// Halt is a noop for the BMP180.
+func (d *Dev) Halt() error {
+	return nil
+}
+
+// Opts is optional options to pass to the constructor.
```
maruel: Since there's only one parameter, pass the parameter to New() directly. No need for a struct.

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+	regSoftReset        = 0xE0 // soft reset register
+	regCtrlMeas         = 0xF4 // control measurement
+	regOutMSB           = 0xF6 // MSB of measurement
+	regOutLSB           = 0xF7 // LSB of measurement
+	regOutXLSB          = 0xF8 // extended LSB of measurement
+)
+
+const softResetValue = 0xB6
+
+func (d *Dev) makeDev(opts *Opts) error {
```
maruel: Since there's no SPI support (the datasheet says it's on custom orders only, so not worth implementing), then there's no need of a makeDev() function.

cmd/bmp180/main.go
```diff
@@ -0,0 +1,119 @@
...
+	for {
+		var env devices.Environment
+		if err := e.Sense(&env); err != nil {
+			return err
+		}
+		fmt.Printf("%8s %10s %9s\n", env.Temperature, env.Pressure, env.Humidity)
+		if !loop {
+			break
+		}
+		time.Sleep(100 * time.Millisecond)
```
maruel: Use the conversion time from the datasheet at page 21. 100ms is waaay too long.

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+	Pressure Oversampling
+}
+
+const (
+	bmp180Address = 0x77 // the address of a BMP180 is fixed at 0x77
+	chipID        = 0x55 // contents of the ID register, always 0x55
+)
+
+// New returns an object that communicates over I²C to BMP180 environmental
+// sensor.
```
maruel: Add a note that the device can work at up to 3.4MHz. This is valuable information.

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+	return nil
+}
+
+// Opts is optional options to pass to the constructor.
+type Opts struct {
+	Pressure Oversampling
+}
+
+const (
+	bmp180Address = 0x77 // the address of a BMP180 is fixed at 0x77
```
maruel: I'm not a huge fan of constants that are used exactly once.

devices/bmp180/bmp180_test.go
```diff
@@ -0,0 +1,29 @@
+package bmp180
```
maruel: add standard header.

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+
+	time.Sleep(100 * time.Millisecond)
+
+	// read value
+	var tempBuf [2]byte
+	if err := d.readReg(regOutMSB, tempBuf[:]); err != nil {
+		return err
+	}
+
+	UT := uint16(tempBuf[0])<<8 | uint16(tempBuf[1])
```
maruel: why uppercase?

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+	// start conversion for pressure
+	cmd = 1 << ctrlStartConversion
+	cmd |= ctrlCommandPressure << ctrlMeasurementControl
+	cmd |= byte(d.os) << ctrlOversampling
+
+	if err := d.d.Tx([]byte{regCtrlMeas, cmd}, nil); err != nil {
+		return err
+	}
+
+	time.Sleep(100 * time.Millisecond)
```
maruel: too long

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+	d.c.B1 = int16(cal[12])<<8 | int16(cal[13])
+	d.c.B2 = int16(cal[14])<<8 | int16(cal[15])
+	d.c.MB = int16(cal[16])<<8 | int16(cal[17])
+	d.c.MC = int16(cal[18])<<8 | int16(cal[19])
+	d.c.MD = int16(cal[20])<<8 | int16(cal[21])
+
+	return nil
+}
+
+func (d *Dev) readReg(reg uint8, b []byte) error {
```
maruel: You may want to use periph.io/x/periph/conn/mmr which exists exactly for this.

The bme280 driver couldn't use it because of the SPI weirdo communication protocol.

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+	return d.d.Tx([]byte{reg}, b)
+}
+
+// calibration data read from the internal EEPROM (datasheet page 13)
+type calibration struct {
+	AC1, AC2, AC3 int16
```
maruel: why export the members? Please use lowercase even if they are uppercase in the datasheet.

maruel:
gohci

maruel:
(I need to fix the gohci check)

maruel:
gohci: test

(sorry I'm using your PR as a test bed to enable gohci jobs on PRs)

devices/bmp180/bmp180.go
fd0: Sure.

devices/bmp180/bmp180.go
fd0: Hm, I disagree here:
 * Is that really necessary? I've use the same style as in the `bme280` driver, which uses exactly this here: https://github.com/google/periph/blob/master/devices/bme280/bme280.go#L257-L269
 *  It feels wrong to me. What's happening here is that the bytes are interpreted as big endian (not little endian) according to the datasheet. Having a separate function call, that allocates a bytes.Reader struct, only to read two bytes out of it feels like A LOT overhead.
 * IMHO it makes the code much harder to read.

cmd/bmp180/main.go
fd0: I think so.

cmd/bmp180/main.go
fd0: Uhm, that's not package `bmp180`, but package `main`. The package `bmp180` has a proper doc string and a link to the datasheet:
```
// Package bmp180 controls a Bosch BMP180 device over I²C.
//
// Datasheet
//
// https://cdn-shop.adafruit.com/datasheets/BST-BMP180-DS000-09.pdf
```

The datasheet you linked is really bad btw, the font in the algorithm on page 15 is unreadable. The one at adafruit that I linked has at least a readable font in the screenshot...

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+
+// Possible oversampling values.
+const (
+	No  Oversampling = 0
+	O2x Oversampling = 1
+	O4x Oversampling = 2
+	O8x Oversampling = 3
+)
+
+// Dev is an handle to a bmp180.
```
tve: s/an/a/

devices/bmp180/bmp180.go
fd0: I used the syntax from the manual.

cmd/bmp180/main.go
fd0: Ok

devices/bmp180/bmp180.go
fd0: Agreed.

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+
+// bit offsets in regCtrlMeas
+const (
+	ctrlMeasurementControl = 0
+	ctrlStartConversion    = 5
+	ctrlOversampling       = 6
+)
+
+// Sense returns measurements as °C and kPa.
+func (d *Dev) Sense(env *devices.Environment) error {
```
tve: @maruel why doesn't sense return devices.Environment?

devices/bmp180/bmp180.go
fd0: Ok.

cmd/bmp180/main.go
tve: Since this is in a loop printing to stdout maybe it should be something slower that one can follow? Such as 2x per second?

devices/bmp180/bmp180.go
fd0: I disagree: IMHO assigning a name helps readability a lot. This gives a magic constant that is suddenly used in the code a name, which can e.g. be searched in the datasheet. This is a common practice in development for embedded systems (at least in my experience). Thoughts?

devices/bmp180/bmp180.go
fd0: OK.

devices/bmp180/bmp180.go
fd0: Ok.

devices/bmp180/bmp180.go
tve: To me the `convert` idea is pretty slick. It could actually be a method on `calibration` that takes the byte array as arg and returns a `calibration` struct, mhh, could actually be newCalibration. You're creating a binary.BigEndian reader for the entire struct, not for each 2-byte element. Or am I not getting it right?

maruel: bme280 uses custom because of weirdo bit packing at nibble boundary for the "h" ones.

https://periph.io/x/periph/conn/mmr#Dev8.ReadStruct is what you want.

devices/bmp180/bmp180.go
maruel: You are in package bmp180, so prefixing a const with "bmp180" is redundant.

It was more of a nit than anything else, I don't mind much.

devices/bmp180/bmp180.go
maruel: in case the struct grows a lot to save on allocation, I agree it could be changed, I'm not confident with this interface.

cmd/bmp180/main.go
maruel: Oh right, for printing to the user, you can make this configurable through https://golang.org/pkg/flag/#Duration with `time.Second` the default. Also you want once per second, not sleep one second, which means using https://golang.org/pkg/time/#After

(I haven't done thing in bme280/main.go, it'd be a good thing to fix there too)

cmd/bmp180/main.go
maruel: Oops sorry I was really confused, ignore my comment about package name.

It kind of suck that the official one is of less quality. I'm not a fan of linking to adafruit's copies.

devices/bmp180/bmp180.go
maruel: still, upper case has special meaning in Go so use lower case.

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+
+const (
+	ctrlCommandTemperature = 0x0E
+	ctrlCommandPressure    = 0x14
+)
+
+// bit offsets in regCtrlMeas
+const (
+	ctrlMeasurementControl = 0
+	ctrlStartConversion    = 5
```
maruel: These are bitshift, so their name should be ctrlMeasurementShift, ctrlStartConversionShift and ctrlOversampleShift. Otherwise the name is confusing.

devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,258 @@
...
+	c  calibration
+	os Oversampling
+}
+
+func (d *Dev) String() string {
+	return fmt.Sprintf("BMP180{%s}", d.d)
+}
+
+const (
+	ctrlCommandTemperature = 0x0E
```
maruel: Frankly it'd make more sense if the constants were fixed here, e.g.
```
ctrlCommandTemperature = 0x0E << ctrlMeasurementControl
```

but since these commands cannot run without the start conversion bit, I don't see value in having these constant separated.

devices/bmp180/bmp180.go
tve: given everything else, the allocation of the struct seems irrelevant to me...

devices/bmp180/bmp180.go
quinte17: the convert idea gets all your single lines into a oneliner..
ok maybe i missed the endianess.
example usage can be seen here:
https://github.com/quinte17/bme280/blob/master/bme280.go#L357
but i agree for bme280 its not perfect either..

fd0: @quinte17 Ah! I re-read your comment and I missed that you pass a pointer to the calibration struct to `convert()`, now it makes sense. Thanks!

devices/bmp180/bmp180.go
fd0: Hm, I'm not aware of any special meaning of the case of variables within a function, am I missing something?

I'll use lower case.

maruel: You are right there is no meaning inside a function, but the code style is to always use lower case names for locally scoped symbols.

cmd/bmp180/main.go
fd0: I agree. Maybe we can put both links in? I also found at least two errors in the official datasheet...

maruel: Sure!

BTW sorry for the feeling of nitpicking fest; I'm trying to keep a very specific style on the library and for the device drivers I really want to behave the best w.r.t. the device (e.g. like sleeping the exact time as specified by the datasheet). I hope it's not a bummer for you.

cmd/bmp180/main.go
fd0: Fine with me.

devices/bmp180/bmp180.go
fd0: Ok, I can live with that :)

What I meant was mainly tailored towards magic constants that have been sourced from somewhere else (e.g. the datasheet).

devices/bmp180/bmp180.go
fd0: Ah, thanks! I wasn't aware of the mmr package.

devices/bmp180/bmp180.go
fd0: Hm, I'm pretty sure nothing is exported here since the struct itself is not exported...

fd0: That doesn't work: Using reflection on not exported struct members fails, so we cannot use `binary.Read()` to parse the calibration data. I'm leaving this upper case.

Besides: The struct itself is not exported, so it's not as if it will show up somewhere, not even in godoc...

devices/bmp180/bmp180_test.go
fd0: "header" as in "package header" or "copyright information"?

devices/bmp180/bmp180.go
maruel: Oh sure, if (and only if) you switch to using reflection, then keep them exported.

The idea here is to always aim for least privilege. Uppercase/lowercase is a way to document to the _reader of the code_ if there are interactions that falls outside of the assignments as shown in the package.

devices/bmp180/bmp180_test.go
maruel: Copyright header.

devices/bmp180/bmp180.go
fd0: I agree, it's irrelevant. But I also think it is outside the scope of this PR to change the API :)

devices/bmp180/bmp180.go
fd0: Agreed.

devices/bmp180/bmp180.go
fd0: I've renamed the constants here to `cmdStartTempConv`, I think that makes more sense. Thanks for the suggestion.

fd0:
I've addressed all comments and moved the device to `experimental/devices` (and I've verified that the driver still works of course). Looking forward to further comments!

cmd/bmp180/main.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		<-t.C
+	}
+	return nil
+}
+
+func mainImpl() error {
+	i2cID := flag.String("i2c", "", "I²C bus to use")
+	hz := flag.Int("hz", 0, "I²C bus speed")
```
maruel: I'd say to not bother, I had added these because I was experimenting but it's really irrelevant for a sensing device, I'll post a change to remove it from the bme280 exec.

Also setspeed is not working fine on the rpi so it's not even usable at the time. :(

cmd/bmp180/main.go
```diff
@@ -0,0 +1,127 @@
...
+	sample2x := flag.Bool("s2", false, "sample at 2x")
+	sample4x := flag.Bool("s4", false, "sample at 4x")
+	sample8x := flag.Bool("s8", false, "sample at 8x")
+	loop := flag.Bool("l", false, "read data continously")
+	interval := flag.Duration("i", time.Second, "read data continously with this interval")
```
maruel: what about just having -i, make it default to 0?
then no need for a second flag. will update bme280 later.

cmd/bmp180/main.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	if *hz != 0 {
+		if err := bus.SetSpeed(int64(*hz)); err != nil {
+			return err
+		}
+	}
+
+	var dev *bmp180.Dev
+
+	fmt.Printf("open bmp180\n")
```
maruel: don't forget to remove

cmd/bmp180/main.go
```diff
@@ -0,0 +1,127 @@
...
+	if dev, err = bmp180.New(bus, os); err != nil {
```
maruel: extract the assignment from the if, so no need to define dev upfront

experimental/devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,259 @@
...
+	// issue soft reset to initialize device
+	if err := d.dev.WriteUint8(regSoftReset, softResetValue); err != nil {
+		return nil, err
+	}
+
+	// wait for restart
+	time.Sleep(10 * time.Millisecond)
+
+	// read calibration data from internal EEPROM, 11 registers with two bytes each
+	var cal [22]byte
```
maruel: cal := calibration {}

experimental/devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,259 @@
...
+	env.Pressure = devices.KPascal(pressure)
+	return nil
+}
+
+// Halt is a noop for the BMP180.
+func (d *Dev) Halt() error {
+	return nil
+}
+
+const (
```
maruel: group all consts groups without functions in between.

but: I generally prefer public symbols to be first, the. private .
then each group has const, var, the code. it's not an hard requirement

experimental/devices/bmp180/bmp180.go
fd0: But then we cannot do the check a few lines later easily, as the datasheet suggests...

maruel: The main reason I prefer this is because of the current code use reflection twice. It is mostly a bug in the API design of `mmr` but still. The fact that big endian decoding is done part manually, part automatically, is kind of weird.

```
func isValid(i int16) bool {
  return i != 0 && i != ^int16(0)
}
func isValidU(i uint16) bool {
  return i != 0 && i != 0xFFFF
}

func (c*calibration) isValid() bool {
  return isValid(c.AC1) && isValid(c.AC2) && isValid(c.AC3) && isValidU(c.AC4) && isValidU(c.AC5) && isValidU(c.AC6) && isValid(c.B1) && isValid(c.B2) && isValid(c.MB) && isValid(c.MC) && isValid(c.MD)
}
```

quinte17: should the newCalibration func validate before returning a correct instance?

cmd/bmp180/main.go
fd0: OK

cmd/bmp180/main.go
fd0: Good idea.

cmd/bmp180/main.go
fd0: OK

experimental/devices/bmp180/bmp180.go
fd0: Nice solution, thanks. And removing the double endian decoding is much better code.

fd0:
Ok, more improvements, thanks!

cmd/bmp180/main.go
```diff
@@ -0,0 +1,110 @@
...
+	if interval != 0 {
+		t = time.NewTicker(interval)
+	}
+
+	for {
+		var env devices.Environment
+		if err := e.Sense(&env); err != nil {
+			return err
+		}
+		fmt.Printf("%8s %10s %9s\n", env.Temperature, env.Pressure, env.Humidity)
```
maruel: There's no humidity sensing so it should not be printed.

experimental/devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,265 @@
...
+	O4x Oversampling = 2
+	O8x Oversampling = 3
+)
+
+// bit offsets in regCtrlMeas
+const (
+	ctrlMeasurementControlShift = 0
+	ctrlStartConversionShift    = 5
+	ctrlOversamplingShift       = 6
+)
```
maruel: optional nit: use one const () section, with one empty line being the separation between logical groups.

experimental/devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,265 @@
...
+func isValidU(i uint16) bool {
+	return i != 0 && i != 0xFFFF
+}
+
+// valid checks whether the calibration data is valid.
+func (c *calibration) isValid() bool {
+	return isValid(c.AC1) && isValid(c.AC2) && isValid(c.AC3) && isValidU(c.AC4) && isValidU(c.AC5) && isValidU(c.AC6) && isValid(c.B1) && isValid(c.B2) && isValid(c.MB) && isValid(c.MC) && isValid(c.MD)
+}
+
+// newCalibration parses calibration data from the buf and checks that it is valid.
```
maruel: remove now dead code

experimental/devices/bmp180/bmp180_test.go
```diff
@@ -0,0 +1,33 @@
...
+		AC5: 32757,
+		AC6: 23153,
+		B1:  6190,
+		B2:  4,
+		MB:  -32768,
+		MC:  -8711,
+		MD:  2868,
+	}
+
+	temp := c.compensateTemp(27898)
```
maruel: merge lines 24-25 into one
merge lines 39-40 into one

experimental/devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,265 @@
...
+	chipID      = 0x55 // contents of the ID register, always 0x55
+)
+
+// registers
+const (
+	regChipID           = 0xD0 // register contains the chip id
+	regCalibrationStart = 0xAA // first calibration register address
+	regSoftReset        = 0xE0 // soft reset register
+	regCtrlMeas         = 0xF4 // control measurement
+	regOutMSB           = 0xF6 // MSB of measurement
```
maruel: why not just `regOut = 0xF6 // 3 bytes register with measurement data`
the rest is not used directly anyway.

experimental/devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,265 @@
...
+	id, err := d.dev.ReadUint8(regChipID)
+	if err != nil {
+		return nil, err
+	}
+
+	if id != chipID {
+		return nil, fmt.Errorf("bmp180: unexpected chip id 0x%x; is this a BMP180?", id)
+	}
+
+	// issue soft reset to initialize device
```
maruel: I don't see why this is needed? I'm fine with exposing `Reset()` as a method but I don't see the need to call it every time the device is instantiated.

experimental/devices/bmp180/bmp180.go
maruel: FTR, the reason is that I care about vertical density; the more dense the code can be vertically, the more a screenful page shows more code, leading to better visualization. Obviously this is a trade off with readability, as some people need more vertical spacing than others but I feel it is generally abused so I prefer not too many empty lines.

cmd/bmp180/main.go
fd0: OK

experimental/devices/bmp180/bmp180.go
fd0: Works for me.

experimental/devices/bmp180/bmp180.go
fd0: OK.

experimental/devices/bmp180/bmp180.go
fd0: You're right: It is not necessary (in most cases). I'd like to make sure that the device is in a known good state before interacting with it. We can move this to a `Reset()` method, though. What do you think?

fd0:
Let me know if there's anything else.

experimental/devices/bmp180/bmp180.go
maruel: Yes please. It's the last request after that I will merge the PR.

fd0: done.

experimental/devices/bmp180/bmp180.go
```diff
@@ -0,0 +1,247 @@
...
+var pressureConvTime = [...]time.Duration{
+	4500 * time.Microsecond,
+	7500 * time.Microsecond,
+	13500 * time.Microsecond,
+	25500 * time.Microsecond,
+}
+
+// Dev is a handle to a bmp180.
+type Dev struct {
+	dev *mmr.Dev8
```
maruel: I just realized you don't need it to be a pointer, please inline instead of allocating on the heap. Last comment I promise.

fd0: He, don't worry, when working with external contributors for [restic](https://github.com/restic/restic) (my project) I'm regularly on the other side of this.

PR #133, opened 2017-04-24
rename device to peripheral

maruel:
Unsure if it's a good idea

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ func main() {
 }
 ~~~
 
-Curious? Look at [supported devices](https://periph.io/device/) for more
+Curious? Look at [supported peripherals](https://periph.io/device/) for more
 examples!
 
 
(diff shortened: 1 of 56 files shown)


maruel:
The main challenge is to have consistent naming convention, so users can have a clearer picture.

Peripherals fits more with the project name "periph" and also is a clearer contrast with "host". On the other hand renaming the packages will break everyone but I'm willing to accept that, if I'm going to go ahead with this, the sooner the better.

maruel:
Then there's the challenge with `Dev` and `DevParams`. My take is to rename `DevParams` to `PeripheralParams` but not touch `Dev`. Very undecided about it.

tve:
I didn't look through the PR with a fine-toothed comb, but as far as I can tell it makes a lot of sense. It's odd at times, like calling the bme280 a peripheral. Or the one-wire lingo is "1-wire devices" and so it's odd to see "1-wire peripheral" instead.
In my mind the main pro is that it's clean and distinguishes linux devices vs. the, ahem, peripherals we actually want to program using this library. The main con is that perhaps it's too odd & pedantic.
Some additional opinions would be good, as well as perhaps looking at how the general docs would read.

maruel:
I'll close this PR, as I'm not sure it's worth.

PR #134, opened 2017-04-24
use a newCalibration func like in bmp180

quinte17:
i like the idea for the newCalibration func in the bmp180 driver.
it should be done for the bme too.

another point that could be adressed is in bme280_test.go.
we could reuse the newCalibration func for the real extracted data.

what do you think?


maruel:
It's a  good idea, thanks!

PR #136, opened 2017-05-20
Onewire using Linux kernel drivers

olliephillips:
Exposes Onewire using Linux kernel drivers. Also includes modifications to the `DS18B20.go` file to implement a constructor and struct methods to uses this device with the kernel drivers.

Usage along these lines:-

```
        ds, err := ds18b20.NewW1Therm()
	if err != nil {
		log.Fatal(err)
	}

	// List devices
	list, err := ds.List()
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(list)

	// Get temp every 5 seconds
	for {
		temp, err := ds.Temperature("28-000006099568")
		if err != nil {
			log.Fatal(err)
		}
		fmt.Println(temp)
		time.Sleep(time.Second * 5)
	}
```
I've picked up most of the Gitter feedback I think and there is a test file for the onewire.go file included.


devices/ds18b20/ds18b20.go
```diff
@@ -24,10 +24,13 @@ package ds18b20
 import (
 	"errors"
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
 	"periph.io/x/periph/conn/onewire"
 	"periph.io/x/periph/devices"
+	"periph.io/x/periph/host/sysfs"
```
maruel: Normally the packages in /devices should not depend on /host, but /conn instead. The current exception is /host/cpu for https://periph.io/x/periph/host/cpu#Nanospin

devices/ds18b20/ds18b20.go
```diff
@@ -175,3 +178,73 @@ func (d *Dev) readScratchpad() ([]byte, error) {
...
+	}
+	if len(list) == 0 {
+		return list, errors.New("ds18b20: could not get device list")
+	}
+	return list, nil
+}
+
+// Temperature returns temperature of requested DS18B20 probe in celsius
+func (dd *W1ThermDevice) Temperature(id string) (devices.Celsius, error) {
+	_, ok := dd.Probe[id]
```
maruel: p, ok

then use this variable at line 229

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+	modProbeCmd  string
+	thermMod     string
+	gpioMod      string
+	masterPrefix string
+	initialized  bool
+}
+
+// OneWireDevice represents a single OneWire device
+type OneWireDevice struct {
+	mtx sync.Mutex
```
maruel: convention is `mu`

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+func NewOneWire() (*oneWire, error) {
+	if isLinux {
+		return newOneWire()
+	}
+	return nil, errors.New("sysfs-onewire: not implemented on non-linux OSes")
+}
+
+func newOneWire() (*oneWire, error) {
+	ow := oneWire{
+		path:         "/sys/bus/w1/devices/",
```
maruel: why make them variable, would they ever change?

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+		thermMod:     "w1-therm",
+		gpioMod:      "w1-gpio",
+		masterPrefix: "w1_bus_master",
+		initialized:  false,
+	}
+
+	return &ow, nil
+}
+
+// LoadDrivers makes calls to kernel drivers w1_therm and w1_gpio
```
maruel: Doc must be much clearer about the side effects, that it shells out. For example when running under systemd as a service the sheer fact of fork()'ing out may be denied.

Calling this function twice will run modprobe twice.

There's no way to initialize the object *without* having modprobe being shelled out.

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// Set state, forcing this method to be called explicitly
+	ow.initialized = true
+	return nil
+}
+
+// Read returns the contents of a OneWire device file as a Reader
+// Assumption is that specific device abstractions will do what they
+// need to with the data
+func (owd *OneWireDevice) Read() (*bufio.Reader, error) {
```
maruel: use just `o` instead of `owd`

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	if _, err := owd.f.Seek(0, 0); err != nil {
+		return reading, fmt.Errorf("sysfs-onewire: %v", err)
+	}
+	reading = bufio.NewReader(owd.f)
+	return reading, nil
+}
+
+// Scan returns map of discovered OneWire devices, filtered by prefix if required
+func (ow *oneWire) Scan(prefix string) (map[string]*OneWireDevice, error) {
```
maruel: `o`

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Read returns the contents of a OneWire device file as a Reader
+// Assumption is that specific device abstractions will do what they
+// need to with the data
+func (owd *OneWireDevice) Read() (*bufio.Reader, error) {
+	var reading *bufio.Reader
+	owd.mtx.Lock()
+	defer owd.mtx.Unlock()
+	if owd.f == nil {
+		return reading, errors.New("sysfs-onewire: device file handle closed")
```
maruel: It's not a big deal to crash

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	if _, err := owd.f.Seek(0, 0); err != nil {
+		return reading, fmt.Errorf("sysfs-onewire: %v", err)
+	}
+	reading = bufio.NewReader(owd.f)
+	return reading, nil
```
maruel: I don't understand why you seek with the lock, but then follow up reading will be done without the lock.

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+		initialized:  false,
+	}
+
+	return &ow, nil
+}
+
+// LoadDrivers makes calls to kernel drivers w1_therm and w1_gpio
+func (ow *oneWire) LoadDrivers() error {
+	// Check modules available
+	mod := exec.Command(ow.modProbeCmd, ow.gpioMod)
```
maruel: modprobe function shall be a standalone function, not a method.

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+	if prefix != "" {
+		for i := range files {
+			if strings.HasPrefix(files[i].Name(), prefix) {
+				if (files[i].Mode() & os.ModeSymlink) == os.ModeSymlink {
+					f, err := os.Open(path.Join(path.Join(ow.path, files[i].Name()), "w1_slave"))
+					if err != nil {
+						return nil, fmt.Errorf("sysfs-onewire: %v", err)
+					}
+					device := &OneWireDevice{
+						mtx: sync.Mutex{},
```
maruel: Do not explicitly initialize mu, not needed.

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+						f:   f,
+					}
+					devices[files[i].Name()] = device
```
maruel: so in the end, it looks like:
`devices[name] = &OneWireDevice{f: f}`

one line instead of 5.

host/sysfs/onewire.go
```diff
@@ -0,0 +1,158 @@
...
+	if !ow.initialized {
+		return devices, fmt.Errorf("sysfs-onewire: Onewire kernel drivers must be loaded with LoadDrivers()")
+	}
+
+	files, err := ioutil.ReadDir(ow.path)
+	if err != nil {
+		return devices, fmt.Errorf("sysfs-onewire: %v", err)
+	}
+	if prefix != "" {
+		for i := range files {
```
maruel: for i, fi := range files {
  name := fi.Name()

so you don't continuously lookup a slice *and* call a method.

host/sysfs/onewire_test.go
```diff
@@ -0,0 +1,121 @@
...
+package sysfs
+
+import (
+	"io/ioutil"
+	"path"
+	"reflect"
+	"strconv"
+	"testing"
+)
+
```
maruel: please include an example, https://golang.org/pkg/testing/#hdr-Examples

devices/ds18b20/ds18b20.go
```diff
@@ -175,3 +178,73 @@ func (d *Dev) readScratchpad() ([]byte, error) {
...
+
+// NewW1Therm loads kernel drivers and scans for DSB1820 family devices only
+func NewW1Therm() (*W1ThermDevice, error) {
+	dd := W1ThermDevice{}
+	ow, err := sysfs.NewOneWire()
+	if err != nil {
+		return &dd, err
+	}
+	// Load kernel drivers
+	err = ow.LoadDrivers()
```
maruel: I'd prefer this function to accept an interface that permits it to scan without depending directly on sysfs. [Bus.Search()](https://godoc.org/periph.io/x/periph/conn/onewire#Bus) currently returns all the addresses found, instead of returning an interesting address, maybe the interface could be adapted to be more polyvalent? Paging @tve.

maruel:
Thanks for the change! Please to look at the googlebot comment about the CLA. Then you can add yourself to AUTHORS and CONTRIBUTORS.

olliephillips:
Thanks for comprehensive review and feedback, I'll revisit come back. Closing for now

PR #138, opened 2017-05-28
Set version of RPi for all known revisions

maxekman:
I needed to add support for the Raspberry Pi Zero W. I thought it would be wiser to specify revisions explicitly and I found a table [here](http://elinux.org/RPi_HardwareHistory) with an exhaustive list.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,3 +6,4 @@
 Google Inc.
 Stephan Sperber
 Thorsten von Eicken <tve@voneicken.com>
+Max Ekman <max@looplab.se>
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,6 @@
 # some cases, their employer may be the copyright holder.  To see the full list
 # of contributors, see the revision history in source control.
 Google Inc.
+Max Ekman <max@looplab.se>
 Stephan Sperber
 Thorsten von Eicken <tve@voneicken.com>
-Max Ekman <max@looplab.se>
--- a/host/rpi/rpi.go
+++ b/host/rpi/rpi.go
@@ -45,32 +45,32 @@ func Present() bool {
 // P1 is also known as J8 on A+, B+, 2 and later.
 var (
 	// Rapsberry Pi A and B, 26 pin header:
-	P1_1  = pin.V3_3       // max 30mA
-	P1_2  = pin.V5         // (filtered)
-	P1_3  = bcm283x.GPIO2  // High, I2C1_SDA
-	P1_4  = pin.V5         //
-	P1_5  = bcm283x.GPIO3  // High, I2C1_SCL
-	P1_6  = pin.GROUND     //
-	P1_7  = bcm283x.GPIO4  // High, GPCLK0
-	P1_8  = bcm283x.GPIO14 // Low,  UART0_TXD, UART1_TXD
-	P1_9  = pin.GROUND     //
-	P1_10 = bcm283x.GPIO15 // Low,  UART0_RXD, UART1_RXD
-	P1_11 = bcm283x.GPIO17 // Low,  UART0_RTS, SPI1_CE1, UART1_RTS
-	P1_12 = bcm283x.GPIO18 // Low,  PCM_CLK, SPI1_CE0, PWM0_OUT
-	P1_13 = bcm283x.GPIO27 // Low,
-	P1_14 = pin.GROUND     //
-	P1_15 = bcm283x.GPIO22 // Low,
-	P1_16 = bcm283x.GPIO23 // Low,
-	P1_17 = pin.V3_3       //
-	P1_18 = bcm283x.GPIO24 // Low,
-	P1_19 = bcm283x.GPIO10 // Low, SPI0_MOSI
-	P1_20 = pin.GROUND     //
-	P1_21 = bcm283x.GPIO9  // Low, SPI0_MISO
-	P1_22 = bcm283x.GPIO25 // Low,
-	P1_23 = bcm283x.GPIO11 // Low, SPI0_CLK
-	P1_24 = bcm283x.GPIO8  // High, SPI0_CE0
-	P1_25 = pin.GROUND     //
-	P1_26 = bcm283x.GPIO7  // High, SPI0_CE1
+	P1_1  pin.Pin    = pin.V3_3       // max 30mA
+	P1_2  pin.Pin    = pin.V5         // (filtered)
+	P1_3  gpio.PinIO = bcm283x.GPIO2  // High, I2C1_SDA
+	P1_4  pin.Pin    = pin.V5         //
+	P1_5  gpio.PinIO = bcm283x.GPIO3  // High, I2C1_SCL
+	P1_6  pin.Pin    = pin.GROUND     //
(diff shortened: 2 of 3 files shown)


maruel:
thanks! will review later today.

maxekman:
Note, I have (obviously) not tested all revisions.

host/rpi/rpi.go
```diff
@@ -128,17 +128,37 @@ func (d *driver) Init() (bool, error) {
...
+			0x0004, 0x0005, 0x0006, // B v2.0
+			0x0007, 0x0008, 0x0009, // A v2.0
+			0x000d, 0x000e, 0x000f, // B v2.0
+			0x0010,  // B+ v1.0
+			0x0011,  // Compute Module 1
+			0x0012,  // A+ v1.1
+			0x0013,  // B+ v1.2
+			0x0014,  // Compute Module 1
+			0x0015,  // A+ v1.1
+			0x90021, // A+ v1.1
```
maxekman: This and the one below are the current version A and B, they should maybe be mapped as V2 as they have the 40 pin header.

maruel: Ok I've read a bit more and I think this needs to be refactored to better represent the wide variety of boards.

What do you think about:  one variable to specify if the main GPIO header is 26 or 40 pins and a variable for presence of each secondary headers. These variables do not need to be exported. Then export below based on these variables. `Version` variable then should goes away since for practical purpose, Pi1 B+ and Pi2 B are similar ports wise and considering a Pi Zero W a Pi1 is very misleading, at this point it only means that the user cares about the type of CPU, which bcm283x is more appropriate to query this kind of information.

I found https://github.com/AndrewFromMelbourne/raspberry_pi_revision to be an interesting source of information.

maxekman:
Sounds like a good idea! I'll give it a spin and see what I can come up with. Thanks for the feedback.

maxekman:
@maruel What do you think about this? It should capture more in detail what the port mapping is trying to achieve.

host/rpi/rpi.go
```diff
@@ -124,28 +120,59 @@ func (d *driver) Init() (bool, error) {
 		return false, errors.New("Raspberry Pi board not detected")
 	}
 
-	// Initialize Version.
+	// Setup headers based on board revision.
 	//
 	// This code is not futureproof, it will error out on a Raspberry Pi 4
 	// whenever it comes out.
+	// Revision codes from: http://elinux.org/RPi_HardwareHistory
+	var has26PinP1Header bool
```
maruel: Please use the form: `has26PinP1Header := false`

Add has40PinP1Header := false too, since it shouldn't be registered for compute module.

host/rpi/rpi.go
```diff
@@ -124,28 +120,59 @@ func (d *driver) Init() (bool, error) {
...
+			0xa22042, // 2 Model B v1.2
+			0x900092, // Zero v1.2
+			0x900093, // Zero v1.3
+			0x920093, // Zero v1.3
+			0x9000c1: // Zero W v1.1
+		// Default 40 pin mapping.
+		case 0x0011, // Compute Module 1
+			0x0014,   // Compute Module 1
+			0xa020a0: // Compute Module 3 v1.0
+		// NOTE: Using the default 40 pin mapping for now.
```
maruel: The compute modules have a SODIMM connector and nothing else. So none of the other header should be registered by default, a [SODIMM header could be defined](https://www.raspberrypi.org/documentation/hardware/computemodule/schematics.md) instead but frankly, I'm not sure it'd be useful to anyone in practice.

I don't know if it is possible to detect if this extension board is connected at runtime:
https://www.raspberrypi.org/products/compute-module-io-board-v3/
but by default it should assume it is not, so no header should be registered.

maxekman:
Thanks for the comments, will update tonight.

maxekman:
@maruel Updated. Also added flags for audio and HDMI to better support the compute module.

maxekman:
Let me know if you want the fix commits squashed before merging.

host/rpi/rpi.go
```diff
@@ -163,13 +196,26 @@ func (d *driver) Init() (bool, error) {
...
-		if err := pinreg.Register("P5", [][]pin.Pin{
-			{P5_1, P5_2},
-			{P5_3, P5_4},
-			{P5_5, P5_6},
-			{P5_7, P5_8},
-		}); err != nil {
-			return true, err
+
+		// Only the A and B v2 PCB has the P5 header.
+		if hasP5Header {
```
maruel: Put it outside of the `has26PinP1Header` condition, it is orthogonal.

maruel:
I use "squash and merge" workflow to keep a linear history so no need to preemptively squash.

maxekman:
@maruel Makes sense. Updated.

host/rpi/rpi.go
```diff
@@ -45,36 +40,39 @@ var Version int
...
-	P1_21 = bcm283x.GPIO9  // Low, SPI0_MISO
-	P1_22 = bcm283x.GPIO25 // Low,
-	P1_23 = bcm283x.GPIO11 // Low, SPI0_CLK
-	P1_24 = bcm283x.GPIO8  // High, SPI0_CE0
-	P1_25 = pin.GROUND     //
-	P1_26 = bcm283x.GPIO7  // High, SPI0_CE1
-
-	// Raspberry Pi 2 and later:
+	// Rapsberry Pi A and B, 26 pin header:
+	P1_1  pin.Pin    = pin.V3_3       // max 30mA
```
maruel: The reason I hadn't specified the type there was to make it easier to leverage the bcm283x specific functionality of these pins but I guess it doesn't matter than much and it's quite awkward that half of the pins are pointers and the other half use interfaces.

host/rpi/rpi.go
```diff
@@ -124,28 +120,65 @@ func (d *driver) Init() (bool, error) {
...
+			0x90032,            // B+ v1.2
+			0xa01040,           // 2 Model B v1.0
+			0xa01041, 0xa21041, // 2 Model B v1.1
+			0xa22042, // 2 Model B v1.2
+			0x900092, // Zero v1.2
+			0x900093, // Zero v1.3
+			0x920093, // Zero v1.3
+			0x9000c1: // Zero W v1.1
+			has40PinP1Header = true
+			hasAudio = true
```
maruel: I don't consider the Pi Zero to have "proper audio" so I'd prefer to not register these.

maxekman: Definitely, my mistake.

maxekman:
@maruel Updated.

maruel:
Could you add yourself to AUTHORS and CONTRIBUTORS? Thanks!

maxekman:
@maruel Done!

AUTHORS
```diff
@@ -6,3 +6,4 @@
 Google Inc.
 Stephan Sperber
 Thorsten von Eicken <tve@voneicken.com>
+Max Ekman <max@looplab.se>
```
maruel: Please keep sorted.

maxekman: Fixed.

maruel:
Perfect! Thanks for getting this through.

maxekman:
My pleasure! Feels great to get involved.

PR #140, opened 2017-05-30
Move the bmp180 cmd into experimental

maxekman:
It looks like this cmd was forgotten when the experimental folder became more structured.


maruel:
ah and we also had forgot to add it to https://github.com/google/periph/blob/master/cmd/README.md.

I'll merge this but I think I've received the two I ordered (lost somewhere in my stack of junk) so we should write a smoke test and unit test and we could migrate this back into non-experimental soonish.

maxekman:
Perfect, let me fix the readme first.

maruel:
@fd0 fyi.

maxekman:
Ok, nvm :)

fd0:
Thanks!

PR #142, opened 2017-06-11
Don't error out when opening an already exported sysfs GPIO pin.

cbane:
The sysfs Pin.open function checked for the pin already being exported, but left p.err set when this happened. Only set the p.err once we've determined that it's a real error.


maruel:
(This cries for a proper unit test as I clearly _tried_ to make it work and failed but that can wait)

maruel:
I'll merge right away since you discovered and fixed serious problem and if you want. I encourage you to post a follow up to add yourself to AUTHORS and CONTRIBUTORS. Thanks again!

PR #143, opened 2017-06-12
spi: rename variables 'bus' to 'port' where applicable

maruel:
As people like to copy paste code, make sure that the semantic of the variable
names is always relevant w.r.t. the object they refer to. In particular, do not
use 'bus' to refer to an spi.Port or spi.PortCloser, as this is blur the
distinction.

Also rename a few fooName to fooID.


maruel:
@tve I'll merge right away but if you can find some time please take a cursory look in case there's something that itches you so I can post a follow up.

While most of this change is busy work, I still prefer to do it as I worry that leaving 'bus' and 'SPI' associated in the code base will confuse readers.

PR #145, opened 2017-07-16
pmem: rename Struct to AsPOD; add support for slice and array

maruel:
- bcm283x: Migrate 'smokeTest' struct size check to unit test.
- pmem: Improve package documentation.
- pmem: Remove the need from callers to use reflect.ValueOf().
- pmem: Add support to map a slice which calculate the len/cap automatically.
- pmem: Rename Struct() to AsPOD().
- pmem: Add AsPOD() support for native POD (plain old data) types like uint8,
  float64, etc.
- pmem: Add AsPOD() support for array of POD.
- pmem: Make AsPOD() check for struct much stricter.

### DIFF
--- a/host/allwinner/dma.go
+++ b/host/allwinner/dma.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"reflect"
 
 	"github.com/kr/pretty"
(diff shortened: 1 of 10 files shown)


maruel:
No worry if you don't have time, just say so; I'm quite confident with the change.

PR #147, opened 2017-08-19
conn: Improve documentation

maruel:
Explains the 'Bus → Port → Conn' concept that I had failed to put into words up to now

No code change.

### DIFF
--- a/conn/conn.go
+++ b/conn/conn.go
@@ -54,7 +54,7 @@ func (i Duplex) String() string {
 // - fmt.Stringer which returns something meaningful to the user like "SPI0.1",
 // "I2C1.76", "COM6", etc.
 //
-// - io.Reader and io.Writer as an way to use io.Copy() for half duplex
+// - io.Reader and io.Writer as a way to use io.Copy() for half duplex
 // operation.
 //
 // - io.Closer for the owner of the communication channel.
--- a/conn/conn.go
+++ b/conn/conn.go
@@ -24,7 +24,7 @@ const (
 	// Full means that communication occurs simultaneously both ways in a
 	// synchronized manner.
 	//
-	// Examples include SPI (except rare variants).
+	// Examples include SPI (except 3-wire variant).
 	Full Duplex = 2
 )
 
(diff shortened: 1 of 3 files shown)


conn/doc.go
```diff
@@ -4,6 +4,45 @@
...
+//
+// Concepts
+//
+// periph uses 3 layered concepts for interfacing:
+//
+//     Bus → Port → Conn
+//
+// Not every subpackage expose all 3 concepts. In fact, most packages don't.
+// For example, SPI doesn't exxpose Bus as the OSes generally only expose the
+// Port, that is, a CS line must be selected right upfront to get an handle.
```
mattetti: Suggestion `CS line`: `Chip Select (CS) line` so developers new to SPI don't get confused by the jargon.

conn/doc.go
```diff
@@ -4,6 +4,45 @@
...
+// sufficient to jump directly from a Bus to a Conn.
+//
+// Bus
+//
+// A Bus is a multi-point communication channel where one "master" and multiple
+// "slaves" communicate together. In the case of periph, the Bus handle is
+// assumed to be the "master". The "master" generally initiates communications
+// and selects the "slave" to talk to.
+//
+// As the "master", selecting a "slave" over a Bus converts the Bus into a Port.
```
mattetti: suggestion: `As the "master" selects a "slave" over a bus, a virtual port is automatically created` or something like that. The purpose of my suggestion is to highlight the fact that the developer doesn't have anything to do and that the bus and the port are available (AFAIK), in other words, the bus doesn't convert into a port and disappears. I'm putting myself in the choose of someone who just bought a rpi for instance and is discovering the various connections.

conn/doc.go
```diff
@@ -4,6 +4,45 @@
...
+//
+// Examples include SPI, I²C and 1-wire. In each case, selecting a
+// communication line (CS line for SPI, address for I²C or 1-wire) converts the
+// Bus into a Port.
+//
+// Port
+//
+// A port is a point-to-point communication channel that is not initialized and
+// thus, could be used with different parameters: number of bits per word,
```
mattetti: I'm confused by the wording here, what the relationship between the fact that the channel isn't initialized and the fact that different parameters can be used? It sounds like a port is a lower level construct to configure a parameterized communication channel when the settings can't be inferred from a bus for instance.

conn/doc.go
```diff
@@ -4,6 +4,45 @@
 
 // Package conn defines core interfaces for protocols and connections.
 //
-// Subpackages implements all the interfaces that can be used to connects
-// peripherals.
+// Subpackages implements all the supported interfaces that can be used to
+// connects peripherals.
+//
+// Concepts
```
mattetti: I would suggest to add a line talking about the connection between the physical world and this package. A developer might have a board and sensor or LED and wants to write code to control the LED. She might have no understanding of what buses, port or communications formats are available but found this package that offers connections to hardware. How can we offer basic concepts so she can clearly map her knowledge with this package and understand where to maybe dig (learning new things or exploring the code)? While this might be considered outside of the scope of code documentation, it might be extremely useful to new users and shouldn't prevent experienced users to get to what they are looking for.

conn/conn.go
```diff
@@ -43,18 +43,21 @@ func (i Duplex) String() string {
...
+// interfaces:
+//
 // - fmt.Stringer which returns something meaningful to the user like "SPI0.1",
-//   "I2C1.76", "COM6", etc.
-// - io.Reader and io.Writer as an way to use io.Copy() on a read-only or
-//   write-only device. For example the FLIR Lepton is a read-only device, the
-//   SSD1306 is a write-only device.
+// "I2C1.76", "COM6", etc.
+//
+// - io.Reader and io.Writer as an way to use io.Copy() for half duplex
```
mattetti: typo: `as an way` => `as a way`

Issue #20, opened 2016-10-19
bme280 forced mode


quinte17:
do we want to support the forced mode on the device?
this would mean that we have to add an option in which mode to start and makedev has to be aware of that.
we also would need to change the sense method or provide a new one. it has to enter forced mode wait a little bit and then get the data.

whats your opinion?

maruel:
Fixed with af3fbceec62dc44086c6731d9f181f33dc92a26c.

PR #150, opened 2017-08-20
devices: Add Environmental.SenseContinuous(); implement for bme280 

maruel:
- Unexport the Standby parameter.
- Have SenseContinuous accept time.Duration period instead.
- Set the filter only when using SenseContinuous.
- Do not enable automatic sensing upon initialization.
- Use forced mode by default.

### DIFF
--- a/devices/bme280/bme280.go
+++ b/devices/bme280/bme280.go
@@ -12,6 +12,9 @@ package bme280
 import (
 	"errors"
 	"fmt"
+	"log"
+	"sync"
+	"time"
 
 	"periph.io/x/periph/conn"
 	"periph.io/x/periph/conn/i2c"
@@ -31,8 +34,12 @@ import (
 type Oversampling uint8
 
 // Possible oversampling values.
+//
+// The higher the more time and power it takes to take a measurement. Even at
+// 16x for all 3 sensor, it is less than 100ms albeit increased power
+// consumption may increase the temperature reading.
 const (
-	No   Oversampling = 0
+	Off  Oversampling = 0
 	O1x  Oversampling = 1
 	O2x  Oversampling = 2
 	O4x  Oversampling = 3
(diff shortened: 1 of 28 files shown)


mattetti:
smoke tests are failing because they weren't updated:

> devices/bme280/bme280smoketest/bme280smoketest.go:127: unknown field 'Standby' in struct literal of type bme280.Opts

devices/bme280/bme280.go
```diff
@@ -31,113 +34,209 @@ import (
...
+		0xF4, byte(d.opts.Temperature)<<5 | byte(d.opts.Pressure)<<2 | byte(normal),
+	})
+	if err != nil {
+		return nil, err
+	}
+	sensing := make(chan devices.Environment)
+	d.stop = make(chan struct{})
+	go func() {
+		defer close(sensing)
+		d.senseContinuous(interval, sensing, d.stop)
```
mattetti: nit: I got confused not noticing that we were calling a different method and not making a recursive call. The recursive call didn't make a lot of sense (pun intended) and the args were different. A suggestions would be to call the unexported methods `sensingContinuously`.

devices/bme280/bme280.go
```diff
@@ -31,113 +34,209 @@ import (
...
-	// 4.1.
-	// Pressure: 0xF7~0xF9
-	// Temperature: 0xFA~0xFC
-	// Humidity: 0xFD~0xFE
-	buf := [0xFF - 0xF7]byte{}
-	if err := d.readReg(0xF7, buf[:]); err != nil {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	if d.stop != nil {
+		return errors.New("bme280: already sensing continuously")
```
mattetti: My bad, the lock seems to be released when running continuously in the background. I'm now kind of confused about the use case of the mutex. I thought it was a way to protect the user from running multiple polls in parallel.

devices/bme280/bme280.go
```diff
@@ -31,113 +34,209 @@ import (
...
-	F16  Filter = 4
+	NoFilter Filter = 0
+	F2       Filter = 1
+	F4       Filter = 2
+	F8       Filter = 3
+	F16      Filter = 4
 )
 
-// Dev is an handle to a bme280.
+// Dev is an handle to an initialized bme280.
```
maruel: Done

devices/bme280/bme280.go
```diff
@@ -31,113 +34,209 @@ import (
...
+	mu   sync.Mutex
+	stop chan struct{}
 }
 
 func (d *Dev) String() string {
 	return fmt.Sprintf("BME280{%s}", d.d)
 }
 
-// Sense returns measurements as °C, kPa and % of relative humidity.
+// Sense request a single one time measurement as °C, kPa and % of relative
```
maruel: Done

devices/bme280/bme280.go
maruel: The lock is only held while talking to the device. It serves two purpose:
- serialize access to the device
- serialize access to `d.stop`

devices/bme280/bme280.go
maruel: Done

maruel:
Paging @quinte17 since this PR will fix issue #20.

mattetti:
+1

PR #156, opened 2017-08-25
bmxx80: Merge bmp180 and bme280 and add support for bmp280

### LINKED ISSUE #155: Support for BMP280
Hello,

Can you please add support for the BMP280 sensor, i see that BME280 is supported which is basically the same (With the lack of humidity) but both sensors are backward and forward compatible, just the chip id is different 0x58 for BMP280 and 0x60 for BME280.

Here's a image from official datasheet:
datasheet
https://cdn-shop.adafruit.com/datasheets/BST-BME280_DS001-10.pdf

Thanks.

maruel:
- cmd/bmxx80 now can read any of these 3 devices.
- Rewrite the driver to support all 3 devices. More devices of the family (like
  the bme680) could be easily added.
- Add -ia to bmx280smoketest

Fix #155


devices/bmxx80/bmp180.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"encoding/binary"
+	"time"
+
+	"periph.io/x/periph/devices"
+)
+
+// sense180 reads the device's registers for bmp180.
+//
+// It must be called with d.mu lock held.
```
mattetti: why doesn't this function hold on the `d.mu` if it must be used?

maruel: Because `Sense()` checks for d.stop as a transaction, and sensing is part of the transaction.
https://github.com/google/periph/pull/156/files#diff-25acf98cebd7f455e61dc13223a25692R155

maruel:
Note for posterity; @Unafraid manually ran `bmx280smoketest` on BMP280 and I ran it on BME280. I also confirmed readings on BME280 and a BMP180.

PR #159, opened 2017-09-17
gpiostream: add PinIn and PinOut, add fakes.

maruel:
- PinIn only accepts BitStream, as supporting EdgeStream would require
  significantly more work and it's not obvious yet this would be valuable.
- Add gpiostreamtest to enable writting unit tests.

### DIFF
--- a/conn/gpio/gpiostream/gpiostream.go
+++ b/conn/gpio/gpiostream/gpiostream.go
@@ -8,6 +8,8 @@ package gpiostream
 import (
 	"sort"
 	"time"
+
+	"periph.io/x/periph/conn/gpio"
 )
 
 // Stream is the interface to define a generic stream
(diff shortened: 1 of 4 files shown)


maruel:
I'm a bit unhappy about the API but I need something more generic to be able to stream to a pin. I added a caveat accordingly.

conn/gpio/gpiostream/gpiostream.go
```diff
@@ -136,6 +138,32 @@ func (p *Program) Duration() time.Duration {
 	return d
 }
 
+//
+
+// PinIn allows to read a bit stream from a pin.
+//
+// Caveat: This interface doesn't enable sampling multiple pins in a
```
mattetti: those are important caveats indeed

maruel:
The main challenge I see is that moving these interfaces to experimental would mean to different packages named gpiostream (which is painful at import time). I guess I could still do it for sake of clarity.

PR #160, opened 2017-09-17
nrzled: add support for ws281x and clones

maruel:
- Includes driver and CLI.


experimental/cmd/nrzled/main.go
```diff
@@ -0,0 +1,186 @@
...
+
+func mainImpl() error {
+	verbose := flag.Bool("v", false, "verbose mode")
+	pin := flag.String("p", "", "GPIO pin to use")
+
+	numPixels := flag.Int("n", 150, "number of pixels on the strip")
+	hz := flag.Int("s", 400000, "speed in Hz")
+	channels := flag.Int("channels", 3, "number of color channels, use 4 for RGBW")
+	color := flag.String("color", "208020", "hex encoded color to show")
+	imgName := flag.String("img", "", "image to load")
```
mattetti: I'd suggest to add a fixture image to show the potential. I was actually curious to see what kind of images you used in your demos and if you generated them using another tool?

experimental/devices/nrzled/nrzled.go
```diff
@@ -0,0 +1,221 @@
...
+	}, nil
+}
+
+//
+
+// raster converts a RGB/RGBW input stream into a binary output stream as it
+// must be sent over the GPIO pin.
+//
+// `in` is RGB 24 bits or RGBW 32 bits. Each bit is encoded over 3 bits so the
+// length of `out` must be 3x as large as `in`.
```
mattetti: shouldn't we check on that in the code to avoid an invalid slice access?

experimental/devices/nrzled/nrzled.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+	}
+	_ = d.p.StreamOut(&d.b)
+}
+
+// Write accepts a stream of raw RGB/RGBW pixels and sends it as NRZ encoded
+// stream.
+//
+// This bypasses the back buffer.
+func (d *Dev) Write(pixels []byte) (int, error) {
```
mattetti: I think this API can be better documented. I expected it to be the defacto API to set the pixels but your comment about bypassing the back buffer makes it sound like this is more of a hack or workaround to a better solution. Is it?
Also I'm not sure what to set the byte values to, is that a RGB byte value?

PR #165, opened 2017-09-29
nrzled: add support for ws281x and clones

maruel:
- Includes driver and CLI. Completely untested.


maruel:
There's a chance the code is broken but it is "close enough" that it should be simple fix, and that offload a fair chunk of local code. :)  The driver is at 100% line coverage already.

maruel:
I'll commit right away since the code is inoffensive and all under `experimental/`, @mattetti feel free to comment if you see any snag post commit.

Issue #168, opened 2017-12-03
Getting PWM to work with Raspberry Pi Zero W


mikegleasonjr:
Hi,

Whenever I use `pi-blaster` to output a `pwm` signal on a pin, it works correctly (the led turns on):

```
sudo ./pi-blaster -D -g 18
...
echo "18=0.75" > /dev/pi-blaster
```
But when I try with `periph`, it doesn't work (I have no errors, but the pin doesn't output any current, the LED stays off):

```
package main

import (
	"fmt"
	"os"
	"os/signal"
	"time"

	"periph.io/x/periph"
	"periph.io/x/periph/conn/gpio"
	"periph.io/x/periph/host/rpi"
)

func main() {
	periph.Init()
	p := rpi.P1_12 // GPIO18
	pwm := p.(gpio.PinPWM)

	err := pwm.PWM(gpio.Duty(50535), 100*time.Microsecond)
	if err != nil {
		fmt.Fprint(os.Stderr, err)
		return
	}

	c := make(chan os.Signal, 1)
	signal.Notify(c, os.Interrupt)
	<-c
}
```

![img_20171203_200641](https://user-images.githubusercontent.com/1644534/33529245-f41ff646-d865-11e7-9f5f-ece4b5ff70dc.jpg)

maruel:
Did it print an error message? To work, you have to run it as root. Also I never tested having both loaded, as pi-blaster will also try to setup the PWM clock source.

There's also known bugs that the clock is not canceled correctly yet; https://periph.io/x/periph/host/bcm283x#pkg-note-bug

mikegleasonjr:
Hi !

* No error messages (`err == nil`)
* I noticed that `pi-blaster` and my own code can conflict, I tried both separately after clean reboots. Whenever I started `pi-blaster`, I started it as not deamon (`-D`) so that I could stop it.

I am using the latest `Stretch` (end of november) with a clean install.

maruel:
Just a note that pi-blaster plays with the same CPU registers and the state modification is not bound by process lifetime. Sometimes resetting the state requires really cutting the power.

That said the implementation is still known to be buggy but should have worked. Will double check in the next days.

simokawa:
I also found current hardware PWM does not work. I found a few problems.
1. PWM address map is broken. For example,  pwmMemory.rng1 is offset by 4 bytes, we need a padding before it.
2. Current configuration of dat1 looks strange.
I will send a pull request for 1.

I am using Pi B+.
[...]

PR #169, opened 2017-12-04
Fix PWM Address Map: add missing paddings.

simokawa:
Fix a problem I described in issue #168.


maruel:
Ugh thanks, I totally misread the datasheet. It's probably because I had too many local branches as I've been working on DMA support and probably mis-merged them, reintroducing the error. :(

mikegleasonjr:
@maruel No worries

I will retry my tests on my board

PR #171, opened 2017-12-09
Make hardware PWM work.

simokawa:
- Use fix 250MHz clock
- Set PWM registers correctly.

### DIFF
--- a/host/bcm283x/clock.go
+++ b/host/bcm283x/clock.go
@@ -215,7 +215,7 @@ func calcSource(hz uint64, maxWaitCycles int) (clockCtl, int, int, uint64, error
 	if hz < 1 {
 		return 0, 0, 0, 0, fmt.Errorf("bcm283x-clock: desired frequency %dHz must be >1hz", hz)
 	}
-	if hz > 250000000 {
+	if hz > clk500MHz {
 		return 0, 0, 0, 0, fmt.Errorf("bcm283x-clock: desired frequency %dHz is too high", hz)
 	}
 	// http://elinux.org/BCM2835_datasheet_errata states that clockSrc19dot2MHz
@@ -227,7 +227,6 @@ func calcSource(hz uint64, maxWaitCycles int) (clockCtl, int, int, uint64, error
 	// Try 500Mhz.
 	div, wait = findDivisorExact(clk500MHz, hz, maxWaitCycles)
 	if div != 0 {
-		fmt.Println("SrcPLLD", clockSrcPLLD, div, wait, hz)
 		return clockSrcPLLD, div, wait, hz, nil
 	}
 
--- a/host/bcm283x/clock.go
+++ b/host/bcm283x/clock.go
@@ -289,7 +289,6 @@ func (c *clock) setRaw(ctl clockCtl, div int) error {
 	c.ctl = clockPasswdCtl | ctl
 	Nanospin(10 * time.Microsecond)
 	c.ctl = clockPasswdCtl | ctl | clockEnable
-	fmt.Println("clock", c.ctl.String(), c.div.String())
 	if c.div != d {
 		return errors.New("can't write to clock divisor CPU register")
 	}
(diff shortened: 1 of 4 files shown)


host/bcm283x/clock.go
```diff
@@ -215,7 +215,7 @@ func calcSource(hz uint64, maxWaitCycles int) (clockCtl, int, int, uint64, error
 	if hz < 1 {
 		return 0, 0, 0, 0, fmt.Errorf("bcm283x-clock: desired frequency %dHz must be >1hz", hz)
 	}
-	if hz > 25000000 {
+	if hz > clk500MHz {
```
maruel: I had selected 25Mhz because of:
- p. 105 about the MASH filter: "the user must ensure that the module
is not exposed to frequencies higher than 25MHz"

But the system can go higher than this, e.g.:
- p. 106 "The maximum operating frequency of the General Purpose clocks is ~125MHz at 1.2V but
this will be reduced if the GPIO pins are heavily loaded or have a capacitive load. "
- p. 21 "If the system clock is 250 MHz and the speed field is zero the SPI clock frequency is 125
MHz. The practical SPI clock will be lower as the I/O pads can not transmit or receive
signals at such high speed."

So I think choosing 125Mhz would be a good idea.

host/bcm283x/clock.go
```diff
@@ -284,10 +284,10 @@ func (c *clock) setRaw(ctl clockCtl, div int) error {
 	}
 	d := clockDiv(div << clockDiviShift)
 	c.div = clockPasswdDiv | d
-	Nanospin(10 * time.Nanosecond)
+	Nanospin(10 * time.Microsecond)
```
maruel: I forget why I had chosen this value. 2x 10µs seems a bit slow though so it be great if we could figure out which lowest value we could use. If you can assert that 10ns fails, we should at least add a TODO to clarify that we don't know the exact lower bound needed.

host/bcm283x/gpio.go
```diff
@@ -368,18 +368,28 @@ func (p *Pin) PWM(duty gpio.Duty, period time.Duration) error {
 		return nil
 	}
 
-	// TODO(maruel): Leverage oversampling.
-	if _, _, err := clockMemory.pwm.set(uint64(time.Second/period), 1); err != nil {
+	base_freq := uint64(250 * 1000 * 1000) // 250MHz
```
maruel: Please keep the TODO, I have a local branch to eventually support approximative clocks via oversampling.

host/bcm283x/clock.go
simokawa: Thanks for the info, 125MHz sounds reasonable. I also changed default PWM frequency to 25MHz.

host/bcm283x/clock.go
simokawa: I thought it was typo because 10µs is used everywhere else. Looks like it was intentional, I reverted them and it seems working so far.

host/bcm283x/gpio.go
simokawa: Reverted.

host/bcm283x/gpio.go
```diff
@@ -369,17 +369,28 @@ func (p *Pin) PWM(duty gpio.Duty, period time.Duration) error {
...
+	if _, _, err := clockMemory.pwm.set(base_freq, 1); err != nil {
 		return p.wrap(err)
 	}
+	// Bit shift for PWM0 and PWM1
 	shift := uint((p.number & 1) * 8)
-	// rng1/rng2 are already 32.
 	if shift == 0 {
-		pwmMemory.dat1 = uint32(duty)<<16 | uint32(duty)
+		pwmMemory.rng1 = uint32(rng)
+		Nanospin(10 * time.Microsecond)
```
maruel: Does this needs 10µs too or 10ns works?

maruel:
And you can add yourself to [CONTRIBUTORS](https://github.com/google/periph/blob/master/CONTRIBUTORS) if you want.

host/bcm283x/gpio.go
simokawa: 10ns worked for me.

PR #173, opened 2017-12-10
bcm283x.Pin: Improve PWM() and implement conn.Resource

maruel:
- Implement correctly stopping the GPCLK/PWM clock source. This was a known bug.
  This also fixes PWM(0, ...) and PWM(gpio.DutyMax, ...) to work as expected.
- Halt() permits to safely stop the clock source, and edge detections.

### DIFF
--- a/host/bcm283x/gpio.go
+++ b/host/bcm283x/gpio.go
@@ -278,7 +278,10 @@ func (p *Pin) Out(l gpio.Level) error {
 
 // Halt implements conn.Resource.
 //
-// If the pin was running a PWM or waiting for edges, it is halted.
+// If the pin is running a clock, PWM or waiting for edges, it is halted.
+//
+// In the case of clock or PWM, all pins with this clock source are also
+// disabled.
 func (p *Pin) Halt() error {
 	if p.usingEdge {
 		if err := p.edge.In(gpio.PullNoChange, gpio.NoEdge); err != nil {
@@ -302,7 +305,8 @@ func (p *Pin) Halt() error {
 // PWM1 is exposed on pins 13, 19, 41 and 45.
 //
 // PWM0 and PWM1 share the same clock source, so even if they can have
-// independent duty cycle, they must have the same period.
+// independent duty cycle, they must have the same period. Pins shared on one
+// PWM (0 or 1) must have the same duty cycle.
 //
 // Clock pins
 //
--- a/host/bcm283x/gpio.go
+++ b/host/bcm283x/gpio.go
@@ -304,9 +304,8 @@ func (p *Pin) Halt() error {
 //
 // PWM1 is exposed on pins 13, 19, 41 and 45.
 //
-// PWM0 and PWM1 share the same clock source, so even if they can have
-// independent duty cycle, they must have the same period. Pins shared on one
-// PWM (0 or 1) must have the same duty cycle.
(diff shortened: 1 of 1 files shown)


maruel:
Tested this branch to work with this code:
https://gist.github.com/maruel/f0230fce67aa713154d84c9e2e175e1c

Paging @mattetti  and @simokawa for review. Thanks!

host/bcm283x/gpio.go
```diff
@@ -179,6 +180,9 @@ func (p *Pin) In(pull gpio.Pull, edge gpio.Edge) error {
 	if gpioMemory == nil {
 		return p.wrap(errors.New("subsystem not initialized"))
 	}
+	if err := p.haltClock(); err != nil {
```
simokawa: This is a bit tricky when we enable a PWM channel for multiple pins, for example,  12 and 18, then disable it for one of them.

I guess current implementation will disable PWM for both pins when we disable PWM for one of them. This may be different from user's expectation.

I think the current behavior is ok as long as commented somewhere.

maruel: Thanks for bringing this up. Added comment in Halt(). I'll eventually sort it out but this can wait to a follow up.

host/bcm283x/gpio.go
```diff
@@ -276,23 +276,49 @@ func (p *Pin) Out(l gpio.Level) error {
...
+// PWM outputs a periodic signal on supported pins.
+//
+// PWM pins
+//
+// PWM0 is exposed on pins 12, 18 and 40.
+//
+// PWM1 is exposed on pins 13, 19, 41 and 45.
+//
+// PWM0 and PWM1 share the same clock source, so even if they can have
+// independent duty cycle, they must have the same period.
```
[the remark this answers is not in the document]
maruel: Exact.

host/bcm283x/gpio.go
mattetti: I agree, that wouldn't have been my expectation either. Great catch!
@maruel I don't think we have a PWM example, maybe you should consider adding your https://gist.github.com/maruel/f0230fce67aa713154d84c9e2e175e1c example to the repo or website?

host/bcm283x/gpio.go
```diff
@@ -302,7 +305,8 @@ func (p *Pin) Halt() error {
 // PWM1 is exposed on pins 13, 19, 41 and 45.
 //
 // PWM0 and PWM1 share the same clock source, so even if they can have
-// independent duty cycle, they must have the same period.
+// independent duty cycle, they must have the same period. Pins shared on one
```
simokawa: No, they don't need to have the same period if the are on different PWM channels.
For example, you can set different periods for GPIO 12 and 13 in your sample code.

I think what you call "oversampling" is already there.
The clock is fixed to 25MHz and we specify the periods by rng1 and rng2 registers as well as duties are specified by dat1 and dat2 independently.

host/bcm283x/gpio.go
maruel: I plan to write an example once it is bug-fixed enough.

host/bcm283x/gpio.go
maruel: True, but it used to not be the case, fixed the doc.

maruel:
Thanks for the review! It's great to have someone take a look to find all the errors I make.

PR #178, opened 2017-12-11
bcm283x: Implement FastOut and corresponding benchmark

maruel:
Paging @simokawa .

I started writing a blog post about it at https://github.com/periph/website/blob/blog_gpio_perf/site/content/news/2017/gpio_perf.md

### DIFF
--- a/host/bcm283x/gpio.go
+++ b/host/bcm283x/gpio.go
@@ -266,6 +266,17 @@ func (p *Pin) Out(l gpio.Level) error {
 		return err
 	}
 	// Change output before changing mode to not create any glitch.
+	p.FastOut(l)
+	p.setFunction(out)
+	return nil
+}
+
+// FastOut sets a pin output level with Absolutely No error checking.
+//
+// Out() Must be called once first before calling FastOut(), otherwise the
+// behavior is undefined. Then FastOut() can be used for minimal CPU overhead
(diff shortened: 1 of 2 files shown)


host/bcm283x/gpio.go
```diff
@@ -266,6 +266,16 @@ func (p *Pin) Out(l gpio.Level) error {
 		return err
 	}
 	// Change output before changing mode to not create any glitch.
+	p.FastOut(l)
+	p.setFunction(out)
+	return nil
+}
+
+// FastOut sets a pin output level with Absolutely No error checking.
```
mattetti: We might want to specify what kind of errors could happen and why/when to use that version vs `Out`.

host/bcm283x/gpio.go
```diff
@@ -266,6 +266,16 @@ func (p *Pin) Out(l gpio.Level) error {
...
+//
+// Out() Must be called once first, then FastOut() can be used for minimal CPU
```
mattetti: what happens of Out isn't called first?

maruel: Undefined behavior. Updated the comment to be clearer, wdyt?

host/bcm283x/gpio.go
maruel: Updated comment to call out undefined behavior.

PR #179, opened 2017-12-11
Add reading benchmarks

maruel:


maruel:
The surprising point is that the read benchmark outputs 2x faster numbers but I don't see a way around that.

PR #181, opened 2017-12-11
allwinner: hardcode groups in Pin.Out()

maruel:
This improves performance on a C.H.I.P. by 10% from 490kHz to 540kHz.

### DIFF
--- a/host/allwinner/gpio.go
+++ b/host/allwinner/gpio.go
@@ -216,16 +216,53 @@ func (p *Pin) Out(l gpio.Level) error {
 		}
 		p.usingEdge = false
 	}
-	p.setFunction(out)
-	// TODO(maruel): Set the value *before* changing the pin to be an output, so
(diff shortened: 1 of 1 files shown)


host/allwinner/gpio.go
```diff
@@ -216,16 +216,53 @@ func (p *Pin) Out(l gpio.Level) error {
...
+			gpioMemory.groups[6].data |= bit
+		} else {
+			gpioMemory.groups[6].data &^= bit
+		}
+	case 7:
+		if l {
+			gpioMemory.groups[7].data |= bit
+		} else {
+			gpioMemory.groups[7].data &^= bit
+		}
```
mattetti: we might want a default case with the old implementation as a safety net

maruel: We know the implementation is correct, as all the values are controlled by the driver.

PR #182, opened 2017-12-11
allwinner: Implement FastOut()

maruel:
This raises output performance from 540kHz to 1.8MHz on a C.H.I.P.!

Include many documentation updates, especially for PinPL to reduce the
duplication of documentation, which always leads to a stale copy. :)

### DIFF
--- a/host/allwinner/gpio.go
+++ b/host/allwinner/gpio.go
@@ -165,8 +165,9 @@ func (p *Pin) In(pull gpio.Pull, edge gpio.Edge) error {
 	return nil
 }
 
-// Read returns the current level of the pin. Due to the way the Allwinner hardware functions it
-// will do this regardless of the pin's function but this should not be relied upon.
+// Read return the current pin level and implements gpio.PinIn.
+//
+// This function is very fast.
 func (p *Pin) Read() gpio.Level {
 	if gpioMemory == nil || !p.available {
 		return gpio.Low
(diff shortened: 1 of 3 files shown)


host/allwinner/gpio.go
```diff
@@ -165,16 +165,19 @@ func (p *Pin) In(pull gpio.Pull, edge gpio.Edge) error {
 	return nil
 }
 
-// Read returns the current level of the pin. Due to the way the Allwinner hardware functions it
-// will do this regardless of the pin's function but this should not be relied upon.
+// Read returns the current level of the pin.
+//
+// Due to the way the Allwinner hardware functions it will do this regardless
```
mattetti: what do you mean by `it will do this`?

host/allwinner/gpio.go
```diff
@@ -216,6 +219,17 @@ func (p *Pin) Out(l gpio.Level) error {
 		}
 		p.usingEdge = false
 	}
+	p.FastOut(l)
+	p.setFunction(out)
+	return nil
+}
+
+// FastOut sets a pin output level with Absolutely No error checking.
```
mattetti: same bas before, I'm curious what could go wrong, what kind of errors and what could be the consequences.

host/allwinner/gpio.go
maruel: I'd have to ask @tve , I forget the rationale. I'll remove the comment as I don't understand what it means either. :)

host/allwinner/gpio.go
maruel: In practice:
- If the driver didn't load, it will cause a panic
- otherwise, nothing happens

mattetti: I'd say that's something we'd want to let the users know because it will help them decide to use this faster approach or not

maruel: I disagree. The documentation tells them to use Out() first, if they don't, there is no guarantee. If they use Out() first, the behavior is defined. I don't think there is any more need to document the lack of guarantees. In particular I don't want to document the failing modes as I want to retain the right to change them.

PR #190, opened 2017-12-16
FYI: make bitstream by DMA work.

simokawa:
@maruel 
Now, I think I understand how to control DMA on RPi.
I am thinking to implement DMA based PWM in the master branch.


maruel:
Thanks, I had rebased and made changes, e.g. to use BitsLSB, so I incorporated your fixes and force pushed 2_bcm_dma

PR #191, opened 2017-12-16
Fix DMA completion detection.

simokawa:


simokawa:
Thanks for the merge. I am still not used with git/github. Let me try to squash commits next time.

PR #192, opened 2017-12-17
 cap1188 experimental device

mattetti:
This driver was tested against a adafruit version of the cap1188 https://learn.adafruit.com/adafruit-cap1188-breakout/overview and the pimoroni drumhat: https://shop.pimoroni.com/collections/raspberry-pi/products/drum-hat (which has its LED reversed and uses a different i2c address).

2 main areas should be improved before it is merged outside of the experimental folder:
* support for more features (tweaking config)
* make the driver more generic to be more of a cap11xx


cmd/cap1188/main.go
```diff
@@ -0,0 +1,153 @@
+package main
```
maruel: Add periph copyright, you can add yourself to COPYRIGHT and AUTHORS

cmd/cap1188/main.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	if *hz != 0 {
+		if err := i2cBus.SetSpeed(int64(*hz)); err != nil {
+			return fmt.Errorf("couldn't set the i2c bus speed - %s", err)
+		}
+	}
+	// The alert pin is the pin connected to the IRQ/interrupt pin and indicates
+	// when a touch event occurs.
+	alertPin := gpioreg.ByName("GPIO25")
```
maruel: Please don't hardcode these.

cmd/cap1188/main.go
```diff
@@ -0,0 +1,153 @@
...
+	if alertPin == nil {
+		return errors.New("invalid alert GPIO pin number")
+	}
+	if err := alertPin.In(gpio.PullUp, gpio.BothEdges); err != nil {
+		return err
+	}
+	if *verbose {
```
maruel: use log.Printf() instead.

cmd/cap1188/main.go
```diff
@@ -0,0 +1,153 @@
...
+	if resetPin == nil {
+		return errors.New("invalid reset GPIO pin number")
+	}
+	opts.AlertPin = alertPin
+	opts.ResetPin = resetPin
+
+	if dev, err = cap1188.NewI2C(i2cBus, opts); err != nil {
+		return fmt.Errorf("couldn't open cap1188 - %s", err)
+	}
+	time.Sleep(200 * time.Millisecond)
```
maruel: ?

cmd/cap1188/main.go
```diff
@@ -0,0 +1,153 @@
...
+	userAskedToLinkLeds := opts.LinkedLEDs
+	// unlinked LED demo
+	if err := dev.UnlinkLeds(); err != nil {
+		log.Println("Failed to unlink leds", err)
+	}
+	for i := 0; i < 8; i++ {
+		dev.SetLed(i, true)
+		time.Sleep(75 * time.Millisecond)
+	}
+	time.Sleep(200 * time.Millisecond)
```
maruel: This is more like a smoke test than a tool. Maybe write this as a smoketest?

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+
+// Package cap1188 controls a Microchip cap1188 device over I²C.
+
+// The device is a 8 Channel Capacitive Touch Sensor with 8 LED Drivers
+//
+// Datasheet
+//
+// The official data sheet can be found here:
+//
+// http://ww1.microchip.com/downloads/en/DeviceDoc/CAP1188%20.pdf
```
maruel: They added a redirector from no-whitespace to whitespace, so for sack of "omg they can't type", let's put:
http://ww1.microchip.com/downloads/en/DeviceDoc/CAP1188.pdf

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+	"periph.io/x/periph/conn/gpio"
+	"periph.io/x/periph/conn/i2c"
+	"periph.io/x/periph/conn/mmr"
+	"periph.io/x/periph/conn/spi"
+	"periph.io/x/periph/devices"
+)
+
+var (
+	// Debug is a debugging flag to turn on extra logging
+	Debug         bool
```
maruel: Why not per device?

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+	inputStatuses []TouchStatus
+)
+
+// TouchStatus is the status of an input sensor
+type TouchStatus int8
+
+func (t TouchStatus) String() string {
```
maruel: I generally use "stringer" to create a compact non-memory allocating form. But no big deal.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+	switch t {
+	case OffStatus:
+		return "Off    "
```
maruel: Don't include whitespace here, you don't know in which context this will be called and trailing whitespaces are be unexpected from a String() method.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+	isSPI      bool
+	opts       *Opts
+	mu         sync.Mutex
+}
+
+func (d *Dev) String() string {
+	return fmt.Sprintf("cap1188{%s}", d.d)
+}
+
+// Halt is a noop for the cap1188.
```
maruel: optional: Maybe turn off the LEDs?

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+
+	// read threshold
+	thresholds := [nbrOfLEDs]byte{}
+	if err = d.regWrapper.ReadStruct(0x30, &thresholds); err != nil {
+		return inputStatuses, wrap(fmt.Errorf("failed to read the threshold values - %s", err))
+	}
+
+	// convert the data into a sensor state
+	var touched bool
+	for i := uint8(0); i < nbrOfLEDs; i++ {
```
maruel: I'd prefer:
for i := uint8(0); i < uint8(len(inputStatuses)); i++ {

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+		touched = isBitSet(status, 7-i)
+
+		// TODO: check if the event is passed the threshold:
```
maruel: put your github username; i.e. TODO(mattetti):

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+	return d.regWrapper.WriteUint8(regID, v)
+}
+
+// resetMemInputStatus resets the in memory slice containing the state of the
+// inputs.
+func resetMemInputStatus() {
+	inputStatuses = make([]TouchStatus, nbrOfLEDs)
+}
+
+// b is the byte to check and position is the bit position
```
maruel: I'd recommend to inline this function, it has exactly one call site.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+// index 0 where 7 is the "most left bit".
+func isBitSet(b byte, pos uint8) bool {
+	return (b>>pos)&1 == 1
+}
+
+// twoComplementsToInt returns the int value of a standard 2’s complement number
```
maruel: Same,  it's not complex enough to warrant extracting.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+// Only works if the LEDs are not linked to the sensors
+func (d *Dev) SetLed(idx int, state bool) error {
+	if d.opts.LinkedLEDs {
+		return wrap(fmt.Errorf("can't manually set LEDs when they are linked to sensors"))
+	}
+	if idx > 7 || idx < 0 {
+		return wrap(fmt.Errorf("invalid led idx %d", idx))
+	}
+	if Debug {
+		fmt.Printf("Set LED state %d - %t\n", idx, state)
```
maruel: Use log.Printf(). And I think it should be part of the Opts.

But that raises an interesting point about potentially interesting logging lines for devices drivers.

Also, somehow I think using an Opts struct is the best long term way to go for most device driver. Not sure why it took me a year to decide.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+		case 0x00:
+			// do not do anything
+		default:
+			return nil, wrap(errors.New("given address not supported by device"))
+		}
+	}
+	if Debug {
+		fmt.Printf("cap1188: Connecting via I2C address: %#X\n", addr)
+	}
+	d := &Dev{d: &i2c.Dev{Bus: b, Addr: addr}, opts: opts, isSPI: false}
```
maruel: Don't keep a pointer to the original options, either copy the whole struct or inline it in the Dev.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+	reg_LEDOutputControl = 0x74
+)
+
+// Dev is a handle to a cap1188.
+type Dev struct {
+	d          conn.Conn
+	regWrapper mmr.Dev8
+	isSPI      bool
+	opts       *Opts
+	mu         sync.Mutex
```
maruel: This mutex is not used.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+}
+
+func (d *Dev) String() string {
+	return fmt.Sprintf("cap1188{%s}", d.d)
```
maruel: replace d.d with d.regWrapper.Conn and then you don't need member d at line 85.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+	"periph.io/x/periph/conn/i2c"
+	"periph.io/x/periph/conn/mmr"
+	"periph.io/x/periph/conn/spi"
+	"periph.io/x/periph/devices"
+)
+
+var (
+	// Debug is a debugging flag to turn on extra logging
+	Debug         bool
+	inputStatuses []TouchStatus
```
maruel: Please make it also part of Dev, so multiple instances of the drivers can work correctly.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,554 @@
...
+// This is quite more efficient than looping through each led and turn them on.
+func (d *Dev) AllLedsOn() error {
+	if d.opts.LinkedLEDs {
+		return wrap(fmt.Errorf("can't manually set LEDs when they are linked to sensors"))
+	}
+	return d.regWrapper.WriteUint8(reg_LEDOutputControl, 0xff)
+}
+
+// AllLedsOff turns all the LEDs off.
+// This is quite more efficient than looping through each led and turn them off.
```
maruel: Always put an empty line after the first sentence, this changes godoc output format.

experimental/devices/cap1188/cap1188.go
mattetti: Makes total sense

experimental/devices/cap1188/cap1188.go
mattetti: The challenge is that I use `d.d` to set `d.regWrapper`:

```golang
d.regWrapper = mmr.Dev8{Conn: d.d, Order: binary.LittleEndian}
```

How do I work around this issue?

experimental/devices/cap1188/cap1188.go
mattetti: while this isn't that complex nor used in multiple places, this data is formatted in a special way and I find that isolating the logic makes that clear. That said, after looking at the datasheet, this approach isn't used anywhere else so I'll indeed inline the decoding.

mattetti: that said @maruel should I also take the same approach for `isBitSet()`, it's pretty simple logic and only has 1 call site?

experimental/devices/cap1188/cap1188.go
mattetti: that would only work if the LEDs aren't linked. Maybe I can reset on halt? But that seems a bit odd

experimental/devices/cap1188/cap1188.go
maruel: Do not make makeDev() a method, make it a standalone function that accepts the Conn and returns a *Dev.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,550 @@
...
+		return d.inputStatuses, wrap(fmt.Errorf("failed to read the input values - %s", err))
+	}
+
+	// read deltas (in two's complement, capped at -128 to 127)
+	deltasB := [nbrOfLEDs]byte{}
+	if err = d.regWrapper.ReadStruct(0x10, &deltasB); err != nil {
+		return d.inputStatuses, wrap(fmt.Errorf("failed to read the delta values - %s", err))
+	}
+	// twoComplementsToInt returns the int value of a standard 2’s complement number
+	twoComplementsToInt := func(v byte) int {
```
maruel: I mean to inline the code directly below, e.g.

```go
for i, b := range deltasB {
  if b&0x80 == 0 {
    deltas[i] = int(b)
  } else {
    deltas[i] = int(b) - 0x100
  }
 }
```

but I think just doing this would work:
```go
for i, b := range deltasB {
  deltas[i] = int(int8(b))
 }
```

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,550 @@
...
+			d.inputStatuses[i] = PressedStatus
+		} else {
+			d.inputStatuses[i] = OffStatus
+		}
+	}
+
+	return d.inputStatuses, nil
+}
+
+// LinkLeds link the behavior of the LEDs to the touch sensors.
```
maruel: Should this be an option instead? What I ask is: is it valuable at runtime to toggle between linked and unlinked? I'm not sure there's a use case but I could be wrong.

maruel: s/Led/LED/g

mattetti: It can also be an option, but yes that's valuable at runtime. A good example is provided in main.go, you unlink the leds to display a pattern and relink. One can for instance build a Simon Says game using this approach.

cmd/cap1188/main.go
mattetti: cap1188_-_8_channel_capacitive_touch_sensor_with_8_led_drivers_datasheet_-_smsc
The device needs some time to be ready

mattetti: should I move this to the constructor instead?

cmd/cap1188/main.go
```diff
@@ -0,0 +1,157 @@
+// Copyright 2017 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
```
maruel: Be sure to include an empty line before the package name otherwise the copyright will be considered package documentation. :)

cmd/cap1188/main.go
maruel: The best would be to note in the constructor when it got initialized and then block in all calls if not enough time passed.

But sleeping in the constructor is fine otherwise.

experimental/devices/cap1188/cap1188_options.go
```diff
@@ -0,0 +1,140 @@
...
+// Address is only used on creation of an I²C-device. Its default value is 0x28.
+// It can be set to other values (0x29, 0x2a, 0x2b, 0x2c) depending on the HW
+// configuration of the ADDR_COMM pin. This has no effect with NewSPI()
+type Opts struct {
+	// Address is the I2C slave address to use
+	Address uint16
+
+	// Debug turns on extra logging capabilities
+	Debug bool
+	// LinkedLEDs indicates if the LEDs should be activated automatically
```
maruel: Oh since you noted that the functions are useful, then do not make it an option and have users call the function explicitly.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,541 @@
...
+			d.inputStatuses[i] = PressedStatus
+		} else {
+			d.inputStatuses[i] = OffStatus
+		}
+	}
+
+	return d.inputStatuses, nil
+}
+
+// LinkLEDs link the behavior of the LEDs to the touch sensors.
```
maruel: I'd prefer a single function that accepts a bool instead of LInkLEDs/UnlinkLEDs.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,541 @@
...
+// control the LEDs.
+func (d *Dev) UnlinkLEDs() error {
+	if err := d.regWrapper.WriteUint8(reg_LEDLinking, 0x00); err != nil {
+		return wrap(fmt.Errorf("failed to unlink LEDs - %s", err))
+	}
+	d.LinkedLEDs = false
+	return nil
+}
+
+// AllLEDsOn turns all the LEDs on.
```
maruel: Similarly, I'd prefer a single function that accepts a bool.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,541 @@
...
+	return d.clearBit(reg_LEDOutputControl, idx)
+}
+
+// Reset issues a soft reset to the device using the reset pin
+// if available.
+func (d *Dev) Reset() (err error) {
+	d.ClearInterrupt()
+	if d != nil && d != nil && d.ResetPin != nil {
+		if d.Debug {
+			fmt.Println("cap1188: Resetting the device using the reset pin")
```
maruel: Use log instead of fmt.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,541 @@
...
+		}
+		if err = d.ResetPin.Out(gpio.Low); err != nil {
+			return err
+		}
+		if err = d.ResetPin.Out(gpio.High); err != nil {
```
maruel: If the device needs a complete pulse, you need to sleep at least a microsecond before, otherwise the pulse could be very fast, or whatever the datasheet says.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,541 @@
...
+	return nil
+}
+
+// ClearInterrupt resets the interrupt flag
+func (d *Dev) ClearInterrupt() error {
+	// clear the main control bit
+	return d.clearBit(0x0, 0)
+}
+
+// Standby turns off the capacitive touch sensor inputs. The status registers will
```
maruel: Let's not export the functions not implemented.

In practice, do we need the 3 followings?

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,541 @@
...
+	}
+	if err = d.regWrapper.WriteUint8(0x20, config); err != nil {
+		return fmt.Errorf("failed to set the device configuration - %s", err)
+	}
+
+	config2 := (
+	// Linked LED Transition controls
+	// default 0: The Linked LED Transition controls set the min duty
+	// cycle equal to the max duty cycle
+	byte(0)<<7 |
```
maruel: You can document the bits you don't set but no need to add code for these (even if it's elided by the compiler). I generally do not document the functionality that are not expected to be used.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,541 @@
...
+		// Determines the ALERT# pin polarity and behavior.
+		// default 1: the ALERT# pin is active low and open drain.
+		byte(1)<<6 |
+		// Determines whether the device will reduce power consumption
```
maruel: Let's keep power saving out for now. Go programs can hardly run on coin cell batteries.

experimental/devices/cap1188/cap1188.go
```diff
@@ -0,0 +1,541 @@
...
+	if err != nil {
+		return err
+	}
+	v &= ^(1 << uint8(idx))
+	return d.regWrapper.WriteUint8(regID, v)
+}
+
+// b is the byte to check and position is the bit position
+// index 0 where 7 is the "most left bit".
+func isBitSet(b byte, pos uint8) bool {
```
maruel: This function is used exactly once and is trivial, please inline.

experimental/devices/cap1188/cap1188.go
maruel: Yes, thanks.

experimental/devices/cap1188/cap1188.go
maruel: It's not a big deal either way. In theory that would be a good place to put the device in deep sleep, just a TODO is good for now.

experimental/devices/cap1188/cap1188.go
mattetti: in practice we don't need them, they'd be nice to have but I personally don't need them.

cmd/cap1188/main.go
mattetti: done

experimental/devices/cap1188/cap1188_options.go
mattetti: options are embedded now, so it's all good, users can set the default behavior and change it as needed.

mattetti:
I made a bunch of changes, tested on the hardware, we should be good to go now.

cmd/cap1188/main.go
```diff
@@ -0,0 +1,159 @@
+// Copyright 2017 The Periph Authors. All rights reserved.
```
maruel: Please move this file into experimental too as it's a mix of smoketest and tool. I can do refactoring afterward. Then it's good for merge.

mattetti: Sounds good. I'm still a little bit confused about the difference between cmd and smoketest, especially because I based mine off of https://github.com/google/periph/blob/master/cmd/apa102/main.go

mattetti:
@maruel should be good to go now :)

PR #195, opened 2017-12-18
Enable DMA driven PWM for all GIO pins except PWM1 pins.

simokawa:
@maruel 
I am not sure if we should enable DMA depending on pins or introduce a flag to select DMA/hardware PWM/50%duty clock. What do you think?


simokawa:
Also, a test is broken. How can I fix it?
gpio_test.go:164: bcm283x-gpio (C1): videocore: failed to open the mailbox to the GPU: file I/O is inhibited

maruel:
You need to mock the code in the tests to make sure that the test would pass even on Windows.

host/bcm283x/gpio.go
```diff
@@ -326,12 +329,17 @@ func (p *Pin) FastOut(l gpio.Level) {
 // PWM pins
 //
 // PWM0 is exposed on pins 12, 18 and 40.
+// However, PWM0 is used for generating clock for DMA and unavailable for PWM.
```
maruel: Wrap on the line before.

host/bcm283x/gpio.go
```diff
@@ -326,12 +329,17 @@ func (p *Pin) FastOut(l gpio.Level) {
...
+// TODO(simokawa): Use other DREQ source like PCM.
```
maruel: Is it relevant to user documentation?

host/bcm283x/gpio.go
```diff
@@ -326,12 +329,17 @@ func (p *Pin) FastOut(l gpio.Level) {
...
 //
 // PWM1 is exposed on pins 13, 19, 41 and 45.
 //
 // PWM0 and PWM1 share the same 25Mhz clock source. The period must be a
 // divisor of 25Mhz.
 //
+// DMA driven PWM is aviable for all pins except PWM1 pins, its resolution is
```
maruel: "available"

host/bcm283x/gpio.go
```diff
@@ -395,58 +388,54 @@ func (p *Pin) PWM(duty gpio.Duty, period time.Duration) error {
...
-		switch clkID {
-		case 0:
-			clk = &clockMemory.gp0
-		case 2:
-			clk = &clockMemory.gp2
-		}
-		actual, waits, err := clk.set(hz, 1)
-		if err != nil {
+		var err error
+		// TODO(simokawa): Reuse DMA buffer if possible.
```
maruel: The thing I'm the most worried is to leave the host on a broken state. One thing that would encourage me is to have bcm283xsmoketest that turns on PWM on and off 20 times. If that works I'll feel better about it.

host/bcm283x/gpio.go
```diff
@@ -395,58 +388,54 @@ func (p *Pin) PWM(duty gpio.Duty, period time.Duration) error {
...
+		if err := p.haltDMA(); err != nil {
 			return p.wrap(err)
 		}
-		p.usingClock = true
-		if actual != hz {
-			return p.wrap(fmt.Errorf("asked for %dHz, got %dHz", hz, actual))
+		if p.dmaCh, p.dmaBuf, err = startPWMbyDMA(p, uint32(rng), dat); err != nil {
```
maruel: I'd prefer to have startPWMbyDMA() explicitly do:
```go
if dmaMemory != nil {
  return p.wrap("bcm283x-dma is not initialized; try running as root?")
}
```

or something like that.

host/bcm283x/gpio.go
simokawa: Good point, I will update the document tomorrow.

host/bcm283x/gpio.go
simokawa: I updated smoketest and run it 20 times. Looks good so far.

host/bcm283x/gpio.go
simokawa: I added it in startPWMbyDMA() but it introduce another test breakage.
Let me fix tests tomorrow.

host/bcm283x/bcm283xsmoketest/bcm283xsmoketest.go
```diff
@@ -47,17 +47,24 @@ func (s *SmokeTest) Run(f *flag.FlagSet, args []string) error {
 	}
 
 	start := time.Now()
-	pClk := &loggingPin{bcm283x.GPIO6, start}
+	pDMA := &loggingPin{bcm283x.GPIO6, start}
```
maruel: keep it as pClk since even if you are doing DMA with it, it's special functionality is to be a GPCLK.

host/bcm283x/bcm283xsmoketest/bcm283xsmoketest.go
```diff
@@ -47,17 +47,24 @@ func (s *SmokeTest) Run(f *flag.FlagSet, args []string) error {
...
-	if err := s.testClock(pClk, pPWM); err != nil {
+	if err := s.testPWMbyDMA(pDMA, pPWM); err != nil {
 		return err
 	}
-	if err := s.testPWM(pPWM, pClk); err != nil {
+	if err := s.testPWM(pPWM, pDMA); err != nil {
+		return err
+	}
+	// Make sure to stop PWM pins
+	if err := pPWM.In(gpio.PullDown, gpio.NoEdge); err != nil {
```
maruel: Use Halt() inside testPWM(). I forgot to add the call when I added the Halt() function.

host/bcm283x/bcm283xsmoketest/bcm283xsmoketest.go
```diff
@@ -80,9 +87,9 @@ func (s *SmokeTest) waitForEdge(p gpio.PinIO) <-chan bool {
 	return c
 }
 
-// testClock tests .PWM() for a clock pin.
```
maruel: Keep this test, add testPWMbyDMA as a separate test.

host/bcm283x/dma.go
```diff
@@ -514,11 +515,7 @@ func (d *dmaChannel) isAvailable() bool {
 // It doesn't clear the local controlBlock cached values.
 func (d *dmaChannel) reset() {
 	d.cs = dmaReset
-	// Clear bits if needed.
```
maruel: You should rebase on origin/master. In short:

git rebase -i origin/master
# squash them all
git mergetool
git push -f origin dma1pull

You'll want to configure vimdiff or something similar. e.g.:
https://github.com/maruel/bin_pub/blob/master/configs/.gitconfig#L71
There's a squash alias in there if you want.

You can also use a merge flow but since I always squash, it's easier to rebase locally. That keeps the history purely linear, which I prefer.

host/bcm283x/gpio.go
```diff
@@ -91,9 +92,11 @@ type Pin struct {
 	defaultPull gpio.Pull
 
 	// Mutable.
-	edge       *sysfs.Pin // Set once, then never set back to nil.
-	usingEdge  bool       // Set when edge detection is enabled.
-	usingClock bool       // Set when a GPCLK or PWM clock is used.
+	edge       *sysfs.Pin     // Set once, then never set back to nil.
+	usingEdge  bool           // Set when edge detection is enabled.
+	usingClock bool           // Set when a GPCLK or PWM clock is used.
+	dmaCh      *dmaChannel    // Set when DMA is used for PWM.
```
maruel: PWM or PCM

host/bcm283x/gpio.go
```diff
@@ -325,13 +328,18 @@ func (p *Pin) FastOut(l gpio.Level) {
 //
 // PWM pins
 //
-// PWM0 is exposed on pins 12, 18 and 40.
+// PWM0 is exposed on pins 12, 18 and 40. However, PWM0 is used for generating
+// clock for DMA and unavailable for PWM.
+// TODO(simokawa): Use other DREQ source like PCM.
```
maruel: We can't use PCM as a source, we'll use it for binary streaming, so remove the TODO here.

host/bcm283x/gpio.go
```diff
@@ -325,13 +328,18 @@ func (p *Pin) FastOut(l gpio.Level) {
...
 //
 // PWM1 is exposed on pins 13, 19, 41 and 45.
 //
 // PWM0 and PWM1 share the same 25Mhz clock source. The period must be a
```
maruel: This is irrelevant to mention PWM0 if PWM0 cannot be used anymore.

host/bcm283x/gpio_test.go
```diff
@@ -156,10 +158,15 @@ func TestPinPWM(t *testing.T) {
 
 	clockMemory = &clockMap{}
 	pwmMemory = &pwmMap{}
+	dmaBufAllocator = func(s int) (*videocore.Mem, error) {
```
maruel: You can put that into init() for the test.

host/bcm283x/gpio.go
```diff
@@ -325,13 +328,18 @@ func (p *Pin) FastOut(l gpio.Level) {
 //
```
maruel: It is very important in the documentation to call out that `Halt()` must be called before exiting the process or the host will be in broken state, potentially requiring an host hard reboot.

simokawa: Agree, especially DMA is problematic because it runs a infinite loop and ownership is not clear. I am not familiar with godoc but can I assume if I update the comments, godoc is updated automatically?

host/bcm283x/bcm283xsmoketest/bcm283xsmoketest.go
simokawa: done

host/bcm283x/bcm283xsmoketest/bcm283xsmoketest.go
simokawa: done

host/bcm283x/bcm283xsmoketest/bcm283xsmoketest.go
simokawa: Reverted but please note I actually disabled clock driven PWM for all pins.

host/bcm283x/dma.go
simokawa: I guess I correctly updated my origin/master and merged them this time.

host/bcm283x/gpio.go
simokawa: done.

host/bcm283x/gpio.go
simokawa: done

host/bcm283x/gpio_test.go
simokawa: done.

host/bcm283x/gpio.go
simokawa: It depends on application. Anyway, I have no plan to work on it in the short term, removed it.

host/bcm283x/bcm283xsmoketest/bcm283xsmoketest.go
simokawa: I removed testClock() again because it does not make sense now. (ex. expects error for arbitrary duty)
If you want to keep clock driven PWM for pins like GPIO6,  please let me know.

host/bcm283x/gpio.go
maruel: godoc.org is updated when you click the refresh link at the bottom. You can run godoc locally with:

go get -u -v golang.org/x/tools/cmd/godoc
godoc -http=:6060
Then visit http://localhost:6060/pkg/periph.io/x/periph/host/bcm283x

host/bcm283x/gpio.go
```diff
@@ -325,28 +328,22 @@ func (p *Pin) FastOut(l gpio.Level) {
...
-// Clock pins
-//
-// Clock GPCLK0 is exposed on pins 4, 20 32 and 34.
-//
-// Clock GPCLK2 is exposed on pins 6 and 43.
-//
-// Clocks can only be used with duty 0, gpio.DutyHalf or gpio.DutyMax. They
-// also have relatively limited frequency range from 4.8kHz to 1MHz with a
-// specific set of values supported.
+// DMA driven PWM is available for all pins except PWM1 pins, its resolution is
```
maruel: Please add: "The number of DMA driven PWM is limited."

host/bcm283x/gpio.go
```diff
@@ -462,42 +440,51 @@ func (p *Pin) DefaultPull() gpio.Pull {
...
+			return p.wrap(err)
+		}
+		p.dmaBuf = nil
+	}
+	return nil
+}
+
 // haltClock disables the GPCLK/PWM clock if used.
 func (p *Pin) haltClock() error {
+	p.haltDMA()
```
maruel: don't ignore the error value

host/bcm283x/gpio_test.go
```diff
@@ -160,6 +162,7 @@ func TestPinPWM(t *testing.T) {
 		t.Fatal(err)
 	}
 	// TODO(maruel): Fix test.
+	dmaMemory = &dmaMap{}
```
maruel: defer func() { dmaMemory = nil }()
so the value is reset correctly.

host/bcm283x/gpio.go
simokawa: Thanks for the explanation!

host/bcm283x/gpio.go
simokawa: done.

host/bcm283x/gpio.go
simokawa: done.

host/bcm283x/gpio_test.go
simokawa: done.

PR #202, opened 2017-12-22
bcm283x: add PinsRead/Clear/Set global functions

maruel:
These CPU specific functions permits to read or change multiple GPIOs at once.

### DIFF
--- a/host/bcm283x/gpio.go
+++ b/host/bcm283x/gpio.go
@@ -82,6 +82,62 @@ func Present() bool {
 	return false
 }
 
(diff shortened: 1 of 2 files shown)


maruel:
I'm not sure about the naming, maybe use the verb at the end, e.g. `Pins0To31Read` instead?

maruel:
Good point, I added an example, which I think should be useful enough to grasps what is happening. I made it so that it actually runs fine as a test.

mattetti:
👍 that's even better than what I was asking for. Pretty awesome!

PR #203, opened 2017-12-23
cap1188: refactor a bit.

maruel:
- Make errors consistents.
- Standardize comments.
- Make unit test much faster by not sleeping.
- Reduce the number of properties of Dev.
  - Do not export Opts.
- Commented out I/O for unused data.
- Slightly reduce memory usage.
- Comment out NewSPI() since it is not implemented.

### DIFF
--- a/experimental/devices/cap1188/cap1188.go
+++ b/experimental/devices/cap1188/cap1188.go
@@ -3,19 +3,18 @@
 // that can be found in the LICENSE file.
 
 // Package cap1188 controls a Microchip cap1188 device over I²C.
-// The device is a 8 Channel Capacitive Touch Sensor with 8 LED Drivers
+//
+// The cap1188 device is a 8 channel capacitive touch sensor with 8 LED drivers.
 //
 // Datasheet
 //
 // The official data sheet can be found here:
 //
 // http://ww1.microchip.com/downloads/en/DeviceDoc/CAP1188.pdf
-//
 package cap1188
 
 import (
 	"encoding/binary"
-	"errors"
 	"fmt"
 	"log"
 	"time"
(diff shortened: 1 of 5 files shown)


experimental/devices/cap1188/cap1188_test.go
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed under the Apache License, Version 2.0
 // that can be found in the LICENSE file.
 
-package cap1188_test
+package cap1188
```
mattetti: I suggest we keep the examples in a test package so the code can be copied/pasted directly.

experimental/devices/cap1188/cap1188.go
```diff
@@ -24,121 +23,87 @@ import (
...
 	for i := uint8(0); i < uint8(len(d.inputStatuses)); i++ {
-		// check if the bit is set.
-		touched = (status>>(7-i))&1 == 1
-
+		// Check if the bit is set.
 		// TODO(mattetti): check if the event is passed the threshold:
 		// deltas[i] > int(thresholds[i])
-
-		if touched {
+		if (status>>(7-i))&1 != 0 {
```
mattetti: I'd suggest to at least add a comment to explain that it means that the sensored was touched

experimental/devices/cap1188/cap1188.go
```diff
@@ -156,369 +121,396 @@ func (d *Dev) InputStatus() ([]TouchStatus, error) {
...
-	if !opts.InterruptOnRelease {
+	if !d.opts.InterruptOnRelease {
 		intOnRel = 1 // 0 = trigger on release
 	}
 
-	// enable all inputs
-	if err = d.regWrapper.WriteUint8(0x21, 0xff); err != nil {
-		return wrap(fmt.Errorf("failed to enable all inputs - %s", err))
+	// Enable all inputs.
+	if err := d.c.WriteUint8(0x21, 0xff); err != nil {
```
mattetti: is there an advantage of allocating a new local variable err for each of these conditions vs declaring and reusing the same one?

experimental/devices/cap1188/cap1188.go
```diff
@@ -156,369 +121,396 @@ func (d *Dev) InputStatus() ([]TouchStatus, error) {
...
+	// corresponding bit is set, then the appropriate LED output will change
+	// states defined by the LED Behavior controls in response to the capacitive
+	// touch sensor input.
+	regLEDLinking = 0x72
+	// regLEDOutputControl is the LED Output Control Register controls the output
+	// state of the LED pins that are not linked to sensor inputs.
+	regLEDOutputControl = 0x74
+)
+
+var sleep = time.Sleep
```
mattetti: clever, I should have thought about that!

experimental/devices/cap1188/cap1188_test.go
```diff
@@ -277,24 +263,24 @@ func TestDev_InputStatus(t *testing.T) {
 		bus := &i2ctest.Playback{
 			Ops: append(setupPlaybackIO(), []i2ctest.IO{
 				{Addr: 40, W: []byte{0x3}, R: []byte{1 << 7}},
-				{Addr: 40, W: []byte{0x10}, R: []byte{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}},
-				{Addr: 40, W: []byte{0x30}, R: []byte{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}},
+				//{Addr: 40, W: []byte{0x10}, R: []byte{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}},
```
mattetti: I'm surprised you were able to remove those

experimental/devices/cap1188/cap1188.go
maruel: It's now needed because there is no more a function local err declared above.

experimental/devices/cap1188/cap1188_test.go
maruel: It must be in the same package to be able to override "sleep"

experimental/devices/cap1188/cap1188_test.go
maruel: It works because it is in the same package. That's the idiomatic way.

experimental/devices/cap1188/cap1188.go
maruel: I switched the order, as the code was doing the indexing in reverse order.

experimental/devices/cap1188/cap1188_test.go
mattetti: Maybe we can move the example to its own file then, it will still have a sleep but it's only 1 example. Thoughts?

experimental/devices/cap1188/cap1188.go
mattetti: I'm not sure what you mean but you can physically compare the output to the touched sensored/LED. I can test again the changes if you'd like

experimental/devices/cap1188/cap1188_test.go
maruel: Yes good idea, done.

PR #204, opened 2017-12-23
Remove all time.Sleep calls during tests

maruel:
It slows down tests, nobody wants that.


devices/bmxx80/bmxx80.go
```diff
@@ -514,6 +514,8 @@ var defaults = Opts{
 	Humidity:    O4x,
 }
 
+var doSleep = time.Sleep
```
mattetti: why are we using `sleep` and `doSleep`, shouldn't we be consistent?

maruel: There's already a package global sleep variable and I didn't feel like renaming it.

maruel:
gohci

Issue #198, opened 2017-12-22
bcm283x: Do not use dmaWaitcyclesMax.


simokawa:
It looks like `dmaWaitCycles` is a count for bus clock rather than `DREQ` which is our DMA clock. 
- Non-zero waits in `initBlock` for DMA CB does not work as a divisor of the DMA clock and we should remove the argument. 
- `dmaWaitcyclesMax` is not meaningful for upper bound of the divisor. We want to replace it with 0 in most cases.

maruel:
Merged to https://github.com/periph/host/issues/44.

PR #205, opened 2017-12-23
Stability improvements for DMA.

simokawa:
- Do not use waits for divisor #198
- Allow larger divisor in setPWMClockSource.
- Start clock before staring DMA to avoid initial glitch.
- Support very long periods in PWM using non-lite DMA engine.


maruel:
gohci

maruel:
(Trying to make gohci to work for non-member PR, may send a few silly comments)

maruel:
gohci

Issue #206, opened 2017-12-23
gpiostream: Stream.Duration() bug


simokawa:
If I understand correctly, Stream.Resolution() is duration for a single bit and Duration() should return for all **bits***.
Current implementation uses the count of **bytes** instead of **bits**, this looks wrong to me.
https://github.com/google/periph/blob/master/conn/gpio/gpiostream/gpiostream.go#L68
It should be
`return b.Res * time.Duration(len(b.Bits)) * 8`
Yeah, it is confusing that len(b.Bits) returns number of **bytes**. 

There may be some code assuming this bug.

maruel:
Yep :(

simokawa:
Assuming number of bits to be 8*len(b.Bit) may work for most cases but we may want to specify number of bits explicitly in the struct. For example, gpiostream.Program{} could be a loop of 3bit-stream.

maruel:
Ported to https://github.com/periph/conn/issues/34.

Issue #172, opened 2017-12-09
bcm283x: complete DMA support


simokawa:
What is the status of BCM283X DMA support?
I am interested in using DMA for
1. Output: PWM for usual GIO pins
2. Input: Frequent periodic GIO pin read (for quadrature encoders)

Both seem supported by [Pigpio](http://abyz.me.uk/rpi/pigpio/).
I found some code in bcm283x/dma.go but it does not seem to have an interface for it.

I am considering to play with DMA in the next few weeks. If there is anything I can help to implementing them in periph, please let me know.

Thanks,

maruel:
Thanks for the inquire! I have various local branches in various states of brokenness. :) There were challenges:
- Figure out a way to describe a stream both as a list of events and as a bit stream, done as [periph.io/x/periph/conn/gpio/gpiostream](https://periph.io/x/periph/conn/gpio/gpiostream).
- Make it work at all on one CPU. Kinda done.
- From the various ways to drive inputs and outputs, find the one that has the best trade off latency vs memory.
  - Specifically with the bcm283x, I tried all of:
    - using one DMA channel, daisy chained one bit toggle per `controlBlock`, one writing to `gpioMap.outputSet`, one to `gpioMap.outputClear`. 
    - using two simultaneous DMA channels one per 0 toggle and one per 1 toggle, one writing to `gpioMap.outputSet`, one to `gpioMap.outputClear`. 
    - using one DMA channel, writing to `gpioMap.level`.
    - using the PCM engine to have a fully synchronized clock + input + output (prototyping).
  - In particular, using a dual DMA channels solution has too much jitter to be able to use it to drive NRZ LEDs. :( 
- Implement on at least two CPU classes to understand differences and similarities. Didn't need to be fully functional, but that was a useful learning experience.
- Figure out an API that is simpler yet powerful to wrap this up.
  - In particular, I'd like to enable functionality for simultaneous I/O if possible.
  - Figure out a teardown / cleanup API that is easy and safe to use, otherwise the user is left in a state where (s)he has to hard reset the host.

I've pushed one of my "less bad" branches as https://github.com/google/periph/tree/2_bcm_dma and https://github.com/google/periph/tree/7_allwinner. Allwinner support is not needed for this feature, it's just FYI.

simokawa:
Thanks for the information! I will take a look at your branch.

maruel:
[...]

PR #207, opened 2017-12-25
bcm283x: Implement bit stream functions, not enabled yet.

simokawa:
- Based on @maruel's implementation.
- DMA driven stream shares a fixed PWM and DREQ clock. Oversampling is
  used to get desired resolution. 
- Addressed #206 #172


host/bcm283x/dma.go
```diff
@@ -702,6 +760,237 @@ func startPWMbyDMA(p *Pin, rng, data uint32) (*dmaChannel, *videocore.Mem, error
...
+	regSet := gpioBaseAddr + 0x1C + 4*uint32(p.number/32)
+	if err := cb[0].initBlock(uint32(bufSet.PhysAddr()), regSet, uint32(l), false, true, true, false, dmaPWM); err != nil {
+		return err
+	}
+	regClear := gpioBaseAddr + 0x28 + 4*uint32(p.number/32)
+	if err := cb[1].initBlock(uint32(bufClear.PhysAddr()), regClear, uint32(l), false, true, true, false, dmaPWM); err != nil {
+		return err
+	}
+
+	// The first channel must be a full bandwidth one.
```
simokawa: @maruel Could you explain why?

host/bcm283x/dma.go
```diff
@@ -655,6 +659,60 @@ func allocateCB(size int) ([]controlBlock, *videocore.Mem, error) {
...
+		return err
+	}
+
+	defer pcmMemory.reset()
+	// Start transfer
+	pcmMemory.set()
+	runIO(pCB, l <= maxLite)
+	// We have to wait PCM to be finished even after DMA finished.
+	for pcmMemory.cs&pcmTXErr == 0 {
+		Nanospin(10 * time.Nanosecond)
```
maruel: In theory it should try to wait the exact right amount of time.

host/bcm283x/dma.go
maruel: Because the "light" ones are effectively a single one, which means that they are interleaved. If both are "light" then the jitter is largely increased.

host/bcm283x/dma.go
```diff
@@ -823,6 +1113,16 @@ func (d *driverDMA) Close() error {
 	return nil
 }
 
+func debugDMA() {
+	for i, ch := range dmaMemory.channels {
+		fmt.Println(i, ch.cs.String())
+		if ch.cs&dmaActive != 0 {
+			fmt.Printf("%x: %s", ch.cbAddr, ch.GoString())
```
maruel: I'd prefer log instead of fmt

simokawa: Fixed in https://github.com/google/periph/pull/209.

host/bcm283x/dma.go
simokawa: I experimented dual channel method but it seems that two channels easily get out of sync. I guess it DREQ clock can be missed.

host/bcm283x/dma.go
simokawa: Yeah, I just wanted avoid a busy loop. Maybe, we should wait 1 PCM clock.
dmaChannel.wait() is a busy loop now.

PR #208, opened 2017-12-27
onewire-list command added

libesz:
I created a simple and generic onewire device listing command. I found it helpful to test host implementations.

PR #214, opened 2018-01-03
cmd: make flag processing more coherent

maruel:
There's value in having tools in cmd/... all behave similarly. Change flag
processing to be slightly more similar.


maruel:
https://github.com/google/periph/pull/208 by @libesz made me realize that the flag processing is a bit inconsistent so I picked "one way" and made all tool except periph-smoketest behave similarly.

PR #220, opened 2018-02-27
Initial support for the Mifare MFRC522 RFID Reader

jdevelop:


experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+// that can be found in the LICENSE file.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"periph.io/x/periph/conn/spi/spireg"
+	rf522 "periph.io/x/periph/experimental/devices/mfrc522"
```
maruel: Either name the package rc522 or use mfrc522 in the file.

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
```
maruel: This is new code, use 2018.

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+	"periph.io/x/periph/host"
+	"strconv"
+	"strings"
+)
+
+var (
+	sector = flag.Int("sector", 1, "Sector to access")
+	block  = flag.Int("block", 0, "Block to access")
+
+	rsPin  = flag.String("rs-pin", "13", "Reset pin")
```
maruel: Do not use default pin names, we don't know if we run on an RPi.

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+	irqPin = flag.String("irq-pin", "12", "IRQ pin")
+
+	keyCommand   = flag.Bool("wa", false, "Overwrite keys")
+	blockCommand = flag.Bool("wb", false, "Overwrite block by [0-15]")
+
+	spiDevice = flag.String("spidev", "/dev/spidev0.0", "SPI device")
```
maruel: Remove default, the code will work correctly. It is not a "device", see the other tools for example, e.g. https://github.com/google/periph/blob/master/cmd/bmxx80/main.go#L75

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+
+	key = flag.String("key", "", "Comma-separated key bytes")
```
maruel: Move flags inside main(), like other tools.

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+func main() {
+
+	flag.Parse()
+
+	fmt.Println("Using pin settings: rsPin", *rsPin, "irqPin", *irqPin)
```
maruel: Remove.

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+	if _, err := host.Init(); err != nil {
+		log.Fatal(err)
+	}
+
+	currentAccessKey := rf522.DefaultKey
+
+	if *key != "" {
+		keyBytes := strings.SplitN(*key, ",", 6)
+		if len(keyBytes) != 6 {
+			log.Fatal("Key should consist of 6 decimal numbers")
```
maruel: return an error, with mainImpl() flow.

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+			if err != nil {
+				log.Fatal(err)
+			}
+			currentAccessKey[i] = byte(intV)
+		}
+	}
+
+	currentAccessMethod := byte(commands.PICC_AUTHENT1B)
+
+	// use BCM numbering here
```
maruel: Remove

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+	log.SetOutput(os.Stdout)
+
+	spiDev, err := spireg.Open(*spiDevice)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	rfid, err := rf522.NewSPI(spiDev, 1000000, *rsPin, *irqPin)
```
maruel: do not specify speed here, that's the driver that defines the maximum speed from the datasheet.

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	data, err := rfid.ReadCard(currentAccessMethod, *sector, *block, currentAccessKey[:])
```
maruel: [:] is unnecessary

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+		log.Fatal(err)
+	}
+	auth, err := rfid.ReadAuth(currentAccessMethod, *sector, currentAccessKey[:])
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	access := rf522.ParseBlockAccess(auth[6:10])
+
+	if err != nil {
```
maruel: ?

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,128 @@
...
+			*sector,
+			[6]byte{1, 2, 3, 4, 5, 6},
+			[6]byte{6, 5, 4, 3, 2, 1},
+			&rf522.BlocksAccess{
+				B0: rf522.RAB_WB_IB_DAB,
+				B1: rf522.RB_WB_IN_DN,
+				B2: rf522.AnyKeyRWID,
+				B3: rf522.KeyA_RN_WN_BITS_RAB_WN_KeyB_RN_WN,
+			},
+			currentAccessKey[:],
```
maruel: [:] is unnecessary.

experimental/devices/mfrc522/commands/picc.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+package commands
```
maruel: Keep an empty line between the copyright and the package name, otherwise this becomes the package's documentation.

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,764 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package mfrc522 controls a Mifare RFID card reader.
+// The code is largely ported from Python library : https://github.com/mxgxw/MFRC522-python
```
maruel: Keep an empty line after the first line.
I'd remove the python reference.
Use the title formatting. For example, https://github.com/google/periph/blob/master/devices/bmxx80/bmxx80.go#L5 becomes https://periph.io/x/periph/devices/bmxx80, which is nice.

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,764 @@
...
+// The datasheet is available at : https://www.nxp.com/docs/en/data-sheet/MFRC522.pdf
+package mfrc522
+
+import (
+	"errors"
+	"fmt"
+	"periph.io/x/periph/conn/gpio"
```
maruel: sort as what goimports would do; golang.org/x/tools/cmd/goimports

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,764 @@
...
+	"periph.io/x/periph/conn/gpio/gpioreg"
+	"periph.io/x/periph/conn/spi"
+	"periph.io/x/periph/experimental/devices/mfrc522/commands"
+	"time"
+)
+
+// Dev defines the RFID reader properties, the SPI bus and GPIO pins.
```
maruel: Dev is an handle to an MFRC522 RFID reader.

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
```
maruel: Make the source name the same name as the package name.

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+	"periph.io/x/periph/conn/gpio"
+	"periph.io/x/periph/conn/gpio/gpioreg"
+	"periph.io/x/periph/conn/spi"
+	"periph.io/x/periph/experimental/devices/mfrc522/commands"
+	"time"
+)
+
+// Dev defines the RFID reader properties, the SPI bus and GPIO pins.
+type Dev struct {
+	ResetPin      gpio.PinIO
```
maruel: Do no export, otherwise people can mess with it.
resetPin can probably be gpio.PinOut and irqPin a gpio.PinIn.

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+	MaxSpeedHz    int
+	spiDev        spi.Conn
+}
+
+// DefaultKey provides the default bytes for card authentication for method B.
+var DefaultKey = []byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF}
+
+// NewSPI creates and initializes the RFID card reader attached to SPI.
+// - spiPort - the SPI device to use.
+// - maxSpeed - maximum speed for SPI data transfers.
```
maruel: Remove maxspeed
resetPin and irqPin should be gpio.PinIn / PinOut instances

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+// - resetPin - reset GPIO pin.
+// - irqPin - irq GPIO pin.
+func NewSPI(spiPort spi.Port, maxSpeed int, resetPin, irqPin string) (device *Dev, err error) {
```
maruel: Do not use named return arguments, it actually makes the code longer.

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+
+	err = dev.Init()
+
+	device = dev
+
+	return
+}
+
+// Init initializes the RFID chip.
+func (r *Dev) Init() (err error) {
```
maruel: Same (about named arguments). Furthermore, there's a risk of aliasing which is easy to mess up.

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+	err = r.Reset()
+	if err != nil {
+		return
+	}
+	err = r.devWrite(commands.TModeReg, 0x8D)
```
maruel: Can't you batch up all these commands?

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+		return
+	}
+	err = r.SetAntenna(true)
+	if err != nil {
+		return
+	}
+	return
+}
+
+func (r *Dev) writeSpiData(dataIn []byte) (out []byte, err error) {
```
maruel: Inline this function.

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+		res = res + fmt.Sprintf("%02x, ", byte(v))
+	}
+	res = res + fmt.Sprintf("%02x", data[len(data)-1])
+	res = res + "]"
+	return
+}
+
+func (r *Dev) devWrite(address int, data byte) (err error) {
+	newData := [2]byte{(byte(address) << 1) & 0x7E, data}
+	_, err = r.writeSpiData(newData[:])
```
maruel: Here you can call Tx() with out being nil

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+	for _, v := range data[0 : len(data)-1] {
+		res = res + fmt.Sprintf("%02x, ", byte(v))
+	}
+	res = res + fmt.Sprintf("%02x", data[len(data)-1])
+	res = res + "]"
+	return
+}
+
+func (r *Dev) devWrite(address int, data byte) (err error) {
+	newData := [2]byte{(byte(address) << 1) & 0x7E, data}
```
maruel: no need for constant 2

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+	return
+}
+
+func (r *Dev) writeSpiData(dataIn []byte) (out []byte, err error) {
+	out = make([]byte, len(dataIn))
+	err = r.spiDev.Tx(dataIn, out)
+	return
+}
+
+func printBytes(data []byte) (res string) {
```
maruel: Remove

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+		return
+	}
+	err = r.devWrite(address, current&^byte(mask))
+	return
+
+}
+
+// SetAntennaGain initiates the antenna gain, values must be between 0 and 7.
+func (r *Dev) SetAntennaGain(gain int) (err error) {
+	if 0 <= gain && gain <= 7 {
```
maruel:
```
if gain < 0 || gain > 7 {
  return errors.New("mfrc522: invalid gain value")
}
r.antennaGain = gain
return
```

But I'm concerned that it doesn't send the command to the device, that's confusing.

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+		}
+	} else {
+		err = r.clearBitmask(commands.TxControlReg, 0x03)
+	}
+	return
+}
+
+// CardWrite the low-level interface to write some raw commands to the card.
+func (r *Dev) CardWrite(command byte, data []byte) (backData []byte, backLength int, err error) {
+	backData = make([]byte, 0)
```
maruel: `var backData []byte` will save an unnecessary memory allocation.

experimental/devices/mfrc522/rf522.go
```diff
@@ -0,0 +1,763 @@
...
+	}
+	_, err = r.SelectTag(uuid)
+	if err != nil {
+		return
+	}
+	return
+}
+
+// ReadCard reads the card sector/block.
+// - auth - the authentication mode.
```
maruel: Always keep an empty line to correctly format the documentation.

https://blog.golang.org/godoc-documenting-go-code is a good introduction.

maruel:
gohci

experimental/cmd/mfrc522/main.go
jdevelop: what is `mainImpl` flow?

experimental/cmd/mfrc522/main.go
jdevelop: Should it be moved inside the SPI driver?

experimental/cmd/mfrc522/main.go
maruel: Here's one of the simplest example: https://github.com/google/periph/blob/master/cmd/ir/main.go

This permits thinking in term of error, not exceptions and this makes the code easier to copy-paste for users.

experimental/cmd/mfrc522/main.go
maruel: The Connect call must be done in the device driver. The device driver accepts a spi.Port and converts it into a Conn.

I wrote a quick overview at https://twitter.com/periphio/status/968590975592468486

Doc for https://periph.io/x/periph/conn/spi#Port also notes this.

experimental/devices/mfrc522/rf522.go
jdevelop: It is used in `devRead` and `devWrite`, does it make sense to copy/paste it there?

experimental/devices/mfrc522/rf522.go
jdevelop: But from the docs:
```
	// For full duplex protocols (generally SPI, UART), the two buffers must have
	// the same length as both reading and writing happen simultaneously.
```

and the card will return the command/result as the first byte.

experimental/devices/mfrc522/rf522.go
maruel: devWrite doesn't use out, so it boils down to the direct Tx call.

experimental/devices/mfrc522/rf522.go
maruel: I agree the documentation is confusing. You can use nil on one of the two if you don't care about the data (or want to send zeros).

experimental/devices/mfrc522/rf522.go
jdevelop: Any call to RFID returns something, and it's full-duplex. So the buffer to receive the result should be of the same size as the input buffer. Even if I don't use the result - I still need to pass the buffer to fill in.

maruel: You don't. The bytes read will be discarded. Here's an example: https://github.com/google/periph/blob/master/host/sysfs/spi.go#L364

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if *key != "" {
+		keyBytes := strings.SplitN(*key, ",", 6)
+		if len(keyBytes) != 6 {
+			return errors.New("key should consist of 6 decimal numbers")
+		}
+		currentAccessKey = make([]byte, 6)
+
+		for i, v := range keyBytes {
+			intV, err := strconv.Atoi(v)
```
maruel: You should also make sure this doesn't overflow. The best way is to use https://golang.org/pkg/strconv/#ParseUint and pass 8 as bitSize.

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,131 @@
...
+			&mfrc522.BlocksAccess{
+				B0: mfrc522.RAB_WB_IB_DAB,
+				B1: mfrc522.RB_WB_IN_DN,
+				B2: mfrc522.AnyKeyRWID,
+				B3: mfrc522.KeyA_RN_WN_BITS_RAB_WN_KeyB_RN_WN,
+			},
+			currentAccessKey,
+		)
+		if err != nil {
+			log.Fatal(err)
```
maruel: return err
and line 119

experimental/devices/mfrc522/commands/registers.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package commands
+
+const (
+	Reserved00    = 0x00
```
maruel: Do not export the reserved values? In fact I don't see the value in having these at all.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,752 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package mfrc522 controls a Mifare RFID card reader.
+//
+// The code is largely ported from Python library : https://github.com/mxgxw/MFRC522-python
+//
+// The datasheet is available at : https://www.nxp.com/docs/en/data-sheet/MFRC522.pdf
```
maruel: What I meant is to use the following form, so that then you can do something like https://periph.io/x/periph/experimental/devices/mfrc522#Datasheet but maybe it's just because I'm easily amused. :)

//
// Datasheet
//
// https://www.nxp.com/docs/en/data-sheet/MFRC522.pdf

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,752 @@
...
+	irqPin        gpio.PinIn
+	Authenticated bool
+	maxSpeedHz    int64
+	spiDev        spi.Conn
+}
+
+// DefaultKey provides the default bytes for card authentication for method B.
+var DefaultKey = []byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF}
+
+const maxSpeed = int64(1000000)
```
maruel: Having a named constant is not useful since it's used exactly once, at the Connect() call. So please inline.

Also I miscalculated the value, on pages 2 and 10 it states it's 10000000.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,752 @@
...
+
+/*
```
maruel: Rewrite as:

// NewSPI creates and initializes the RFID card reader attached to SPI.
//
//    spiPort - the SPI device to use.
//    resetPin - reset GPIO pin.
//    irqPin - irq GPIO pin.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,752 @@
...
+NewSPI creates and initializes the RFID card reader attached to SPI.
+	spiPort - the SPI device to use.
+	resetPin - reset GPIO pin.
+	irqPin - irq GPIO pin.
+*/
+func NewSPI(spiPort spi.Port, resetPin, irqPin string) (*Dev, error) {
```
maruel: change to gpio.PinOut and gpio.PinIn

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,131 @@
...
+
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"periph.io/x/periph/conn/spi/spireg"
```
maruel: Please sort all the imports via goimports.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,752 @@
...
+	"periph.io/x/periph/experimental/devices/mfrc522/commands"
+	"time"
+)
+
+// Dev is an handle to an MFRC522 RFID reader.
+type Dev struct {
+	resetPin      gpio.PinOut
+	irqPin        gpio.PinIn
+	Authenticated bool
+	maxSpeedHz    int64
```
maruel: Remove it, not necessary.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,752 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	dev := &Dev{
+		spiDev:     spiDev,
+		maxSpeedHz: maxSpeed,
+	}
+
+	pin := gpioreg.ByName(resetPin)
```
maruel: Do not use gpioreg here.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,752 @@
...
+	dev.resetPin = pin
+	dev.resetPin.Out(gpio.High)
+
+	pin = gpioreg.ByName(irqPin)
+	dev.irqPin = pin
+	dev.irqPin.In(gpio.PullUp, gpio.FallingEdge)
+
+	err = dev.Init()
```
maruel: err is ignored;

if err = dev.Init(); err != nil {
  return nil, err
}
return dev, nil

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,752 @@
...
+	{commands.TPrescalerReg, 0x3E},
+	{commands.TReloadRegL, 30},
+	{commands.TReloadRegH, 0},
+	{commands.TxAutoReg, 0x40},
+	{commands.ModeReg, 0x3D},
+}
+
+// Init initializes the RFID chip.
+func (r *Dev) Init() error {
+	err := r.Reset()
```
maruel: if err := r.Reset(); err != nil {

Do the same at lines 87 and 92, which will save 3 lines and make the function looks much more trivial.

Since this device is effectively a memory mapped register, you can coalesce the commands TModeReg, TPrescalerReg, TReloadRegH, TReloadRegL as a single write. The problem is that you call them "commands" but they are effectively memory addresses.

So basically, initCommands would look like:

```
var initCommands = [][]byte{
  // Mode, Prescaler, ReloadH, ReloadL
  {0x2A, 0x8B, 0x3E, 0x00, 0x30},
  {commands.TxAutoReg, 0x40},
  {commands.ModeReg, 0x3D},
}
```

Makes sense?

More reference at https://periph.io/x/periph/conn/mmr

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,767 @@
...
+		}
+	}
+	err = r.SetAntenna(true)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *Dev) devWrite(address int, data byte) (err error) {
```
maruel: Do not use named return values unless you need to modify the return value in a defer function. Here and below.

In this case, the whole function reduces to one or two lines.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,767 @@
...
+		if current&0x03 == 0 {
+			err = r.setBitmask(commands.TxControlReg, 0x03)
+		}
+	} else {
+		err = r.clearBitmask(commands.TxControlReg, 0x03)
+	}
+	return
+}
+
+/*
```
maruel: Use // style everywhere.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,767 @@
...
+			}
+			backData[i] = byteVal
+		}
+
+	}
+
+	return
+}
+
+// Request request the card information. Returns number of blocks available on the card.
```
maruel: // Request requests the card information.

(yeah I know)
But still, that's kind of fuzzy what this means.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,767 @@
...
+	}
+
+	return
+}
+
+// Wait wait for IRQ to strobe on the IRQ pin when the card is detected.
+func (r *Dev) Wait() (err error) {
+	irqChannel := make(chan bool)
+	go func() {
+		if r.irqPin.WaitForEdge(20 * time.Second) {
```
maruel: Why 20s?

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,767 @@
...
+
+	if backLen == 0x18 {
+		blocks = backData[0]
+	} else {
+		blocks = 0
+	}
+	return
+}
+
+type AuthStatus byte
```
maruel: Needs docstring if exported.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,767 @@
...
+		return
+	}
+	err = r.Reset()
+	return
+}
+
+func (r *Dev) String() string {
+	return fmt.Sprintf("Mifare MFRC522 [bus: %v, reset pin: %s, irq pin: %s, max speed :%d]",
+		r.spiDev, r.resetPin.Name(), r.irqPin.Name(), maxSpeed)
+}
```
maruel: Add a interface check like:

var _ conn.Resource = &Dev{}
var _ fmt.Stringer = &Dev{}

Example: https://github.com/google/periph/blob/master/devices/lepton/lepton.go#L457

experimental/devices/mfrc522/mfrc522.go
jdevelop: It seems that using the block comment with tabs results in exactly the same output in `godoc`

experimental/devices/mfrc522/mfrc522.go
jdevelop: Why do you think that it is memory-mapped?

experimental/devices/mfrc522/mfrc522.go
jdevelop: Will move that to the configuration.

experimental/devices/mfrc522/mfrc522.go
maruel: Oh I mean to consistently use // and not /* */ but the markdown malformed what I wrote above.

experimental/devices/mfrc522/mfrc522.go
maruel: See pages 10~11 of the datasheet.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,767 @@
...
+	return
+}
+
+//Halt implements Resource.Halt
+func (r *Dev) Halt() (err error) {
+	err = r.StopCrypto()
+	if err != nil {
+		return
+	}
+	err = r.Reset()
```
maruel: You don't want to send a Reset(), you want to send a soft power down as explained at page 33.

experimental/devices/mfrc522/mfrc522.go
maruel: Pages 36~37 present the Register addresses and each register is described up to page 69. In this parlance, register ~= memory. This is not "commands" per se, but writing to a register may induce an operation, which is alike sending a command. But strictly speaking, you are simply writing a value in a registry at the specified address.

Formal commands are described at page 70~73 and they are executed by writing to the pseudo-register "CommandReg". This is the only real "command" that there is.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,767 @@
...
+	if n&0x08 != 0 {
+		authS = AuthFailure
+	}
+	authS = AuthOk
+	return
+}
+
+// StopCrypto stops the crypto chip.
+func (r *Dev) StopCrypto() (err error) {
+	err = r.clearBitmask(commands.Status2Reg, 0x08)
```
maruel: These constants should reference the datasheet page and section, in this case page 43, section 9.3.1.9. That's much more useful than named constants IMHO.

This can be done in a follow up.

experimental/devices/mfrc522/mfrc522.go
jdevelop: I think that it's better to postpone this update for the next iteration, because I'll need some more time to figure out how to deal with the memory mapping registers in this case. I'll focus on making the code more "standard", and then we can get back to this topic once I finish with the rest. The device is functional and works just fine with the current way of the communication.

will that work?

maruel: Yes this is fine.

experimental/cmd/mfrc522/main.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	currentAccessMethod := byte(commands.PICC_AUTHENT1B)
+
+	spiDev, err := spireg.Open(*spiID)
+	if err != nil {
+		return err
+	}
+
+	rfid, err := mfrc522.NewSPI(spiDev, gpioreg.ByName(*rsPin), gpioreg.ByName(*irqPin))
```
maruel: In practice, you should confirm that ByName() call didn't return nil, as you need to inform the user that the pin name doesn't exist.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,732 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package mfrc522 controls a Mifare RFID card reader.
+//
+// The code is largely ported from Python library : https://github.com/mxgxw/MFRC522-python
```
maruel: Is still the case? The goal of periph drivers is to not just be ports but really look at the datasheet and implement what matters.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,732 @@
...
+// 	resetPin - reset GPIO pin.
+// 	irqPin - irq GPIO pin.
+func NewSPI(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn) (*Dev, error) {
+
+	spiDev, err := spiPort.Connect(10000000, spi.Mode0, 8)
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
```
maruel: oops?

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,732 @@
...
+		return nil, err
+	}
+
+	dev := &Dev{
+		spiDev:           spiDev,
+		operationTimeout: 30 * time.Second,
+	}
+
+	dev.resetPin = resetPin
+	dev.resetPin.Out(gpio.High)
```
maruel: Do the calls at lines 58 and 61 right after line 47, and if either return an error, return it.
Then set the variables at Dev{} construction.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,732 @@
...
+
+	dev.irqPin = irqPin
+	dev.irqPin.In(gpio.PullUp, gpio.FallingEdge)
+
+	err = dev.Init()
```
maruel: return nil if err != nil

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,732 @@
...
+
+// SetAntenna configures the antenna state, on/off.
+func (r *Dev) SetAntenna(state bool) error {
+	if state {
+		current, err := r.devRead(commands.TxControlReg)
+		if err != nil {
+			return err
+		}
+		if current&0x03 != 0 {
+			return errors.New("Can not set the bitmask for antenna")
```
maruel: Eventually you'll want to add a wrapf() function that wraps all errors with a "mfrc522:" prefix, so when the error is surfaced to the user the user can figure out where the error comes from. You can grep for "wrapf" in the code base for examples.

experimental/devices/mfrc522/commands/picc.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package commands
+
+// Operation coinstants
```
maruel: constants

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,732 @@
...
+		r.devWrite(commands.FIFODataReg, v)
+	}
+
+	r.devWrite(commands.CommandReg, command)
+
+	if command == commands.PCD_TRANSCEIVE {
+		r.setBitmask(commands.BitFramingReg, 0x80)
+	}
+
+	i := 2000
```
maruel: That is very adhoc and will have to be rewritten. IIUC, it should wait for the irqPin, no need to pool.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,732 @@
...
+	if err != nil {
+		return AuthReadFailure, err
+	}
+	if n, err := r.devRead(commands.Status2Reg); err != nil || n&0x08 == 0 {
+		return AuthFailure, err
+	}
+	return AuthOk, nil
+}
+
+// StopCrypto stops the crypto chip.
```
maruel: Does it need to be exported?

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,732 @@
...
+
+func (r *Dev) write(blockAddr byte, data []byte) error {
+	read, backLen, err := r.preAccess(blockAddr, commands.PICC_WRITE)
+	if err != nil || backLen != 4 {
+		return err
+	}
+	if read[0]&0x0F != 0x0A {
+		return errors.New("can't authorize write")
+	}
+	newData := make([]byte, 18)
```
maruel: var newData [18]byte
no need to forcibly put it in the heap

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,732 @@
...
+	if err != nil {
+		return err
+	}
+	if backLen != 4 || read[0]&0x0F != 0x0A {
+		err = errors.New("can not write data")
+	}
+	return nil
+}
+
+func calcBlockAddress(sector int, block int) (addr byte) {
```
maruel: remove named return arguments everywhere, this code is just more opaque than the unnamed way.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,732 @@
...
+func ParseBlockAccess(ad []byte) *BlocksAccess {
+	ba := new(BlocksAccess)
+	ba.B0 = BlockAccess(ad[1]&0x10>>4 | ad[2]&0x01<<1 | ad[2]&0x10>>2)
+	ba.B1 = BlockAccess(ad[1]&0x20>>5 | ad[2]&0x02 | ad[2]&0x20>>3)
+	ba.B2 = BlockAccess(ad[1]&0x40>>6 | ad[2]&0x04>>1 | ad[2]&0x40>>4)
+	ba.B3 = SectorTrailerAccess(ad[1]&0x80>>7 | ad[2]&0x08>>2 | ad[2]&0x80>>5)
+	return ba
+}
+
+func (r *Dev) String() string {
```
maruel: I'll ask you to move code around in this form:

- All public utility types like BlocksAccess and AuthStatus
- Dev struct
- Dev public methods
- Dev private methods
- Other private methods

experimental/devices/mfrc522/mfrc522.go
jdevelop: This is the code I derived from the python library, so I'm not sure whether it's ad-hoc or not.

experimental/devices/mfrc522/mfrc522.go
jdevelop: I think yes, for some use-cases. Or we can move the refinement of the public API to the next iteration.

experimental/devices/mfrc522/mfrc522.go
jdevelop: May be it could be done on the file level, to separate the public API from the private one:
* mfrc522_public.go
* mfrc522_private.go

Will that work?

experimental/cmd/mfrc522/main.go
jdevelop: I think that it could be wise to check that the passed pin references are not nil inside the `NewSPI` method too, to make it a bit more defensive.

experimental/devices/mfrc522/mfrc522.go
jdevelop: Ok, will remove :)

experimental/devices/mfrc522/mfrc522.go
jdevelop: okay, makes sense

experimental/devices/mfrc522/mfrc522.go
maruel: The file is not long enough to be worth it.

experimental/devices/mfrc522/mfrc522.go
maruel: I'd speculate it is. :) Let's revisit this later.

experimental/cmd/mfrc522/main.go
maruel: No driver does that. We can't start checking for nil for the pins but not the spi bus as it would be inconsistent. Then we'd have to start checking for nil everywhere on every driver. I prefer the default that passing nil is never supported.

experimental/devices/mfrc522/mfrc522.go
jdevelop: yep, removed

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,757 @@
...
+	return fmt.Sprintf("Mifare MFRC522 [bus: %v, reset pin: %s, irq pin: %s]",
+		r.spiDev, r.resetPin.Name(), r.irqPin.Name())
+}
+
+///////////////////////////////////////////////////////////////////////////////////////////
```
maruel: Not a fan of these, I agree something like:

```
// Dev private/helper functions
```

is helpful to categorize the code, but no need to spend 5 lines on this.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,757 @@
...
+	}
+	if state != AuthOk {
+		err = wrapf("authentication failed")
+		return
+	}
+
+	return r.write(calcBlockAddress(sector, block%3), data[:])
+}
+
+/*
```
maruel: Convert to // so all comments have consistent style.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,757 @@
...
+// WriteBlock writes the data into the card block.
+//
+// 	auth - the authentiction mode.
+// 	sector - the sector on the card to write to.
+// 	block - the block within the sector to write into.
+// 	data - 16 bytes if data to write
+// 	key - the key used to authenticate the card - depends on the used auth method.
+func (r *Dev) WriteBlock(auth byte, sector int, block int, data [16]byte, key []byte) (err error) {
+	defer func() {
+		if err == nil {
```
maruel: That's a value use of named return argument :)

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,757 @@
...
+			err = r.StopCrypto()
+		}
+	}()
+	uuid, err := r.selectCard()
```
maruel: I'm always frightened here as to if err is aliased or not.

experimental/devices/mfrc522/commands/registers.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package commands
+
+// Command register constants
```
maruel: Technically, it is just Register contants, only one of all of them is a comment register. :)

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,757 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	dev := &Dev{
+		spiDev:           spiDev,
+		operationTimeout: 30 * time.Second,
+	}
+
+	if err := resetPin.Out(gpio.High); err != nil {
```
maruel: Move these before constructing Dev, so that you can use:

```
dev := &Dev{
  spiDev:           spiDev,
  resetPin: resetPin,
  irqPin: irqPin,
  operationTimeout: 30 * time.Second,
}
```

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,757 @@
...
+	switch command {
+	case commands.PCD_AUTHENT:
+		irqEn = 0x12
+		irqWait = 0x10
+	case commands.PCD_TRANSCEIVE:
+		irqEn = 0x77
+		irqWait = 0x30
+	}
+
+	r.devWrite(commands.CommIEnReg, irqEn|0x80)
```
maruel: You forgot to check errors here.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -0,0 +1,757 @@
...
+		n, err := r.devRead(commands.CommIrqReg)
+		if err != nil {
+			return nil, -1, err
+		}
+		if n&(irqWait|1) != 0 {
+			break
+		}
+	}
+
+	r.clearBitmask(commands.BitFramingReg, 0x80)
```
maruel: still here

I know it's tedious, but after 10+ years of python, I still prefer to check manually than to have to catch random exceptions.

maruel:
gohci

PR #223, opened 2018-03-20
Hitachi HD44780U Dot Matrix Liquid Crystal Display

jdevelop:


experimental/cmd/hd44780/main.go
```diff
@@ -0,0 +1,93 @@
...
+	rsPinReg := gpioreg.ByName(*rsPin)
+	if rsPinReg == nil {
+		return fmt.Errorf("Register select pin %s can not be found", *rsPin)
+	}
+	ePinReg := gpioreg.ByName(*ePin)
+	if ePin == nil {
+		return fmt.Errorf("Strobe pin %s can not be found", *ePin)
+	}
+
+	dataPins := make([]gpio.PinOut, 4)
```
maruel: var dataPins [4]gpio.PinOut ? There's no need to put the array in the heap.

experimental/cmd/hd44780/main.go
```diff
@@ -0,0 +1,93 @@
...
+	for i, pinName := range pinsStr {
+		tmp := gpioreg.ByName(pinName)
```
maruel: if dataPins[i] = gpioreg.ByName(pinName); dataPins[i] == nil {

experimental/cmd/hd44780/main.go
```diff
@@ -0,0 +1,93 @@
...
+		if tmp == nil {
+			return fmt.Errorf("Data pin %s can not be found", tmp)
+		}
+		dataPins[i] = tmp
+	}
+
+	rpi, err := lcd.NewLCD4Bit(dataPins, rsPinReg, ePinReg)
```
maruel: why "rpi" ?

experimental/cmd/hd44780/main.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+
+	if *text == "" {
+		return nil
+	}
+
+	strs := strings.Split(*text, "\n")
+
+	for i := 0; i < len(strs) && i < 2; i++ {
```
maruel: Why not have a function that takes [][]rune and prints the whole buffer at once?

Actually, Print does it already (?)

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
```
maruel: merge this file into hd44780.go

experimental/devices/hd44780/lcd.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
```
maruel: rename to hd44780.go

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
...
+import (
+	"time"
+
+	"periph.io/x/periph/conn/gpio"
+)
+
+// LineTwo offset for the second line in the LCD buffer.
+const LineTwo = 0x40
+
+// LCD4 is the 4-bit addressing device for HD-44780
```
maruel: Rename it to Dev

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
...
+// https://www.sparkfun.com/datasheets/LCD/HD44780.pdf
+package hd44780
+
+import (
+	"time"
+
+	"periph.io/x/periph/conn/gpio"
+)
+
+// LineTwo offset for the second line in the LCD buffer.
```
maruel: Does that needs to be exported?

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
...
+	// enable pin
+	enablePin gpio.PinOut
+}
+
+// NewLCD4Bit creates and initializes the LCD device
+//	data - references to data pins
+//	rs - rs pin
+//	e - strobe pin
+func NewLCD4Bit(data []gpio.PinOut, rs, e gpio.PinOut) (*LCD4, error) {
+	dev := LCD4{
```
maruel: - you do not verify the length of data
- make dev a pointer right away, since it's meant to be returned

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
...
+		enablePin: e,
+		rsPin:     rs,
+	}
+	if err := dev.Init(); err != nil {
+		return nil, err
+	}
+	return &dev, nil
+}
+
+var resetSequence = [][]uint{
```
maruel: move all private symbols down the file, it makes it easier to scan the source with public first, private after.

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
...
+	{0x10, 0},    // disable display
+	{0x01, 2000}, // clear screen
+	{0x06, 0},    // cursor shift right, no display move
+	{0x0c, 0},    // enable display no cursor
+	{0x01, 2000}, // clear screen
+	{0x02, 2000}, // cursor home
+}
+
+// Init initializes the HC-44780 chipset, clears the screen buffer and moves cursor to the
+// home of screen (line 0, column 0)
```
maruel: End with a period.

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
...
+	delayMs(15)
+
+	if err := r.rsPin.Out(gpio.Low); err != nil {
+		return err
+	}
+	if err := r.enablePin.Out(gpio.Low); err != nil {
+		return err
+	}
+
+	wrapper := func(seq [][]uint, f func(_lcd *LCD4, _data uint8) error) error {
```
maruel: make it a real external function instead of a local one

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
...
+
+	return nil
+}
+
+// Cls clears the LCD screen
+func (r *LCD4) Cls() error {
+	if err := writeInstruction(r, 0x01); err != nil {
+		return err
+	}
+	delayMs(2)
```
maruel: Is the delay needed? Instead you could note the last timestamp of the command inside Dev and then wait for a delay on every function call, e.g.
```
type Dev struct {
 ...
  waitUntil time.Time
}

func (d *Dev) Clear() error {
  if s := time.Until(d.waitUntil); s > 0 {
    time.Sleep(s)
  }
  ...
  d.waitUntil = time.Now().Add(time.Millisecond*2)
  return nil
}
```

and in all the public methods, do

```
  if s := time.Until(d.waitUntil); s > 0 {
    time.Sleep(s)
  }
```

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
...
+		return err
+	}
+	return nil
+}
+
+// Print the data string
+//	data string to display
+func (r *LCD4) Print(data string) error {
+	for _, v := range []byte(data) {
+		if err := r.WriteChar(v); err != nil {
```
maruel: Is it possible to write the whole thing at once? I'm pretty sure it is.

In fact, I think a more interesting function would be:

```
func (d* Dev) WriteChars(y, x int, s string) error {
}
```

and remove SetCursor(), Print() and WriteChar().

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
...
+	if err := write4Bits(r, data); err != nil {
+		return err
+	}
+	delayUs(10)
+	return nil
+}
+
+// service methods
+
+func clearBits(r *LCD4) error {
```
maruel: You should make these methods.

experimental/devices/hd44780/lcd.go
```diff
@@ -0,0 +1,23 @@
...
+// that can be found in the LICENSE file.
+
+// Package hd44780 controls the Hitachi LCD display chipset HD-44780
+//
+// Datasheet
+//
+// https://www.sparkfun.com/datasheets/LCD/HD44780.pdf
+package hd44780
+
+// LCD interface to interact with the controller.
```
maruel: This interface is not used anywhere, I'd remove it as I don't see value.

experimental/devices/hd44780/lcd_hd44780_4bit.go
```diff
@@ -0,0 +1,230 @@
...
+	return nil
+}
+
+func delayUs(ms uint) {
+	time.Sleep(time.Duration(ms) * time.Microsecond)
+}
+
+func delayMs(ms int) {
+	time.Sleep(time.Duration(ms) * time.Millisecond)
+}
```
maruel: At the end, add:

```
var _ conn.Resource = &Dev{}
var _ fmt.Stringer = &Dev{}
```

and make sure it compiles. :)

experimental/devices/hd44780/lcd_hd44780_4bit.go
jdevelop: I double checked the specs, and it looks like you can't send the sequence of the characters to the internal RAM, since the programming interface is limited to one register. The datasheet contains an example on the page 40 that makes it clear - every symbol must be transferred along with the separate command request (setting up Rs/E pins and then performing the transfer of 2 4-bit sequences). Perhaps this WriteChar method has to be moved to the private interface.

jdevelop: Also I think that `SetCursor` could be useful as well, because the chip will keep track of the current position and then advance the chars w/o the need for a user to keep this counter internally.

experimental/cmd/hd44780/main.go
```diff
@@ -0,0 +1,91 @@
...
+	"os"
+	"strings"
+
+	"periph.io/x/periph/conn/gpio"
+	"periph.io/x/periph/conn/gpio/gpioreg"
+	"periph.io/x/periph/experimental/devices/hd44780"
+	"periph.io/x/periph/host"
+)
+
+var pinPattern = "no %s pin specified. Please provide the pin via '%s' flag, for example '%s'"
```
maruel: Make it a const inside mainFunc() around line 33, since it's not used elsewhere.

experimental/cmd/hd44780/main.go
```diff
@@ -0,0 +1,91 @@
...
+			return fmt.Errorf("Data pin %s can not be found", pinName)
+		}
+	}
+
+	dev, err := hd44780.New(dataPins[:], rsPinReg, ePinReg)
+	if err != nil {
+		return err
+	}
+
+	if *text == "" {
```
maruel: In this case, I'd expect the tool to at least clear out the screen?

experimental/devices/hd44780/hd44780.go
```diff
@@ -0,0 +1,261 @@
...
+		enablePin: e,
+		rsPin:     rs,
+	}
+	if err := dev.Init(); err != nil {
+		return nil, err
+	}
+	return dev, nil
+}
+
+// Init initializes the HC-44780 chipset, clears the screen buffer and moves cursor to the
```
maruel: No need to export this function.

experimental/devices/hd44780/hd44780.go
```diff
@@ -0,0 +1,261 @@
...
+// Halt will re-initialize the display.
+func (r *Dev) Halt() error {
+	return r.Init()
+}
+
+func (r *Dev) String() string {
+	return "HD44870, 4 bit mode"
+}
+
+// Cls clears the LCD screen
```
maruel: This should be Halt().

experimental/devices/hd44780/hd44780.go
```diff
@@ -0,0 +1,261 @@
...
+// Dev is the 4-bit addressing device for HD-44780
+type Dev struct {
+	// data pins
+	dataPins []gpio.PinOut
+
+	// register select pin
+	rsPin gpio.PinOut
+
+	// enable pin
+	enablePin gpio.PinOut
```
maruel: I think the driver should keep track of the current cursor position, so it can skip updating the cursor when unnecessary.

You said the chip keep track of the current position, so you should leverage this by keeping state here too.

experimental/devices/hd44780/hd44780.go
```diff
@@ -0,0 +1,261 @@
...
+func (r *Dev) Print(data string) error {
+	for _, v := range []byte(data) {
+		if err := r.WriteChar(v); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// PrintAt the data string at specified position
```
maruel: I don't see the need of exporting SetCursor(), Print() and WriteChar() with this function.

I'd recommend to explicitly return an error when the string overflows the known boundaries.

experimental/devices/hd44780/hd44780.go
```diff
@@ -0,0 +1,261 @@
...
+	return nil
+}
+
+// Halt will re-initialize the display.
+func (r *Dev) Halt() error {
+	return r.Init()
+}
+
+func (r *Dev) String() string {
+	return "HD44870, 4 bit mode"
```
maruel: Do you think it's worth implementing 8 bits mode?

experimental/devices/hd44780/hd44780.go
```diff
@@ -0,0 +1,261 @@
...
+	return nil
+}
+
+// PrintAt the data string at specified position
+//	data string to display
+func (r *Dev) PrintAt(row, column uint8, data string) error {
+	if err := r.SetCursor(row, column); err != nil {
+		return err
+	}
+	for _, v := range []byte(data) {
```
maruel: That's incorrect. In Go, string represents an UTF-8 encoded string. You need to report unsupported characters and manually translate the supported ones into the code page.

This is particularly important as there's multiple chips with different code pages (HCD44780UA00 vs HCD44780UA02), and it's valuable to support custom characters.

I can't find in the datasheet a way to identify the code page of the chip. (?)

experimental/devices/hd44780/hd44780.go
```diff
@@ -0,0 +1,261 @@
...
+
+	// enable pin
+	enablePin gpio.PinOut
+}
+
+// New creates and initializes the LCD device
+//	data - references to data pins
+//	rs - rs pin
+//	e - strobe pin
+func New(data []gpio.PinOut, rs, e gpio.PinOut) (*Dev, error) {
```
maruel: Number of lines should be provided as an argument (1 or 2).

experimental/devices/hd44780/hd44780.go
jdevelop: The logic here is: you can use `SetCursor` to position the cursor at certain point, then you can call arbitrary number of `Print` to update the content of string. No need to track current cursor position inside the driver.

Additionally, you can use `PrintAt` to specify the exact location of the text to be printed on the screen. Again, I don't see a value to track the cursor inside the driver. If a client app needs to have something sophisticated - they can do it internally.

experimental/devices/hd44780/hd44780.go
jdevelop: If the cursor is out of range - nothing will happen (at least with the current device I have for testing). So I presume that the invalid addresses are ignored.

experimental/devices/hd44780/hd44780.go
jdevelop: I don't think so, it requires too much of pins. However it could be helpful to notify the end user that we use 4 bit mode for now. I don't see any advantage in 8-bit mode as for now.

experimental/devices/hd44780/hd44780.go
jdevelop: AFAIK there's no way to tell which codepage the chip has. You can order the chip with your own codepage. 

Still you can upload up to 8 of custom character patterns and use those.

From this point, I don't see any value in working with runes here, `byte` seems to be a reasonable primitive - the end users may experiment with the encodings on the client side.

experimental/devices/hd44780/hd44780.go
jdevelop: I don't think that it makes much difference here, there's nothing related to the line calculation in the code, and the client knows what exactly she has for the hardware side. Especialy it is not possible to figure out whether the device supports 1 line or 2, so the extra validation makes no sense.

maruel: The application author and the end user are often different persons. If validation is skipped, it'd be good to note this in the documentation.

experimental/devices/hd44780/hd44780.go
maruel: Please document this.

experimental/devices/hd44780/hd44780.go
maruel: Clarifying my idea:

```
type CodePage map[rune]string

var A00 = CodePage{...}
var A02 = CodePage{...}

func New(data []gpio.PinOut, rs, e gpio.PinOut, cp CodePage) (*Dev, error) {
...
```

Then here you can do:

```
for _, r := range data {
  b, ok := d.cp[r]
  if !ok {
     c = ' '
   }
   ....
```

Then in the doc, add:

// Unknown characters are printed as whitespace

experimental/devices/hd44780/hd44780.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	if err := r.enablePin.Out(gpio.Low); err != nil {
+		return err
+	}
+
+	if err := r.bulkSendData(resetSequence, r.write4Bits); err != nil {
+		return err
+	}
+
+	if err := r.bulkSendData(initSequence, r.writeInstruction); err != nil {
```
maruel: return r.bulkSendData(initSequence, r.writeInstruction)

experimental/devices/hd44780/hd44780.go
```diff
@@ -0,0 +1,256 @@
...
+// New creates and initializes the LCD device
+//	data - references to data pins
+//	rs - rs pin
+//	e - strobe pin
+func New(data []gpio.PinOut, rs, e gpio.PinOut) (*Dev, error) {
+	if len(data) != 4 {
+		return nil, fmt.Errorf("expected 4 data pins, passed %s", len(data))
+	}
+	dev := &Dev{
+		dataPins:  data,
```
maruel: I'm wondering if we should play defensive here and create a local slice. Otherwise if the caller alter data after this function call, it messes up the code here. A way to do this is to replace line 26 as an array instead:

```
dataPins [4]gpio.PinOut
```

and do

```
copy(d.dataPins, data)
```

experimental/devices/hd44780/hd44780.go
```diff
@@ -0,0 +1,256 @@
...
+// Init initializes the HC-44780 chipset, clears the screen buffer and moves cursor to the
+// home of screen (line 0, column 0).
+func (r *Dev) Init() error {
+	if err := r.clearBits(); err != nil {
+		return err
+	}
+
+	delayMs(15)
+
+	if err := r.rsPin.Out(gpio.Low); err != nil {
```
maruel: Instead of the sendData and sendInstruction and here, you should do:
```
func (r *Dev) setCtlLines(r, e gpioLevel) error {
	if err := r.rsPin.Out(r); err != nil {
		return err
	}
	return r.enablePin.Out(e)
}
```

and call this function instead with a comment at call site about why the chosen levels were chosen.

experimental/devices/hd44780/hd44780.go
jdevelop: According to https://en.wikipedia.org/wiki/Hitachi_HD44780_LCD_controller
* HD44780UA00 - the standard (Japanese) version, which includes katakana characters and some Greek letters and mathematical symbols
* HD44780UA02 - a European version, which includes the complete set of Greek, Cyrillic and Western European characters (with diacritics)

Now on the datasheet: https://www.sparkfun.com/datasheets/LCD/HD44780.pdf, page 17 table 4 there is the list of stuff that is supported by this chipset.

Do seriously expect to have all the symbols enumerated in the code and that check to be implemented? I'm just curious.

maruel: I do, but that's not a blocker for this CL.

maruel:
Why did you close the CL?

jdevelop:
I feel like it will be much more productive for me to focus on the tasks I need in my projects while using some of the great codebase you have provided. So I will continue with my fork, and if you feel like there's something you'd like to see in the mainstream - feel free to cherry-pick.

maruel:
Ok, I think the code is good, so I'll merge it as-is so it can be iterated upon later.

maruel:
gohci

maruel:
Humm not sure why gohci didn't trigger, but travis worked.

PR #224, opened 2018-03-25
Fixed MFRC522 trailing sector overflow

jdevelop:
If the passed key is longer than 6 bytes, it will corrupt the access
bits for the traling sector.

Fixed the bit calculation, updates tests.

Changed signatures to accept 6-byte long arrays instead of slices.

Minor typo for HD44780

Signed-off-by: Eugene Dzhurinsky <jdevelop@gmail.com>

### DIFF
--- a/experimental/devices/mfrc522/mfrc522.go
+++ b/experimental/devices/mfrc522/mfrc522.go
@@ -497,7 +497,6 @@ func (r *Dev) WriteSectorTrail(auth byte, sector int, keyA [6]byte, keyB [6]byte
 	accessData := CalculateBlockAccess(access)
 	copy(data[6:], accessData[:4])
 	copy(data[10:], keyB[:])
-	fmt.Printf("Data bytes %v\n", data)
 	return r.write(calcBlockAddress(sector&0xFF, 3), data[:])
 }
 
--- a/experimental/devices/mfrc522/mfrc522.go
+++ b/experimental/devices/mfrc522/mfrc522.go
@@ -63,8 +63,12 @@ type BlocksAccess struct {
 	B3         SectorTrailerAccess
 }
 
+func (ba *BlocksAccess) String() string {
+	return fmt.Sprintf("B0: %d, B1: %d, B2: %d, B3: %d", ba.B0, ba.B1, ba.B2, ba.B3)
+}
+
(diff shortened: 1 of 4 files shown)


maruel:
gohci

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -432,7 +436,7 @@ func (r *Dev) ReadBlock(sector int, block int) ([]byte, error) {
 // 	block - the block within the sector to write into.
 // 	data - 16 bytes if data to write
 // 	key - the key used to authenticate the card - depends on the used auth method.
-func (r *Dev) WriteBlock(auth byte, sector int, block int, data [16]byte, key []byte) (err error) {
+func (r *Dev) WriteBlock(auth byte, sector int, block int, data [16]byte, key [6]byte) (err error) {
```
maruel: I'm ambivalent about this, because this copy the whole array at the invocation.

But copying a slice is copying 3 pointers, so in practice it's 12 or 24 bytes anyway. ¯\_(ツ)_/¯

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -488,12 +492,13 @@ func (r *Dev) WriteSectorTrail(auth byte, sector int, keyA [6]byte, keyB [6]byte
...
 
-	data := make([]byte, 16)
-	copy(data, keyA[:])
+	var data [16]byte
+	copy(data[:], keyA[:])
 	accessData := CalculateBlockAccess(access)
 	copy(data[6:], accessData[:4])
 	copy(data[10:], keyB[:])
-	return r.write(calcBlockAddress(sector&0xFF, 3), data)
+	fmt.Printf("Data bytes %v\n", data)
```
maruel: Left over?

jdevelop: yeah

experimental/devices/mfrc522/mfrc522.go
jdevelop: Here I wanted to ensure in the signature of the function that the user must send exactly 16 bytes. It might be a good idea to revert to slices and capture things at runtime, but from my perspective it's better to handle that at the compile time.

Perhaps I should pass a pointer to an array?

Issue #201, opened 2017-12-22
beagle: implement header tables for beaglebone/beagleboard board family


maruel:
It will be useful coupled with am335x support (issue #199) and PRU-ICSS support (issue #200).

https://beagleboard.org/

bodokaiser:
For the raspberry pi you [set up the pin layout](https://github.com/google/periph/blob/master/host/rpi/rpi.go#L50) with [bcm283x.GPIO](https://github.com/google/periph/blob/master/host/bcm283x/gpio.go#L25).

For the Beaglebone (Black) the analogue would be the am335x?

maruel:
Yes, there's scaffolding at http://periph.io/x/periph/experimental/host/am335x and http://periph.io/x/periph/experimental/host/beagle.

In practice a full am33x implementation is not initially needed, only detecting the presence. But even so, I think there's beaglebones on different processors (?) so detecting the board is what is important here, and registering the right header.

bodokaiser:
I only used the Beaglebone Black so far. Detection is done rudimentary by checking sysfs files?

How do you plan to support pin configuration (as we discussed on slack)? We could try to mimic the `config-pin` tool which I think just interfaces sysfs to find the current pin configuration, however this would make things more complicated. Alternatively we could just allow the user to overwrite specific pins?

maruel:
Detection should be done via `/proc/device-tree`.

Let's aim to have the header definition first, and add magicness later.

maruel:
Done (enough for now)

PR #229, opened 2018-04-13
add support for the BeagleBone black

bodokaiser:
See #201. Awaiting feedback.

### DIFF
--- a/experimental/host/beagle/beagle.go
+++ b/experimental/host/beagle/beagle.go
@@ -6,9 +6,103 @@ package beagle
 
 import (
 	"errors"
+	"os"
 
 	"periph.io/x/periph"
+	"periph.io/x/periph/conn/pin"
+	"periph.io/x/periph/conn/pin/pinreg"
+	"periph.io/x/periph/host/sysfs"
 )
+
+var (
+	P8_1  pin.Pin = pin.GROUND
+	P8_2  pin.Pin = pin.GROUND
+	P8_3  pin.Pin = sysfs.Pins[38]
+	P8_4  pin.Pin = sysfs.Pins[39]
+	P8_5  pin.Pin = sysfs.Pins[34]
+	P8_6  pin.Pin = sysfs.Pins[35]
+	P8_7  pin.Pin = sysfs.Pins[66]
+	P8_8  pin.Pin = sysfs.Pins[67]
+	P8_9  pin.Pin = sysfs.Pins[69]
+	P8_10 pin.Pin = sysfs.Pins[68]
+	P8_11 pin.Pin = sysfs.Pins[45]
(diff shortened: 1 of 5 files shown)


experimental/host/beagle/beagle.go
```diff
@@ -6,10 +6,104 @@ package beagle
...
+	P9_31 pin.Pin = sysfs.Pins[110]
+	P9_41 pin.Pin = sysfs.Pins[20]
+	P9_42 pin.Pin = sysfs.Pins[7]
+)
+
+// Present returns true if there is evidence that we have a
+// Beagleboard present.
+func Present() bool {
+	if isArm {
+		_, err := os.Stat("/sys/firmware/devicetree/base/bone_capemgr")
```
maruel: While this is good enough to detect beagle bone family platform, it is not precise enough to guarantee it is a beaglebone black.

Since BeagleXX is really a board family, what about:

- create a subpackage named `black`
- Move all the pin definitions and the driver struct there.
- Keep this Present() function here, which just states "it's a beagle-something"
- Create black.Present() with a more specific check, likely with devicetree.

What do you think?

bodokaiser: These are good suggestions. When I am back in the lab on Monday I can check the sysfs of the Beaglebone for the `Present()` implementation.

How do you suggest to implement SPI, ADC, I2C pins?

experimental/host/beagle/black/black.go
```diff
@@ -0,0 +1,177 @@
+// Copyright 2017 The Periph Authors. All rights reserved.
```
maruel: 2018

experimental/host/beagle/beagle.go
maruel: Lets's start with the gpio in this CL, we can see the rest afterward.

In practice SPI and I²C are provided by the OS (at least on debian) so it's fine.

For ADC, I don't have a strong plan yet, as I'm still nailing down gpio bitstreams and I see ADC streams in some ways as a superset.

experimental/host/beagle/black/black.go
```diff
@@ -0,0 +1,177 @@
...
+	P9_31 pin.Pin = sysfs.Pins[110]
+	P9_41 pin.Pin = sysfs.Pins[20]
+	P9_42 pin.Pin = sysfs.Pins[7]
+)
+
+// Present returns true if there is evidence that we have a
+// Beagleboard present.
+func Present() bool {
+	if isArm {
+		_, err := os.Stat("/sys/firmware/devicetree/base/bone_capemgr")
```
maruel: Add a TODO comment about fixing the check later.

maruel:
gohci

PR #234, opened 2018-04-18
updated beagle/black Present() check

bodokaiser:
Works on `Linux beaglebone 4.4.113-ti-r147` (official debian-stretch image), however cannot confirm if this works on older versions.

It may make sense to define the values I check as constants or even (assuming this works universally) do this check automatically for different platforms, thus a platform would only have to provide a model string to validate against.

### DIFF
--- a/experimental/host/beagle/black/black.go
+++ b/experimental/host/beagle/black/black.go
@@ -5,13 +5,12 @@
 package black
 
 import (
-	"bytes"
 	"errors"
-	"io/ioutil"
 
 	"periph.io/x/periph"
 	"periph.io/x/periph/conn/pin"
 	"periph.io/x/periph/conn/pin/pinreg"
+	"periph.io/x/periph/host/distro"
 	"periph.io/x/periph/host/sysfs"
 )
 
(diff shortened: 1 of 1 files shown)


experimental/host/beagle/black/black.go
```diff
@@ -97,10 +98,15 @@ var (
 // Beagleboard present.
 func Present() bool {
 	if isArm {
-		// TODO: check for beaglebone black in particular
-		_, err := os.Stat("/sys/firmware/devicetree/base/bone_capemgr")
+		buf, err := ioutil.ReadFile("/sys/firmware/devicetree/base/model")
```
maruel: Oh there's already a function for this; I prefer the function to be reused:
https://godoc.org/periph.io/x/periph/host/distro#DTModel
Sorry I hadn't been clear about this.

If /proc/device-tree/model doesn't work well, update https://github.com/google/periph/blob/master/host/distro/devtree.go#L47 and https://github.com/google/periph/blob/master/host/distro/devtree.go#L57 instead.

experimental/host/beagle/black/black.go
```diff
@@ -5,12 +5,14 @@
 package black
 
 import (
+	"bytes"
 	"errors"
-	"os"
+	"io/ioutil"
```
maruel: I recommend to integrate goimports to run on save in your text editor so imports are updated automatically.

maruel:
gohci

Issue #14, opened 2016-10-17
Rename lock* to *Mu where sync.Mutex is used


rakyll:
e.g. pio.go:212

```
var (
    lockDrivers sync.Mutex
    allDrivers  [nbPriorities][]Driver
    byName      = map[string]Driver{}
    lockState   sync.Mutex
    state       *State
)
```

The canonical name for Mutex instances are mu. It might be useful to name lockDrivers to driversMu and lockState to stateMu to be consistent with the standard library style.

PR #235, opened 2018-04-19
bcm283x: hard-code channel #14

### LINKED ISSUE #226: bcm283x: gpio.PWM() Interfering with gpio.Out()
At it simplest a call to
``bcm283x.GPIO25.PWM(dutyCycle, period) //rpi.P1_22``
causes a subsequent call to
``bcm283x.GPIO27.Out(gpio.High) //rpi.P1_13``
to not do-the-right-thing.
**Observed behaviour:** physical pin 13 outputs a random square wave
**Expected behaviour:** physical pin 13 should output a high signal

	package main

	import (
		"bufio"
		"fmt"
		"os"
		"time"

		"periph.io/x/periph/conn/gpio"
		"periph.io/x/periph/host"
		"periph.io/x/periph/host/bcm283x"
	)

	func main() {
		reader := bufio.NewReader(os.Stdin)
		host.Init()

		fmt.Println("one")
		fmt.Println(" - both pin 22 and 13 low or the state they were in at the end of the previous run")
		reader.ReadRune()

		dutyCycle := gpio.Duty(gpio.DutyHalf)
		period := time.Millisecond * 20.0
		bcm283x.GPIO25.PWM(dutyCycle, period) //rpi.P1_22

		fmt.Println("two")
		fmt.Println(" - pin 22 50% dutyCycle 20ms period")
		fmt.Println(" - pin 13 low first run after reboot, random square wave if not first run")
		reader.ReadRune()

		bcm283x.GPIO27.Out(gpio.High) //rpi.P1_13

		fmt.Println("three")
		fmt.Println(" - pin 22 50% dutyCycle 20ms period")
		fmt.Println(" - pin 13 random square wave")
		reader.ReadRune()

		bcm283x.GPIO25.Halt()

		fmt.Println("four")
		fmt.Println(" - both pin 22 and 13 low or high")
	}

I've observed this happening to more than just pin13 though only after setting an output. I've not done an exhaustive search. This random square wave behaviour seems to persist multiple runs and clears on reboot.

maruel:
This is what pi-blaster uses, and people seem to be happy about it.

Stop running the smoke test upon driver initialization, as we're not 100% sure
it's safe.

Fixes #226.

### DIFF
--- a/host/bcm283x/dma.go
+++ b/host/bcm283x/dma.go
@@ -67,6 +67,9 @@ const (
 	periphBus  = 0x7E000000
 	// maxLite is the maximum transfer allowed by a lite channel.
 	maxLite = 65535
+	// dmaChosenChannel is the DMA channel to use. That's the one used by pi-blaster.
+	// It is a lite channel though so it is limited in performance.
+	dmaChosenChannel = 14
 )
 
 // Pages 47-50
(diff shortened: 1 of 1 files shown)


maruel:
Paging @Ivaan, @simokawa , @skyflyer.

I ran it on an RPi3 on Stretch Lite but it hung, sadly.

simokawa:
If I understand correctly, the patch always uses a single fix channel and this does not work for multiple PWM outputs in our implementation.

Our PWM implementation is slightly different from pi-blaster.
In pi-blaster, it handles all input/output using a single DMA channel by oversampling.
In our implementation, we use different channel for different PWM output with different pacing and we need multiple channels.

simokawa:
For random wave case, an interesting point is that PWM pin works as expected but random noise is observed in other pins. It suggests that we allocate DMA channel successfully but enabling PWM clock or DREQ may be interfering other systems.

If we stole in-use DMA channel, we would not see activities in other channels.

PR #237, opened 2018-04-25
enhance logging wrappers

bodokaiser:
Let me know if there is more to do.

### DIFF
--- a/conn/gpio/gpiotest/gpiotest.go
+++ b/conn/gpio/gpiotest/gpiotest.go
@@ -8,6 +8,7 @@ package gpiotest
 import (
 	"errors"
 	"fmt"
+	"log"
 	"sync"
 	"time"
 
@@ -126,6 +127,49 @@ func (p *PinPWM) PWM(duty gpio.Duty, period time.Duration) error {
 	p.D = duty
 	p.P = period
 	return nil
+}
+
+// LogPinIO logs when its state changes.
+type LogPinIO struct {
+	gpio.PinIO
+}
+
+// Real implements gpio.RealPin.
+func (p *LogPinIO) Real() gpio.PinIO {
+	return p.PinIO
+}
+
+// In implements gpio.PinIO.
+func (p *LogPinIO) In(pull gpio.Pull, edge gpio.Edge) error {
+	log.Printf("%s.In(%s, %s)", p, pull, edge)
+	return p.PinIO.In(pull, edge)
+}
+
+// Out mplements gpio.PinIO.
+func (p *LogPinIO) Out(l gpio.Level) error {
+	log.Printf("%s.Out(%s)", p, l)
+	return p.PinIO.Out(l)
+}
+
+// Read implements gpio.PinIO.
+func (p *LogPinIO) Read() gpio.Level {
+	l := p.PinIO.Read()
+	log.Printf("%s.Read() %s", p, l)
+	return l
+}
+
+// Pull implements gpio.PinIO.
+func (p *LogPinIO) Pull() gpio.Pull {
+	log.Printf("%s.Read()", p)
+	return p.PinIO.Pull()
+}
+
+// WaitForEdge implements gpio.PinIO.
+func (p *LogPinIO) WaitForEdge(timeout time.Duration) bool {
+	s := time.Now()
+	r := p.PinIO.WaitForEdge(timeout)
+	log.Printf("%s.WaitForEdge(%s) -> %t after %s", p, timeout, r, time.Since(s))
+	return r
 }
 
 var _ gpio.PinIO = &Pin{}


maruel:
The smoketest is not meant to be a general purpose package. General purpose logging should be in gpiotest. Also it needs to implement RealPin.

bodokaiser:
I had to duplicate the `since(...)` helper function 

https://github.com/google/periph/pull/237/commits/b23169381454fa9d36d2bc9289d41b603b4849e0#diff-22d747d7eae9c9b12c68d86f1cbdc1caR133

https://github.com/google/periph/pull/237/commits/b23169381454fa9d36d2bc9289d41b603b4849e0#diff-d2dff21c2fe1079b4d1c091fb40fca32R476

as `gpiotest.LogPin` and `gpiosmoketest` both depend on it. You probably either want to have `since(..)` moved to some util package or remove its dependency from `LogPin`?

conn/gpio/gpiotest/gpiotest.go
```diff
@@ -128,5 +129,60 @@ func (p *PinPWM) PWM(duty gpio.Duty, period time.Duration) error {
 	return nil
 }
 
+// since returns time in µs since the test start.
```
maruel: In general, the log.Print() statement already include time. The reason the gpiosmoketest prints time is for specialized function which I don't think is generalizeable. As such, I don't think the smoke test should use gpiotest.LogPinIO.

conn/gpio/gpiotest/gpiotest.go
```diff
@@ -128,5 +129,60 @@ func (p *PinPWM) PWM(duty gpio.Duty, period time.Duration) error {
...
+	µs := (time.Since(start) + time.Microsecond/2) / time.Microsecond
+	ms := µs / 1000
+	µs %= 1000
+	return fmt.Sprintf("%3d.%03dms", ms, µs)
+}
+
+// LogPin logs when its state changes.
+type LogPin struct {
+	gpio.PinIO
+	start time.Time
```
maruel: I'm not sure this is a good point here, as in general users just want to log and if they need timestamp, they shall use log.SetFlags(Lmicroseconds).
https://golang.org/pkg/log/#SetFlags

conn/gpio/gpiotest/gpiotest.go
```diff
@@ -128,5 +129,60 @@ func (p *PinPWM) PWM(duty gpio.Duty, period time.Duration) error {
...
+}
+
+// NewLogPin returns a new LogPin with current time.
```
maruel: I don't think this function is necessary if start is removed.

conn/gpio/gpiotest/gpiotest.go
```diff
@@ -128,5 +129,60 @@ func (p *PinPWM) PWM(duty gpio.Duty, period time.Duration) error {
...
+}
+
+// Real implements gpio.RealPin.
+func (p *LogPin) Real() gpio.PinIO {
+	return p.PinIO
+}
+
+// In implements gpio.PinIO.
+func (p *LogPin) In(pull gpio.Pull, edge gpio.Edge) error {
+	log.Printf("%s %s.In(%s, %s)\n", since(p.start), p, pull, edge)
```
maruel: \n is superfluous in the Printf() calls.

conn/gpio/gpiotest/gpiotest.go
```diff
@@ -128,5 +129,60 @@ func (p *PinPWM) PWM(duty gpio.Duty, period time.Duration) error {
...
 
+// since returns time in µs since the test start.
+func since(start time.Time) string {
+	µs := (time.Since(start) + time.Microsecond/2) / time.Microsecond
+	ms := µs / 1000
+	µs %= 1000
+	return fmt.Sprintf("%3d.%03dms", ms, µs)
+}
+
+// LogPin logs when its state changes.
```
maruel: Please name LogPinIO.

conn/gpio/gpiotest/gpiotest.go
```diff
@@ -128,5 +129,60 @@ func (p *PinPWM) PWM(duty gpio.Duty, period time.Duration) error {
...
+
+// Pull implements gpio.PinIO.
+func (p *LogPin) Pull() gpio.Pull {
+	log.Printf("%s %s.Read()\n", since(p.start), p)
+	return p.PinIO.Pull()
+}
+
+// WaitForEdge implements gpio.PinIO.
+func (p *LogPin) WaitForEdge(timeout time.Duration) bool {
+	log.Printf("%s %s.WaitForEdge(%s)\n", since(p.start), p, timeout)
```
maruel: I think it's more valuable to log after, e.g.

```
s := time.Now()
r := p.PinIO.WaitForEdge(timeout)
log.Printf("%s.WaitForEdge(%s) -> %t after %s", p, timeout, r, time.Since(s))
return r
```

That's probably the only case where I do see value to print out how long it waited.

conn/gpio/gpiosmoketest/gpiosmoketest.go
```diff
@@ -478,19 +479,3 @@ func since(start time.Time) string {
...
 }
-
-// loggingPin logs when its state changes.
-type loggingPin struct {
-	gpio.PinIO
-	start time.Time
-}
-
-func (p *loggingPin) In(pull gpio.Pull, edge gpio.Edge) error {
-	fmt.Printf("    %s %s.In(%s, %s)\n", since(p.start), p, pull, edge)
```
maruel: In particular, this is using fmt.Printf, not log, and this is intentional since it is not general logging but really structured output. I just realized I removed them from the CI system by accident, will add this back.

bodokaiser:
Falsely rebased on master instead of v3 branch. However now it should be fine.

conn/gpio/gpiotest/gpiotest.go
```diff
@@ -128,5 +129,56 @@ func (p *PinPWM) PWM(duty gpio.Duty, period time.Duration) error {
 	return nil
 }
 
+// since returns time in µs since the test start.
+func since(start time.Time) string {
```
maruel: Oops please remove since it's now unused

maruel:
gohci

maruel:
Normally non experimental packages should have 100% test coverage, I will add as a follow up.

PR #240, opened 2018-05-13
Update URL link Periph users

kant:


cmd/README.md
```diff
@@ -82,5 +82,5 @@ it](https://periph.io/project/contributing/).
 
 ## Troubleshooting
 
-Having getting the tools to run? See [users/](https://periph.io/users/) for more
+Having getting the tools to run? See [users/](https://groups.google.com/forum/#!forum/periph-users) for more
```
maruel: Hi! Thanks for the PR, that's a good catch.

The the link test should be updated too. In practice, I think it'd be better to encourage people to visit the slack channel, since it's much more active and the feedback loop is quicker.

maruel:
Hi! What do you think about my comment?

maruel:
Superseded by a2bb1b9bbb51748.

Issue #218, opened 2018-01-03
devices/lepton: clean up reading loop and support Lepton 3


maruel:
In parallel to issue #217, @mjs wrote Lepton 3 support as a separate driver at https://github.com/TheCacophonyProject/lepton3 because of issues in the way the driver is currently written.

Issues:

- The Lepton 3 uses "segments", which is different from the way Lepton 2 handles frames.
- The reading loop is suboptimal (including manual CS line management and goroutine) and requires a rewrite.
- Desire to use a []uint16 instead of image.Gray16 which use uint8 underneath.

maruel:
This issue was kinda unfocused. Filed issue #258 specifically for the lepton 3, as the memory allocation was removed and the new image format was added.

PR #241, opened 2018-05-23
lepton: Add new image14bit subpackage

mjs:
This defines a image/draw compatible Gray14 image type and Color14 color type. These will soon be used to store the frames return by the lepton driver.

This is part of the work for #218.

### DIFF
--- a/devices/lepton/image14bit/gray14.go
+++ b/devices/lepton/image14bit/gray14.go
@@ -1,0 +1,82 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package image14bit implements 14-bit per pixel images.
+//
+// It is compatible with the image/draw package.
+package image14bit
+
+import (
+	"image"
+	"image/color"
+	"image/draw"
+)
+
+// Gray14 represents an image of 14-bit values.
+type Gray14 struct {
+	// Pix holds the image's pixels. Each uint16 element represents one 14-bit
+	// pixel.
+	Pix []uint16
+	// Stride is the Pix stride (in pixels) between vertically adjacent pixels.
+	Stride int
+	// Rect is the image's bounds.
+	Rect image.Rectangle
+}
+
+// NewGray14 returns an initialized Gray14 instance.
+func NewGray14(r image.Rectangle) *Gray14 {
+	w, h := r.Dx(), r.Dy()
+	pix := make([]uint16, w*h)
+	return &Gray14{Pix: pix, Stride: w, Rect: r}
+}
+
+// ColorModel implements image.Image.
+func (i *Gray14) ColorModel() color.Model {
+	return Intensity14Model
+}
+
+// Bounds implements image.Image.
+func (i *Gray14) Bounds() image.Rectangle {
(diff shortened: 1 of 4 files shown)


mjs:
I'm a little unsure about the naming of `Color14`. The image and related color types in the standard library use the same name but in different packages. Defining a new package just for the color seemed like overkill. I'm open to suggestions.

devices/lepton/image14bit/color.go
```diff
@@ -0,0 +1,44 @@
...
+// that can be found in the LICENSE file.
+
+package image14bit
+
+import (
+	"image/color"
+	"strconv"
+)
+
+// Color14 implements a 14-bit color.
```
maruel: I'd prefer Intensity14, Color is misleading IMHO. Comments could be:

// Intensity14 is a 14 bits grayscale implementation of color.Color
//
// Valid range is between 0 and 16383.

devices/lepton/image14bit/color.go
```diff
@@ -0,0 +1,44 @@
...
+type Color14 uint16
+
+// RGBA returns a grayscale result.
+func (g Color14) RGBA() (uint32, uint32, uint32, uint32) {
+	// This isn't perfect but is the best we can do without using
```
maruel: I'd remove the command and do instead:

b := uint32(i) & 1
i := uint32(i) << 2 || b << 1 || b

That's better than floating point math. :)

devices/lepton/image14bit/color_test.go
```diff
@@ -0,0 +1,24 @@
...
+	if r, g, b, a := Color14(0).RGBA(); r != 0 || g != r || b != r || a != 65535 {
+		t.Fatal(r, g, b, a)
+	}
+	if Color14(16383).String() != "Color14(16383)" || Color14(0).String() != "Color14(0)" {
+		t.Fail()
+	}
+	if Color14(8192) != convertColor14(Color14(8192)) {
+		t.Fail()
+	}
+}
```
maruel: Please a test for convertColor14 code path lines 38-43, i.e. with a different color model.

devices/lepton/image14bit/image.go
```diff
@@ -0,0 +1,82 @@
...
+func (i *Gray14) Opaque() bool {
+	return true
+}
+
+// At implements image.Image.
+func (i *Gray14) At(x, y int) color.Color {
+	return i.Color14At(x, y)
+}
+
+// Color14At returns the Color14 value at a point.
```
maruel: Rename this and SetColor14 to use Intensity instead.

devices/lepton/image14bit/color.go
mjs: Done. Agreed that Intensity14 is a better name.

devices/lepton/image14bit/color.go
mjs: Nice. I had to make some minor corrections but I understood what you meant.

devices/lepton/image14bit/image.go
mjs: Done

maruel:
gohci

maruel:
You may want to add yourself into AUTHORS and CONTRIBUTORS in a follow up.

PR #243, opened 2018-06-18
display: move devices.Display as display.Drawer

maruel:
This removes the last interface from devices, which was misplaced due to
historical accident. The new package display will also be the location for an
interface for text output only devices.


conn/display/display.go
```diff
@@ -0,0 +1,43 @@
...
+	// color format.
+	//
+	// If the image is smaller than the device frame size, the result is
+	// undefined. In most drivers the previous pixels are left intact but this
+	// behavior is not guaranteed. If the image is larger than the frame, the
+	// extraneous pixels are ignored and no error returned.
+	//
+	// sp aligns the image at this offset, use image.Point{} to take the image
+	// as-is.
+	Draw(img image.Image, sp image.Point) error
```
mattetti: I'd suggest to document a bit more what the image.Point reference is for, I would have 2 questions reading this document:
- Is that the origin, if that's the case, is it bottom left or top left?
- why is it abbreviated as `sp`

maruel: I took it from https://golang.org/pkg/image/draw/#Drawer

I agree it's a bit obscure. I initially tried without the sp argument but I can see use cases for it, even if it would be fairly rare. The other option is to skip the sp argument and force users to create an intermediary image themselves (it's actually not a bad idea).

maruel: I added conn/display/example_test.go which discusses this more.

maruel:
I redid it again, I decided to reduce the amount of change, keeping the destination rectangle. I increased documentation.

PR #247, opened 2018-06-26
weather: move Env and SenseEnv out of physic

maruel:
physic is now just a bucket for units.


maruel:
The next step is to rename physic to metric or measurement.

maruel:
Another option is to inverse the names. Make the package named environment, and change the struct Env to Weather. Then it'd make it more sensible to have a struct Position for geospatial position sensing.

maruel:
Thinking about it more; I'll drop this change, as all in all, it's not really worth the churn.

PR #251, opened 2018-07-09
apa102: Refactor tests

benlazarus:
- Convert existing tests to table-driven
- Make reading []byte diffs easier by aligning lines
- Should be a no-op in terms of what's being tested and with what data

### DIFF
--- a/devices/apa102/apa102_test.go
+++ b/devices/apa102/apa102_test.go
@@ -334,7 +334,7 @@ func TestDevEmpty(t *testing.T) {
 		t.Fatalf("%d %v", n, err)
 	}
 	if expected := []byte{0x0, 0x0, 0x0, 0x0, 0xFF}; !bytes.Equal(expected, buf.Bytes()) {
-		t.Fatalf("%#v != %#v", expected, buf.Bytes())
+		t.Fatalf("\ngot: %#v\nwant: %#v\n", buf.Bytes(), expected)
 	}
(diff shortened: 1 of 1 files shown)


maruel:
Thanks for the change! 

I try to keep the library working on older Go version (so users can just `sudo apt install golang` and get going), I try to not make tests fail to compile there.

The way to use t.Run() is to move the tests into a conditionally compiled file on Go 1.8+. You can find an example at https://github.com/google/periph/tree/master/devices/cap1xxx

- Move all the test cases using t.Run() into a new file apa102_go_1_8_test.go
- Include a `// +build go1.8` statement to this file

Sounds good?

benlazarus:
Thanks for the quick reply!  What version do want to target in general?  I'm doing this in preparation for a functional change, so I'll make sure I target the version you want there, too.

FWIW, it looks like t.Run() was added in 1.7.  Would you prefer if I just made them not be subtests?  Like: https://github.com/benlazarus/periph/commit/f93936d833533a9bfb57d8512e9f9ca3ecf2a901 ?

maruel:
gohci

maruel:
I was aiming for go1.5.3 but I tested locally on 1.4.2 because I happened to have it handy and realized that it was broken here, so I fixed it. As long as the cost is low enough, I prefer to keep compatibility with fairly old versions because that's coherent with the project philosophy to not get in the way. That said that may eventually become impossible if we want to start using context.Context.

Re: subtest vs simple table, I don't mind, I'm fine with only partial test coverage on older toolchains, as long as "go test" still passes.

PR #252, opened 2018-07-11
apa102: refactor benchmarks

benlazarus:
Should be a no-op.  Benchmark runs look roughly the same as head on my machine.

### DIFF
--- a/devices/apa102/apa102_test.go
+++ b/devices/apa102/apa102_test.go
@@ -636,14 +636,15 @@ func TestInit(t *testing.T) {
 
 //
 
-type fn func(int) byte
+type genColor func(int) [3]byte
 
-func benchmarkWrite(b *testing.B, o Opts, length int, f fn) {
+func benchmarkWrite(b *testing.B, o Opts, length int, f genColor) {
 	var pixels []byte
 	for i := 0; i < length; i++ {
-		pixels = append(pixels, f(i))
+		c := f(i)
+		pixels = append(pixels, c[:]...)
 	}
-	o.NumPixels = length / 3
+	o.NumPixels = length
 	b.ReportAllocs()
 	d, _ := New(spitest.NewRecordRaw(ioutil.Discard), &o)
 	_, _ = d.Write(pixels[:])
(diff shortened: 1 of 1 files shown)


devices/apa102/apa102_test.go
```diff
@@ -636,15 +636,15 @@ func TestInit(t *testing.T) {
 
 //
 
-func BenchmarkWriteWhite(b *testing.B) {
-	b.ReportAllocs()
-	pixels := [150 * 3]byte{}
-	for i := range pixels {
-		pixels[i] = 0xFF
+type fn func(int) byte
```
maruel: I would have expected a triplet, e.g.

    type genColor func(int) [3]byte

then

    pixel = append(pixels, f(i)...)

but I don't mind that much.

devices/apa102/apa102_test.go
```diff
@@ -636,15 +636,15 @@ func TestInit(t *testing.T) {
...
+
+func benchmarkWrite(b *testing.B, o Opts, length int, f fn) {
```
maruel: I'd prefer if length was in number of pixels, not in the length of the buffer. This is confusing IMHO.

devices/apa102/apa102_test.go
```diff
@@ -653,40 +653,35 @@ func BenchmarkWriteWhite(b *testing.B) {
...
+	o.Intensity = 250
+	o.Temperature = 5000
+	benchmarkWrite(b, o, 150*3, func(i int) byte { return uint8(i) + uint8(i>>8) })
+}
+
+func BenchmarkWriteColorfulVariation(b *testing.B) {
+	// Continuously vary the lookup tables.
 	b.ReportAllocs()
-	pixels := [150 * 3]byte{}
+	pixels := [256 * 3]byte{}
```
maruel: Good point.

devices/apa102/apa102_test.go
```diff
@@ -698,21 +693,25 @@ func BenchmarkWriteColorful(b *testing.B) {
 	_, _ = d.Write(pixels[:])
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
+		d.Intensity = uint8(i)
```
maruel: I'm confused as to why it was this way, I probably messed up a prior refactoring, I don't recall. Thanks for catching!

devices/apa102/apa102_test.go
```diff
@@ -698,21 +693,25 @@ func BenchmarkWriteColorful(b *testing.B) {
...
 }
 
-func BenchmarkDrawNRGBAColorful(b *testing.B) {
-	// Takes the fast path.
-	b.ReportAllocs()
-	img := image.NewNRGBA(image.Rect(0, 0, 150, 1))
-	for i := range img.Pix {
-		img.Pix[i] = uint8(i) + uint8(i>>8)
+func benchmarkDraw(b *testing.B, o Opts, img image.Image, f fn) {
+	switch im := img.(type) {
```
maruel: A generic way to do:
```
for x := 0; x < bounds.Dx(); x++ {
  pix := fn(i)
  c : = color.NRGBA{R: pix[0], G: pix[1], B:pix[2], A:255}
  img.Set(0, x, c)
}
```

since setup time is not a big deal here.

devices/apa102/apa102_test.go
benlazarus: Done.

devices/apa102/apa102_test.go
benlazarus: Done.

devices/apa102/apa102_test.go
benlazarus: I didn't actually change this -- the way it's presenting the diff is pretty confusing.  My goal was to keep the same benchmarks with the same data for each.

devices/apa102/apa102_test.go
benlazarus: I didn't change this either .. the diff presentation is not awesome :)

devices/apa102/apa102_test.go
benlazarus: Done.  Though I'm also allowing for a non-zero y even though the current tests don't use any.

devices/apa102/apa102_test.go
```diff
@@ -698,21 +702,37 @@ func BenchmarkWriteColorful(b *testing.B) {
...
 }
 
-func BenchmarkDrawNRGBAColorful(b *testing.B) {
-	// Takes the fast path.
-	b.ReportAllocs()
-	img := image.NewNRGBA(image.Rect(0, 0, 150, 1))
-	for i := range img.Pix {
-		img.Pix[i] = uint8(i) + uint8(i>>8)
+func fillImage(img image.Image, f genColor) {
+	switch im := img.(type) {
```
maruel: Oh I meant you don't need the type switch case at all. It can be rewritten as

```
func fillImage(img image.Image, f genColor) {
  for x := 0; x < im.Bounds().Dx(); x++ {
    pix := f(x)
    im.Set(x, 0, color.NRGBA{R: pix[0], G: pix[1], B: pix[2], A: 255})
  }
}
```

maruel:
gohci

maruel:
Committing anyway you can do a follow up to simplify fillImage if you want.

devices/apa102/apa102_test.go
benlazarus: Hm, I couldn't find a reasonable way to not have the type assertion because image.Image doesn't have a Set() method.  Thank for the review!

maruel: Ah it's because you want draw.Image;
https://golang.org/pkg/image/draw/#Image

PR #254, opened 2018-07-15
apa102: refactor raster functions

benlazarus:


benlazarus:
Benchmark runs: https://gist.github.com/benlazarus/82f8c77e0bdeb2ce45a20ae1121273a5

maruel:
I also reproduce locally the ~25% performance reduction for BenchmarkDrawNRGBAColorful. It's probably why I had created the separate loops in the first place.

I'd prefer that performance for this path to not regress. FYI, I had requests to create a fast path for image.RGBA too, since some drawing libraries use this instead of NRGBA.

benlazarus:
Okay.  After tinkering a little, I think you're right, it's probably due to moving the branches into the loop.  I have an idea for a solution (which can also fast path RGBA).  It might be a day or two before I have time to play with it.

maruel:
It's weird, as some benchmarks like BenchmarkWriteDim significantly increased in performance. But since the benchmarks are being changed, it's hard to make apple to apple comparison. Would you mind splitting this PR in two? Let's commit the benchmark refactor first, then we can properly assess the performance impact.

maruel:
The reason I ask is that I did observe the performance degradation (not too bad just 10%) on BenchmarkDrawNRGBAColorful. I'm fine to live with it, but having two separate commits would be helpful when looking back.

benlazarus:
Sure thing, I'll do it later tonight.  Though I didn't change any of the existing benchmarks here -- I just added a couple new ones.

benlazarus:
Done.  I squashed everything down here and put the benchmark changes into https://github.com/google/periph/pull/259

maruel:
gochi

maruel:
gohci

maruel:
(Typing is hard)

Thanks!

PR #260, opened 2018-07-19
apa102: Add RawColors option

benlazarus:
This adds a new option to bypass the perceptual remapping code and
directly write the RGB values to the APA102, similar to what the
AdaFruit DotStar Python library does.


maruel:
gochi

maruel:
gohci

PR #264, opened 2018-07-24
fs: add support for GOARCH=mips/mipsle on GOOS=linux

maruel:
It uses a slightly different linux opcode encoding. Just shuffle the bits as it
goes.

Tested with:
  GOARCH=mipsle go build ./...
  GOARCH=mips go build ./...
  go test ./...

### DIFF
--- a/host/fs/fs.go
+++ b/host/fs/fs.go
@@ -16,6 +16,10 @@ import (
 
 // Ioctler is a file handle that supports ioctl calls.
 type Ioctler interface {
+	// Ioctl sends a linux ioctl on the file handle. op is effectively a uint32.
+	//
+	// The op is expected to be encoded in the format on x64. ARM happens to
+	// share the same format.
 	Ioctl(op uint, data uintptr) error
 }
 
@@ -58,6 +62,12 @@ type File struct {
 
 // Ioctl sends an ioctl to the file handle.
 func (f *File) Ioctl(op uint, data uintptr) error {
+	if isMIPS {
+		var err error
+		if op, err = translateOpMIPS(op); err != nil {
+			return err
+		}
+	}
 	return ioctl(f.Fd(), op, data)
 }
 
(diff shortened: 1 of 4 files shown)


maruel:
@svenschwermer please test as I can't do locally. If you have hardware to recommend and where to source it, that'd be appreciated.

host/fs/ioctl_mips_like.go
```diff
@@ -0,0 +1,28 @@
...
+
+package fs
+
+func translateOp(op uint) uint {
+	// Decode the arm/x64 encoding and reencode as MIPS specific linux ioctl.
+	// arm/x64: DIR(2), SIZE(14), TYPE(8), NR(8)
+	// mips:    DIR(3), SIZE(13), TYPE(8), NR(8)
+	const mask = 1<<(14+8+8) - 1
+	dir := op &^ mask
+	if (dir & 1 << (13 + 8 + 8)) != 0 {
```
svenschwermer: Haven't you just cleared all but the dir bits in the line above? I guess, this `if` would always be false.

maruel: You are right, the code was particularly broken. Sorry about that. I moved the code out so I could test it unconditionally.

maruel: Please take a look at the new commit.

svenschwermer:
Thanks, @maruel. I'll do some testing soon.

host/fs/fs.go
```diff
@@ -58,6 +62,12 @@ type File struct {
 
 // Ioctl sends an ioctl to the file handle.
 func (f *File) Ioctl(op uint, data uintptr) error {
+	if isMIPS {
```
svenschwermer: I'd remove `ioctl_mips_like.go` and `ioctl_other.go` and replace this line with `if runtime.GOARCH == "mips" || runtime.GOARCH == "mipsle" {`. It costs the same as it will be evaluated at compile time.

maruel: That's true but I prefer this pattern because there can't be typo (like the one I had done with "mipsel") and I tend to break stuff all the time otherwise; i.e. I don't trust myself.

host/fs/fs.go
```diff
@@ -94,3 +104,27 @@ var (
...
+	const mask = (1 << (13 + 8 + 8)) - 1
+	out := op & mask
+	// Convert dir.
+	switch op >> (14 + 8 + 8) {
+	case 0: // none
+		out |= 1 << (13 + 8 + 8)
+	case 1: // write
+		out |= 2 << (13 + 8 + 8)
+	case 2: // read
+		out |= 3 << (13 + 8 + 8)
```
svenschwermer: I believe, this is wrong. It should be
```go
case 1: // write
	out |= 4 << (13 + 8 + 8)
case 2: // read
	out |= 2 << (13 + 8 + 8)
```

maruel: Oh you're right, sorry about that. Pushed a fixed version.

svenschwermer:
@maruel Works as expected now. Even though I must specify a non-nil read buffer for SPI transfers to work. However, I suspect that this is a quirk of the SPI driver on my SoC. x/exp/io/spi [did this implicitly](https://github.com/golang/exp/commit/10900705ca5eaf0adcea5ec2645015bbf1f7a479).

maruel:
I explicitly do not want to allocate memory unless needed. Write-only definitely works on a RPi. We can follow up on an issue if you want.

PR #267, opened 2018-08-06
Implements clock mode handling for bitbang/spi

### LINKED ISSUE #266: bitbang/spi: clock does not fire
[Lines 168 and 175](https://github.com/google/periph/blob/master/experimental/devices/bitbang/spi.go#L168) both set clock to Low, so it never goes high.  The exact values of these depend on Mode (which is currently hard-filtered to Mode 3, which is supposed to use reversed polarity), but in all cases, the clock must toggle to be useful.

Ulexus:
Handles clock-related Mode options for bitbanging SPI.

Fixes #266

### DIFF
--- a/experimental/devices/bitbang/spi.go
+++ b/experimental/devices/bitbang/spi.go
@@ -70,9 +70,6 @@ func (s *SPI) Connect(f physic.Frequency, mode spi.Mode, bits int) (spi.Conn, er
 	if f < 0 {
 		return nil, errors.New("bitbang-spi: invalid frequency")
 	}
-	if mode != spi.Mode3 {
-		return nil, fmt.Errorf("bitbang-spi: mode %v is not implemented", mode)
-	}
 	s.spiConn.mu.Lock()
 	defer s.spiConn.mu.Unlock()
 	s.spiConn.freqDev = f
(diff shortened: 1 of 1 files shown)


experimental/devices/bitbang/spi.go
```diff
@@ -147,36 +150,97 @@ func (s *spiConn) Duplex() conn.Duplex {
 	return conn.Full
 }
 
+func (s *spiConn) clockOn() error {
+	if s.mode&spi.Mode2 == spi.Mode2 {
```
maruel: One option is to add properties to spiConn for clock, i.e. at line 135, replace mode by:

    clockIdle           gpio.Level
    readAfterClockPulse bool

and initialize them in Connect(). Then you can change the clockOn/clockOff functions to be:

    s.sck.Out(s.clockIdle)
    s.sck.Out(!s.clockIdle)

maruel:
gohci

experimental/devices/bitbang/spi.go
```diff
@@ -70,8 +70,11 @@ func (s *SPI) Connect(f physic.Frequency, mode spi.Mode, bits int) (spi.Conn, er
...
 		return nil, errors.New("bitbang-spi: invalid frequency")
 	}
-	if mode != spi.Mode3 {
-		return nil, fmt.Errorf("bitbang-spi: mode %v is not implemented", mode)
+	if mode&spi.HalfDuplex == spi.HalfDuplex {
+		return nil, errors.New("bitbang-spi: half-duplex mode not supported")
+	}
+	if mode&spi.LSBFirst == spi.LSBFirst {
+		return nil, errors.New("bitbang-spi: LSBFirst mode not supported")
 	}
```
maruel: In practice, you'd want to check explicitly for 0 <= mode <= 3, so that if new flags are added, they are not silently ignored.

experimental/devices/bitbang/spi.go
Ulexus: That sounds nicer, yes.

experimental/devices/bitbang/spi.go
Ulexus: Good point.

PR #268, opened 2018-08-06
bitbang/spi cleanup

Ulexus:
Changes based on feedback from @maruel 

 - precalculate clock idle polarity and remove clockOn()/clockOff()
 - check for higher (unknown) modes and reject them
 - move internal functions down, to be grouped together

### DIFF
--- a/experimental/devices/bitbang/spi.go
+++ b/experimental/devices/bitbang/spi.go
@@ -91,7 +91,9 @@ func (s *SPI) Connect(f physic.Frequency, mode spi.Mode, bits int) (spi.Conn, er
 
 	// Set clock idle polarity, ensuring an idle clock to start
 	s.spiConn.clockIdle = gpio.Level(mode&spi.Mode2 == spi.Mode2)
-	s.spiConn.sck.Out(s.spiConn.clockIdle)
+	if err := s.spiConn.sck.Out(s.spiConn.clockIdle); err != nil {
+		return nil, fmt.Errorf("bitbang-spi: failed to idle clock: %v", err)
+	}
 
 	return &s.spiConn, nil
 }
--- a/experimental/devices/bitbang/spi.go
+++ b/experimental/devices/bitbang/spi.go
@@ -76,6 +76,9 @@ func (s *SPI) Connect(f physic.Frequency, mode spi.Mode, bits int) (spi.Conn, er
 	if mode&spi.LSBFirst == spi.LSBFirst {
 		return nil, errors.New("bitbang-spi: LSBFirst mode not supported")
 	}
+	if mode >= 0x20 {
+		return nil, fmt.Errorf("bitbang-spi: unhandled mode %d(%s)", mode, mode.String())
+	}
 	s.spiConn.mu.Lock()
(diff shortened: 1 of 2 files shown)


experimental/devices/bitbang/spi.go
```diff
@@ -84,6 +87,15 @@ func (s *SPI) Connect(f physic.Frequency, mode spi.Mode, bits int) (spi.Conn, er
 	}
 	s.spiConn.mode = mode
 	s.spiConn.bits = bits
+	s.spiConn.readAfterClockPulse = mode&spi.Mode1 == spi.Mode1
+
+	// Set clock idle polarity
+	if mode&spi.Mode2 == spi.Mode2 {
```
maruel: It's a bit of a hack but you can do:

    s.spiConn.clockIdle = gpio.Level(mode&spi.Mode2 == spi.Mode2)

Also, we would want to have the pin setup to idle state right here.

maruel:
gohci

experimental/devices/bitbang/spi.go
Ulexus: I wouldn't normally compress the logic so much, but since the rest of the connect options are handled similarly, that sounds reasonable.

experimental/devices/bitbang/spi.go
```diff
@@ -84,6 +87,12 @@ func (s *SPI) Connect(f physic.Frequency, mode spi.Mode, bits int) (spi.Conn, er
 	}
 	s.spiConn.mode = mode
 	s.spiConn.bits = bits
+	s.spiConn.readAfterClockPulse = mode&spi.Mode1 == spi.Mode1
+
+	// Set clock idle polarity, ensuring an idle clock to start
+	s.spiConn.clockIdle = gpio.Level(mode&spi.Mode2 == spi.Mode2)
+	s.spiConn.sck.Out(s.spiConn.clockIdle)
```
maruel: Please check the error. Also set CS too when NoCS is not set, and MISO and MOSI as In()/Out(Low?)

maruel: Ah ignore MISO/MOSI/CS, that's done above. Just check the error and it'll be fine.

Ulexus: Yep, thanks

maruel:
gohci

experimental/devices/bitbang/spi.go
Ulexus: Actually, I'm not so sure all the pin initialization shouldn't be moved to Connect() anyway, in the realm of expected results.  My assumption would be that if there is a separate Connect(), the initial SPI device creation would not modify the hardware.  The rest of the codebase, however, does not appear to make that distinction.  This is definitely in the nit domain.

maruel: That's a good point.

This was done this way initially so the failure (aka an invalid pin) would occur earlier at SPI instantiation, because Connect() is called within a device driver, which would be confusing for the users.

Ulexus: Yeah, that is totally understandable.  Thanks for the reviewing!

PR #274, opened 2018-08-11
pin: add PinFunc

maruel:
- This will eventually be implemented by all pin.Pin implementations and merged
  back into pin.Pin for v4.
- The pin.Func const in other conn/ packages and host/ implementation is in a
  follow up.

### DIFF
--- a/conn/pin/example_test.go
+++ b/conn/pin/example_test.go
@@ -6,7 +6,9 @@ package pin_test
 
 import (
 	"fmt"
+	"log"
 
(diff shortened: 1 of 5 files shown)


maruel:
The rest is in https://github.com/google/periph/compare/pin_func_1...pin_func_2?expand=1.

I'm concerned about: pin.FuncInvalid and friends, it looks ugly but since there's already pin.INVALID for the actual pin, I'm not sure what else to do. The other option is to not create one?

maruel:
Removed pin.FuncInvalid, which solves the problems I had.

PR #276, opened 2018-08-13
bitbang/spi: disable CS before manipulating CLK

### LINKED ISSUE #275: bitbang/spi: CS pin should be deactivated before clock pin is modified
In `NewSPI`, we do not yet know the clock polarity, so we cannot know in which direction the clock line should be set in order to move it to the idle state.  If we guess incorrectly and the CS line happens to be active, we could inadvertently send data.

The easiest fix is to move the CS line reset to occur _before_ the CLK line reset.  Ideally, we shouldn't manipulate CLK before we know the idle polarity (in Connect), but there is value in failing early, should the CLK line be invalid.

Ulexus:
Fixed #275

### DIFF
--- a/experimental/devices/bitbang/spi.go
+++ b/experimental/devices/bitbang/spi.go
@@ -30,6 +30,11 @@ import (
 //
 // cs can be nil.
 func NewSPI(clk, mosi gpio.PinOut, miso gpio.PinIn, cs gpio.PinOut) (*SPI, error) {
+	if cs != nil {
+		if err := cs.Out(gpio.High); err != nil {
+			return nil, err
+		}
+	}
 	if err := clk.Out(gpio.High); err != nil {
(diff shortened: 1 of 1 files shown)


Ulexus:
@maruel Would just _reading_ CLK and MOSI be sufficient for your "early fail" philosophy, then properly resetting them in Connect() after we know polarity?  I think that would be a cleaner approach.  The defaults here assume Mode3, which means all of my manipulations result in a wasted clock and data pulse (using Mode0) with the current code.

maruel:
The more I think about it the more this does of setting the pins before knowing the edges was a bad idea. So I wouldn't mind moving all the setup code to Connect(). Sorry for the back and forth on this topic.

Thanks for the fix!

maruel:
gohci

PR #278, opened 2018-08-13
bitbang/spi: move pin init to Connect() from NewSPI()

### LINKED ISSUE #277: bitbang/spi: pin initialization should occur in `Connect()` rather than `NewSPI()`
We should move the pin initialization to `Connect()` so that we know the options to determine the initial states, which are indeterminate at `NewSPI()`.

Ulexus:
Fixes #277


maruel:
gohci

maruel:
Thanks!

PR #284, opened 2018-09-02
sysfs-gpio: always reset edge detection to 'none' after starting epoll

davidsansome:
Without this edges are never delivered for me.  *shrug*

### DIFF
--- a/host/sysfs/gpio.go
+++ b/host/sysfs/gpio.go
@@ -167,19 +167,22 @@ func (p *Pin) In(pull gpio.Pull, edge gpio.Edge) error {
 				return p.wrap(err)
 			}
 		}
-		if p.edge != edge {
-			var b []byte
-			switch edge {
-			case gpio.RisingEdge:
-				b = bRising
-			case gpio.FallingEdge:
-				b = bFalling
-			case gpio.BothEdges:
-				b = bBoth
-			}
-			if err := seekWrite(p.fEdge, b); err != nil {
-				return p.wrap(err)
-			}
+		// Always reset the edge detection mode to none after starting the epoll,
+		// otherwise edges are not always delivered.
+		if err := seekWrite(p.fEdge, bNone); err != nil {
+			return p.wrap(err)
+		}
+		var b []byte
+		switch edge {
+		case gpio.RisingEdge:
+			b = bRising
+		case gpio.FallingEdge:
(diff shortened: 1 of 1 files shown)


host/sysfs/gpio.go
```diff
@@ -167,19 +167,22 @@ func (p *Pin) In(pull gpio.Pull, edge gpio.Edge) error {
...
-			case gpio.FallingEdge:
-				b = bFalling
-			case gpio.BothEdges:
-				b = bBoth
-			}
-			if err := seekWrite(p.fEdge, b); err != nil {
-				return p.wrap(err)
-			}
+		// Always reset the edge detection mode to none after starting the epoll,
+		// otherwise edges are not always delivered.
```
maruel: Can you expand the comment like "as observed on a Allwinner A20 running kernel x.y.z". Otherwise we'll have no clue why this was added.

maruel:
gohci

maruel:
sysfs-benchmark doesn't show significant difference, mainly because it's not testing edge change:
before:
https://gist.github.com/gohci-bot/17edc443d079c2a80b17a0fc654809a8#file-cmd12-in-7-557s
after:
https://gist.github.com/gohci-bot/1739bfe0f8ab672133cdfa708ee86e9c#file-cmd12-in-7-556s

I don't think it's a performance sensitive path so it's fine, I'll just wait for a better comment. Thanks for the contribution!

host/sysfs/gpio.go
davidsansome: Good thinking, done!

host/sysfs/gpio.go
```diff
@@ -167,19 +167,23 @@ func (p *Pin) In(pull gpio.Pull, edge gpio.Edge) error {
...
+			b = bRising
+		case gpio.FallingEdge:
+			b = bFalling
+		case gpio.BothEdges:
+			b = bBoth
+		}
+		if err := seekWrite(p.fEdge, b); err != nil {
+			return p.wrap(err)
 		}
 	}
```
maruel: One thing I realized is that when going from edge to no edge it's not disabling it. It's not a big deal but that's generating unneeded interrupts. That's out of scope for this change though.

maruel:
gohci

maruel:
Thanks a lot!

PR #289, opened 2018-09-26
hosts: Add support for mt7688

matthewhartstonge:
Aims to add support for the MediaTek MT7688 chipset.

Development Boards currently using this chipset:
* LinkIt Smart 7688 (and duo)
* Onion Omega 2

PR #290, opened 2018-09-28
experimental/host: mt7688 code review fixes

matthewhartstonge:
Fixes code review requests as per #289

I noticed I never `init()`'d the package Pins, so this has also added that in.


maruel:
gohci

experimental/host/mt7688/pin.go
```diff
@@ -27,67 +26,58 @@ const (
...
+	{number: 43, name: "GPIO43", defaultPull: gpio.Float},
+	{number: 44, name: "GPIO44", defaultPull: gpio.Float},
+	{number: 45, name: "GPIO45", defaultPull: gpio.Float},
+	{number: 46, name: "GPIO46", defaultPull: gpio.Float},
 }
 
+// Pin is a GPIO number (GPIOnn) on MT7688(AN/KN).
 type Pin struct {
 	// Immutable.
 	number      int
```
maruel: [optional style comment]

An option is to inline alt as:
```
  alt [3]pin.Func
```

so that you would inline the functions inside the cpuPins definition directly. In practice it doesn't change much things, so it's just FYI as an idea.

experimental/host/mt7688/gpio.go
```diff
@@ -67,54 +67,104 @@ var (
...
-	"EPHY_LED4_N_JTRST_N (7688AN)": {"EPHY_LED4_N", "", "JTAG_RST_N"},
-	"EPHY_LED3_N_JTCLK (7688AN)":   {"EPHY_LED3_N", "", "JTAG_CLK"}, // 40
-	"EPHY_LED2_N_JTMS (7688AN)":    {"EPHY_LED2_N", "", "JTAG_TMS"},
-	"EPHY_LED1_N_JTDI (7688AN)":    {"EPHY_LED1_N", "", "JTAG_TDI"},
-	"EPHY_LED0_N_JTDO (7688AN)":    {"EPHY_LED0_N", "", "JTAG_TDO"},
-	"WLED_N (7688AN)":              {"WLED_N"},
-	"UART_TXD1":                    {"UART_TXD1", "PWM_CH0"}, // 45
-	"UART_RXD1":                    {"UART_RXD1", "PWM_CH1"},
+var mapping = [][3]pin.Func{
+	{"I2SSDI", "PCMDRX"}, // 0
```
maruel: In the next commit, you can update the function names.

maruel:
Thanks!

experimental/host/mt7688/pin.go
matthewhartstonge: I had a look into this and I decided to go with the common conventions - more for new members perusing across multiple host packages.
A mapping of some sort seems to appear in most host packages.

Again, if you would like to converge convention on storing the alt functions in the Pin struct for encapsulation, happy to work that into another PR :smile:

experimental/host/mt7688/gpio.go
matthewhartstonge: Resolved via #313

PR #291, opened 2018-09-28
Add SPI driver for Neopixel LEDs

whmeitzler:
This commit adds support and tests for WS2812B Neopixel LEDs.
It's been tested on a Pi0 driving 50 WS2812B LEDs.
Comments and criticism are welcome.


whmeitzler:
Hey @googlebot, I've signed your CLA.

maruel:
Ah, you may have to amend the commit with the email address you used in the CLA, or redo the CLA with the email you use as author for commit, as you prefer. I'll try to review ASAP but I'm having dinner now, so it may take a few hours.

maruel:
Oh you may want to take a look at experimental/devices/nrzled. The reason it is not named ws2812b is that there are a ton of clones using the same protocol.

whmeitzler:
@googlebot let's try that again. 
@maruel Thanks for pointing out the NRZ driver, I had played with it and was unable to drive LEDs from my Pi0 - even the 400kHz freq was high enough for the driver to fail out. This implementation uses SPI instead of memory-mapped bitbanging, so it should be a bit less demanding on CPU. It also avoids memory-mapped access, so root is not needed.

maruel:
Yes, my thinking is that this code could be added inside nrzled. example_test.go is incorrect.

experimental/devices/ws2812b/ws2812b.go
```diff
@@ -0,0 +1,506 @@
...
+		// Offset into the output buffer using rect
+		d.raster(dst[4*rect.Min.X:], m.Pix[start:end], true)
+	}
+}
+
+var _ display.Drawer = &Dev{}
+
+// The bit lookup table converts a single byte into its 3 byte SPI symbol
+// 0 => 1000 and 1 => 1110
+var bitlut = map[byte][]byte{
```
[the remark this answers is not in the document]
whmeitzler: > A map[byte] is the equivalent of [256]. An array is more efficient for both memory usage and access speed. 
The map (and the lookup table as a whole) is a byproduct of hacking this to a working state- resource concerns took a back seat to debug legibility. Happy to move this to a [256]byte.

> [NRZ func] is fast enough so [the lookup table (?)] is not needed
The mapping is static, so calling a translation function on each write seems dubious. Even if it's fast enough, it's work we don't really need to do. Of course, a giant hardcoded lookup table is odd too. What if we used NRZ or similar to generate the lookup table during the proposed call to nrzled.NewSPI()? 

Thanks for the review! Expect a commit to this branch incorporating these discussions in the next few days.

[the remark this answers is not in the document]
maruel: Fine with hardcoding the table, as long as there's a unit test that confirms that the values are the same with the algorithm. This can be done as a separate CL.

Sorry for requesting big changes, and thanks for sending this PR. I think this is a cool idea.

Eventually the driver should be smart about the SPI buffer size (i.e. https://periph.io/platform/raspberrypi/#spi-buffer-size) and at least when possible adjust and error accordingly. It can be done by querying if the spi.Port implements https://periph.io/x/periph/host/sysfs#SPI.MaxTxSize

experimental/cmd/nrzled/main.go
```diff
@@ -140,27 +143,51 @@ func mainImpl() error {
 	if _, err := host.Init(); err != nil {
 		return err
 	}
+	var disp nrzled.Strip
+	{
```
maruel: Please remove the additional scoping here, I don't see value in this.

experimental/devices/nrzled/doc.go
```diff
@@ -8,6 +8,15 @@
 //
 // Note that some ICs are 7 bits with the least significant bit ignored, others
 // are using a real 8 bits PWM. The PWM frequency varies across ICs.
+// User can select a driver implementation using a memory-mapped GPIO line
+// or leverage a SPI connection's MISO pin.
+//
+// The SPI implementation is sensitive to variations in SPI clock speed.
+// On the Raspberry Pi, you will need to add `core_freq=250`
```
maruel: That's the kind of information that would eventually should be put on the web site; https://periph.io/device/ with a new nrzled entry that. For now it's fine.

250 is already the default on RPi1 and 2. So you mean for RPi3, Zero and later? Can you expand on the glitching issue?

experimental/devices/nrzled/doc.go
```diff
@@ -8,6 +8,15 @@
...
+// to /boot/config.txt to prevent glitching.
+//
+// You may also need to increase your SPI buffer size to 12*num_pixels+3, or just max it out
+// with `spidev.bufsize=65536`. That should allopw you to buffer over 5400 Neopixels.
```
maruel: [just FYI, not action]

I have a 256 LEDs board at home that I'm very eager to try out.

I also want to try out via an FT232H but hadn't had the chance to complete the driver yet; https://periph.io/x/extra/hostextra/d2xx

experimental/devices/nrzled/nrz_spi.go
```diff
@@ -0,0 +1,453 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
```
maruel: 2018

experimental/devices/nrzled/nrz_spi.go
```diff
@@ -0,0 +1,453 @@
...
+	"periph.io/x/periph/conn/display"
+	"periph.io/x/periph/conn/physic"
+	"periph.io/x/periph/conn/spi"
+)
+
+// SPIDev represents a strip of WS2812b LEDs as a strip connected over a SPI port.
+// It accepts a stream of raw RGB pixels and converts it to a bit pattern consistent
+// with the WS812b protocol.
+// Includes intensity and temperature correction.
+type SPIDev struct {
```
maruel: Oh I was thinking about using Dev and switching based on if `p` or `s` is nil. The runtime cost would be trivial, and the switch would only be at two locations: String(), Draw() and Write(), but Draw and Write() could call a small utility function instead. That would remove a *ton* of code duplication.

experimental/devices/nrzled/nrz_spi.go
```diff
@@ -0,0 +1,453 @@
...
+		// Offset into the output buffer using rect
+		d.raster(dst[4*rect.Min.X:], m.Pix[start:end], true)
+	}
+}
+
+var _ display.Drawer = &Dev{}
+
+// The bit lookup table converts a single byte into its 4 byte SPI symbol
+// 0 => 1000 and 1 => 1110
+var bitlut = [][]byte{
```
maruel: Make it an array instead of a slice, i.e. use:

    var bitlut = [...][4]byte{

The rationale is that it becomes a variable instead of a pointer, it enables the compiler faster checks.

experimental/devices/nrzled/nrzled.go
```diff
@@ -40,6 +40,12 @@ func NRZ(b byte) uint32 {
 	return out
 }
 
+//A Strip is the high level interface all hardware implementations conform to
```
maruel: Won't be needed by merging SPIDev into Dev.

experimental/devices/nrzled/nrz_spi.go
```diff
@@ -0,0 +1,453 @@
...
+	rawBuf    []byte          // Raw buffer sent over SPI. Cached to reduce heap fragmentation.
+	pixels    []byte          // Double buffer of pixels, to enable partial painting via Draw(). Effectively points inside rawBuf.
+	rect      image.Rectangle // Device bounds
+	pixelMap  map[image.Point]int
+}
+
+// NewSPI returns a strip that communicates over SPI to NRZ encoded LEDs.
+//
+// Due to the tight timing demands of these LEDs,
+// the SPI port speed must be a reliable 2.5MHz
```
maruel: Shouldn't it be 2.4MHz? I guess it doesn't work because of it being a multiple of the core's speed (?)
I know that the SPI port is rarely what is exactly asked. :/

experimental/devices/nrzled/nrz_spi.go
```diff
@@ -0,0 +1,453 @@
...
+func NewSPI(p spi.Port, o *Opts) (*SPIDev, error) {
+	c, err := p.Connect(2500*physic.KiloHertz, spi.Mode3|spi.NoCS, 8)
+	if err != nil {
+		return nil, err
+	}
+	rawBCt := 4 * (3 * o.NumPixels) //3 bytes per pixel, 4 symbol bytes per byte
+	buf := make([]byte, rawBCt+3)   //3 bytes for latch. 24*400ns = 9600ns.
+	tail := buf[rawBCt:]
+	for i := range tail {
+		tail[i] = 0x00
```
maruel: That's not needed, memory is always zero'ed out in Go unlike in C.

maruel:
Sorry for asking you to rewrite this PR twice but I think the end result is both 1) much less code and 2) simpler for the users. If you are still interested I'll be happy to review, otherwise I'll take it bite-meal, probably with NRZ first, then the rest.

experimental/devices/nrzled/doc.go
whmeitzler: I did my testing on a Pi0/0W, and saw random "sparkles" of white on my strip around 20% of the time. 
Turns out that, while  Neopixels are more [timing-tolerant](https://wp.josh.com/2014/05/13/ws2812-neopixels-are-not-so-finicky-once-you-get-to-know-them/) than most people expect, variations in the core freq threw off the timing enough to flip bits in the transmission. [This discussion](https://github.com/raspberrypi/linux/issues/2060) gave me the clue to lock core_freq

maruel: Thanks, this can be put it here for now, and moved later.

whmeitzler:
Pardon the delay on getting back to your review. I'm still interested in getting this merged, though I wrote it as a stepping stone for a larger project (Muxing several Neopixel streams on one SPI port) that's sucking up my hobby time. I think at this point offering the code to you piecemeal is the most expedient course of action. I'm also interested to see your iteration of the API design. 

The more I think about these drivers, the more I'm tempted to open a 'Add a generic RGBStrip interface' issue.

maruel:
Ok no worries. I'll merge it in but with no guarantee that I won't make follow up breaking changes. Is that ok?

maruel:
gohci

whmeitzler:
`no guarantee that I won't make follow up breaking changes. Is that ok?`
Absolutely, the code's all yours to play with. Thanks for your reviews and happy hacking!

PR #294, opened 2018-10-14
add more SI units

NeuralSpaz:
add Power as Watt
add Electrical Capacitance as farad
add Luminous Flux as lumen
add Luminous Intensity as candela
add Energy as joule
add easy to verify tests for new units and there prefixes(eg 1GJ == GigaJoule etc...)
add picoAsString to support ElectricalCapacitance
add picoAsString test


conn/physic/units.go
```diff
@@ -352,6 +352,114 @@ const (
...
+	NanoFarad  ElectricalCapacitance = 1000 * PicoFarad
+	MicroFarad ElectricalCapacitance = 1000 * NanoFarad
+	MilliFarad ElectricalCapacitance = 1000 * MicroFarad
+	Farad      ElectricalCapacitance = 1000 * MilliFarad
+	KiloFarad  ElectricalCapacitance = 1000 * Farad
+	MegaFarad  ElectricalCapacitance = 1000 * KiloFarad
+)
+
+// LuminousIntensity is a measurement of wavelength power stored as a
+// nano candela.
```
maruel: The comment should have a note that this is only about human visible wavelengths via a standardized luminosity function.
https://en.m.wikipedia.org/wiki/Luminosity_function
https://en.m.wikipedia.org/wiki/Photopic_vision

Something like ".. a measurement of luminous power per solid angle, on wavelengths based on 1931 CIE photopic luminosity function model of the human eye, stored as .."

conn/physic/units.go
```diff
@@ -352,6 +352,114 @@ const (
...
+	NanoCandela  LuminousIntensity = 1
+	MicroCandela LuminousIntensity = 1000 * NanoCandela
+	MilliCandela LuminousIntensity = 1000 * MicroCandela
+	Candela      LuminousIntensity = 1000 * MilliCandela
+	KiloCandela  LuminousIntensity = 1000 * Candela
+	MegaCandela  LuminousIntensity = 1000 * KiloCandela
+	GigaCandela  LuminousIntensity = 1000 * MegaCandela
+)
+
+// LuminousFlux is a measurement of luminous flux stored as a nano lumens.
```
maruel: ".. a measurement of the total quantity of visible light emitted by a source, stored as ..."

maruel:
You may want to rebase on top of origin/master to help with the PR.

conn/physic/units.go
NeuralSpaz: I'm not sure the clarity is there in my wording.

maruel:
gohci

conn/physic/units.go
maruel: It's a bit verbose but it will do, I'm fine with editing later.

PR #295, opened 2018-10-28
epd: add ePaper display support

alvarowolfx:
Initial ePaper display support for the Waveshare family of displays. I have the 2.13inch model, that is also a Raspberry Pi Hat. 
Product: https://www.waveshare.com/wiki/2.13inch_e-Paper_HAT

I almost ported from this implementation in Python:
https://github.com/soonuse/epd-library-python

img_5129

### DIFF
--- a/experimental/devices/epd/doc.go
+++ b/experimental/devices/epd/doc.go
@@ -2,15 +2,17 @@
 // Use of this source code is governed under the Apache License, Version 2.0
 // that can be found in the LICENSE file.
 
-// Package epd controls Waveshare e-paper series displays
+// Package epd controls Waveshare e-paper series displays.
 //
 // More details
 //
-// See https://periph.io/device/epd/ for more details about the device.
+// Datasheets
 //
-// Based on https://github.com/soonuse/epd-library-python
+// https://www.waveshare.com/w/upload/e/e6/2.13inch_e-Paper_Datasheet.pdf
 //
 // Product page:
-// https://www.waveshare.com/wiki/2.13inch_e-Paper_HAT
-// https://www.waveshare.com/wiki/1.54inch_e-Paper_Module
+//
+// 2.13 Inch version: https://www.waveshare.com/wiki/2.13inch_e-Paper_HAT
+//
+// 1.54 inch version: https://www.waveshare.com/wiki/1.54inch_e-Paper_Module
 package epd
--- a/experimental/devices/epd/doc.go
+++ b/experimental/devices/epd/doc.go
@@ -1,0 +1,16 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package epd controls Waveshare e-paper series displays
+//
+// More details
+//
+// See https://periph.io/device/epd/ for more details about the device.
+//
+// Based on https://github.com/soonuse/epd-library-python
+//
+// Product page:
+// https://www.waveshare.com/wiki/2.13inch_e-Paper_HAT
+// https://www.waveshare.com/wiki/1.54inch_e-Paper_Module
+package epd
--- a/experimental/devices/epd/epd.go
+++ b/experimental/devices/epd/epd.go
@@ -1,0 +1,463 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package epd
+
+import (
+	"errors"
+	"fmt"
+	"image"
+	"image/color"
+	"image/draw"
+	"time"
+
+	"periph.io/x/periph/conn"
+	"periph.io/x/periph/conn/display"
+	"periph.io/x/periph/conn/gpio"
+	"periph.io/x/periph/conn/gpio/gpioreg"
+	"periph.io/x/periph/conn/physic"
+	"periph.io/x/periph/conn/spi"
+	"periph.io/x/periph/devices/ssd1306/image1bit"
+)
+
+// Raspberry pi hat pins
+const (
+	ResetPin       = "17"
+	DataCommandPin = "25"
+	ChipSelectPin  = "8"
+	BusyStatePin   = "24"
+)
+
+// EPD commands
+const (
+	DriverOutputControl            = 0x01
+	BoosterSoftStartControl        = 0x0C
+	GateScanStartPosition          = 0x0F
+	DeepSleepMode                  = 0x10
+	DataEntryModeSetting           = 0x11
+	SwReset                        = 0x12
+	TemperatureSensorControl       = 0x1A
+	MasterActivation               = 0x20
+	DisplayUpdateControl1          = 0x21
+	DisplayUpdateControl2          = 0x22
+	WriteRAM                       = 0x24
+	WriteVcomRegister              = 0x2C
+	WriteLutRegister               = 0x32
+	SetDummyLinePeriod             = 0x3A
+	SetGateTime                    = 0x3B
+	BorderWaveformControl          = 0x3C
+	SetRAMXAddressStartEndPosition = 0x44
+	SetRAMYAddressStartEndPosition = 0x45
+	SetRAMXAddressCounter          = 0x4E
+	SetRAMYAddressCounter          = 0x4F
+	TerminateFrameReadWrite        = 0xFF
+)
+
+// LUT contains the display specific waveform for the pixel programming of the display.
+type LUT []byte
+
+// LUTType full or partial display update
+type LUTType int
+
+const (
+	// Full LUT config to update all the display
+	Full LUTType = iota
+	// Partial LUT config only a part of the display
+	Partial
+)
+
+// EPD2in13 is the config for the 2.13 inch display.
+var EPD2in13 = DisplayConfig{
+	W: 128,
+	H: 250,
+	FullUpdate: LUT{
+		0x22, 0x55, 0xAA, 0x55, 0xAA, 0x55, 0xAA, 0x11,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		0x1E, 0x1E, 0x1E, 0x1E, 0x1E, 0x1E, 0x1E, 0x1E,
+		0x01, 0x00, 0x00, 0x00, 0x00, 0x00,
+	},
+	PartialUpdate: LUT{
+		0x18, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		0x0F, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+	},
+}
+
+// EPD1in54 is the config for the 1.54 inch display.
+var EPD1in54 = DisplayConfig{
+	W: 200,
+	H: 200,
+	FullUpdate: LUT{
+		0x02, 0x02, 0x01, 0x11, 0x12, 0x12, 0x22, 0x22,
+		0x66, 0x69, 0x69, 0x59, 0x58, 0x99, 0x99, 0x88,
+		0x00, 0x00, 0x00, 0x00, 0xF8, 0xB4, 0x13, 0x51,
+		0x35, 0x51, 0x51, 0x19, 0x01, 0x00,
+	},
+	PartialUpdate: LUT{
+		0x10, 0x18, 0x18, 0x08, 0x18, 0x18, 0x08, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x13, 0x14, 0x44, 0x12,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+	},
+}
+
+// DisplayConfig defines the options for the ePaper Device.
+type DisplayConfig struct {
+	W             int
+	H             int
+	FullUpdate    LUT
+	PartialUpdate LUT
+}
+
+// NewSPI returns a Dev object that communicates over SPI to a E-Paper display controller.
+func NewSPI(p spi.Port, dc, cs, rst gpio.PinOut, busy gpio.PinIO, config *DisplayConfig) (*Dev, error) {
+	if dc == gpio.INVALID {
+		return nil, errors.New("epd: use nil for dc to use 3-wire mode, do not use gpio.INVALID")
+	}
+
+	if err := dc.Out(gpio.Low); err != nil {
+		return nil, err
+	}
+
+	c, err := p.Connect(2*physic.MegaHertz, spi.Mode0, 8)
+	if err != nil {
+		return nil, err
+	}
+
+	d := &Dev{
+		c:      c,
+		dc:     dc,
+		cs:     cs,
+		rst:    rst,
+		busy:   busy,
+		lt:     Full,
+		config: config,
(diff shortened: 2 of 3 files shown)


experimental/devices/epd/doc.go
```diff
@@ -0,0 +1,16 @@
...
+
+// Package epd controls Waveshare e-paper series displays
+//
+// More details
+//
+// See https://periph.io/device/epd/ for more details about the device.
+//
+// Based on https://github.com/soonuse/epd-library-python
+//
+// Product page:
```
maruel: Please ensure this looks nice in godoc, you can see locally with `godoc -http=localhost:6060` then navigate to http://localhost:6060/pkg/periph.io/x/periph/experimental/devices/epd

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,463 @@
...
+
+	return d, nil
+}
+
+// NewSPIHat returns a Dev object that communicates over SPI
+// and have the default config for the e-paper hat for raspberry pi
+func NewSPIHat(p spi.Port, config *DisplayConfig) (*Dev, error) {
+	dc := gpioreg.ByName(DataCommandPin)
+	cs := gpioreg.ByName(ChipSelectPin)
+	rst := gpioreg.ByName(ResetPin)
```
maruel: You don't need to use the GPIO registry here, you can use use rpi.P1_11 for this pin, and the corresponding for others: https://periph.io/x/periph/host/rpi#pkg-variables

Thanks for separating the hat support from the device support.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,463 @@
...
+	// send the color data
+	for i := 0; i < (w / 8 * h); i++ {
+		if err := d.sendData([]byte{color}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// DisplayFrame update the display
```
maruel: Please make this look nice in godoc.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,463 @@
...
+		return err
+	}
+
+	d.waitUntilIdle()
+	return nil
+}
+
+// Halt turns off the display.
+func (d *Dev) Halt() error {
+	return d.Sleep()
```
maruel: I don't think sleep clears the display?

experimental/devices/epd/doc.go
```diff
@@ -0,0 +1,16 @@
...
+// https://www.waveshare.com/wiki/2.13inch_e-Paper_HAT
+// https://www.waveshare.com/wiki/1.54inch_e-Paper_Module
```
maruel: What I find more useful is the datasheet; https://www.waveshare.com/w/upload/e/e6/2.13inch_e-Paper_Datasheet.pdf

For example, you can implement temperature reading (see page 9).

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,463 @@
...
+// Raspberry pi hat pins
+const (
+	ResetPin       = "17"
+	DataCommandPin = "25"
+	ChipSelectPin  = "8"
+	BusyStatePin   = "24"
+)
+
+// EPD commands
+const (
```
maruel: Please do not export them, the user is not expected to send raw commands to the device.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,463 @@
...
+	"periph.io/x/periph/conn/gpio"
+	"periph.io/x/periph/conn/gpio/gpioreg"
+	"periph.io/x/periph/conn/physic"
+	"periph.io/x/periph/conn/spi"
+	"periph.io/x/periph/devices/ssd1306/image1bit"
+)
+
+// Raspberry pi hat pins
+const (
+	ResetPin       = "17"
```
maruel: No need to export them, see below.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,463 @@
...
+	SetRAMXAddressCounter          = 0x4E
+	SetRAMYAddressCounter          = 0x4F
+	TerminateFrameReadWrite        = 0xFF
+)
+
+// LUT contains the display specific waveform for the pixel programming of the display.
+type LUT []byte
+
+// LUTType full or partial display update
+type LUTType int
```
maruel: Unless there's other types, a bool would be fine, like `PartialUpdate bool`

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,463 @@
...
+	},
+	PartialUpdate: LUT{
+		0x10, 0x18, 0x18, 0x08, 0x18, 0x18, 0x08, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x13, 0x14, 0x44, 0x12,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+	},
+}
+
+// DisplayConfig defines the options for the ePaper Device.
```
maruel: Name it Opts to be consistent with the other drivers.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,463 @@
...
+func NewSPI(p spi.Port, dc, cs, rst gpio.PinOut, busy gpio.PinIO, config *DisplayConfig) (*Dev, error) {
+	if dc == gpio.INVALID {
+		return nil, errors.New("epd: use nil for dc to use 3-wire mode, do not use gpio.INVALID")
+	}
+
+	if err := dc.Out(gpio.Low); err != nil {
+		return nil, err
+	}
+
+	c, err := p.Connect(2*physic.MegaHertz, spi.Mode0, 8)
```
maruel: The datasheet seems to imply 5MHz is fine on page 16.

experimental/devices/epd/doc.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package epd controls Waveshare e-paper series displays
+//
+// More details
+//
+// See https://periph.io/device/epd/ for more details about the device.
```
maruel: Maybe add this only once it's added?

experimental/devices/epd/epd.go
alvarowolfx: I changed the name of this type alias but I don't know if it became better to understand if someone takes a look at the code. If you can give more ideas on that would be nice.

experimental/devices/epd/epd.go
alvarowolfx: I'm not sure what is the best thing to do here. Indeed sending the device to sleep was not the best option, but now I'm clearing the screen. In what moment is recommended to make the call to `Halt`, because in the case of the ePaper display, the image will stay there even after powering off the display, so I don't think that the user wants to clear the display.

alvarowolfx:
Thanks for the comments, I made some changes and pushed here. I'll be sending more drivers during this week.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+		return nil, err
+	}
+
+	return d, nil
+}
+
+// NewSPIHat returns a Dev object that communicates over SPI
+// and have the default config for the e-paper hat for raspberry pi
+func NewSPIHat(p spi.Port, opts *Opts) (*Dev, error) {
+	dc := rpi.P1_22
```
maruel: You can inline, this will fit the line.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+	if yStart+imageH >= h {
+		yEnd = h - 1
+	}
+
+	d.setMemoryArea(xStart, yStart, xEnd, yEnd)
+
+	next := image1bit.NewVerticalLSB(d.rect)
+	draw.Src.Draw(next, r, src, sp)
+	var byteToSend byte = 0x00
+	for y := yStart; y < yEnd+1; y++ {
```
maruel: y <= yEnd?

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+		d.setMemoryPointer(xStart, y)
+		if err := d.sendCommand([]byte{writeRAM}); err != nil {
+			return err
+		}
+		for x := xStart; x < xEnd+1; x++ {
```
maruel: x <= xEnd?

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+func (d *Dev) DisplayFrame() error {
+	if err := d.sendCommand([]byte{displayUpdateControl2}); err != nil {
+		return err
+	}
+
+	if err := d.sendData([]byte{byte(0xC4)}); err != nil {
+		return err
+	}
+
+	if err := d.sendCommand([]byte{masterActivation}); err != nil {
```
maruel: Can you combine this command and the next one?

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+// Init initialize the display config. This method is already called when creating
+// a device using NewSPI and NewSPIHat methods.
+//
+// It should be only used when you put the device to sleep and need to re-init the device.
+func (d *Dev) Init() error {
+	if err := d.sendCommand([]byte{driverOutputControl}); err != nil {
+		return err
+	}
+
+	if err := d.sendData([]byte{byte((d.opts.H - 1) & 0xFF)}); err != nil {
```
maruel: I'm pretty sure you can concatenate the sendData() calls here.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+
+	if err := d.sendData([]byte{byte(0x00)}); err != nil {
+		return err
+	}
+
+	if err := d.sendCommand([]byte{boosterSoftStartControl}); err != nil {
+		return err
+	}
+
+	if err := d.sendData([]byte{byte(0xD7)}); err != nil {
```
maruel: Same

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+		return err
+	}
+
+	d.waitUntilIdle()
+
+	return nil
+}
+
+func (d *Dev) waitUntilIdle() {
+	for d.busy.Read() == gpio.High {
```
maruel: That'd be a good place for WaitForEdge() instead of polling.

maruel:
gohci

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+	FullUpdate    LUT
+	PartialUpdate LUT
+}
+
+// NewSPI returns a Dev object that communicates over SPI to a E-Paper display controller.
+func NewSPI(p spi.Port, dc, cs, rst gpio.PinOut, busy gpio.PinIO, opts *Opts) (*Dev, error) {
+	if dc == gpio.INVALID {
+		return nil, errors.New("epd: use nil for dc to use 3-wire mode, do not use gpio.INVALID")
+	}
+
```
maruel: call busy.In()

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+// Opts defines the options for the ePaper Device.
+type Opts struct {
+	W             int
+	H             int
+	FullUpdate    LUT
+	PartialUpdate LUT
+}
+
+// NewSPI returns a Dev object that communicates over SPI to a E-Paper display controller.
+func NewSPI(p spi.Port, dc, cs, rst gpio.PinOut, busy gpio.PinIO, opts *Opts) (*Dev, error) {
```
maruel: cs is never used, why is it specified? I'd remove this argument.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+	if xStart+imageW >= w {
+		xEnd = w - 1
+	}
+
+	yEnd := yStart + imageH - 1
+	if yStart+imageH >= h {
+		yEnd = h - 1
+	}
+
+	d.setMemoryArea(xStart, yStart, xEnd, yEnd)
```
maruel: This function returns an error, please check it.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+
+	next := image1bit.NewVerticalLSB(d.rect)
+	draw.Src.Draw(next, r, src, sp)
+	var byteToSend byte = 0x00
+	for y := yStart; y < yEnd+1; y++ {
+		d.setMemoryPointer(xStart, y)
```
maruel: Same

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+
+	return nil
+}
+
+// ClearFrameMemory clear the frame memory with the specified color.
+// this won't update the display.
+func (d *Dev) ClearFrameMemory(color byte) error {
+	w := d.rect.Dx()
+	h := d.rect.Dy()
+	d.setMemoryArea(0, 0, w-1, h-1)
```
maruel: Same

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+		return err
+	}
+	return d.c.Tx(c, nil)
+}
+
+func (d *Dev) sendCommand(c []byte) error {
+	if err := d.dc.Out(gpio.Low); err != nil {
+		return err
+	}
+	return d.c.Tx(c, nil)
```
maruel: I'd recommand to keep dc High after the call, so that consecutive sendData() calls like at line 451 are faster since sendData() doesn't need to touch d.dc at all.

Or keep track of the level in the object.

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+	if d.update == Partial {
+		lut = d.opts.PartialUpdate
+	}
+
+	if err := d.sendCommand([]byte{writeLutRegister}); err != nil {
+		return err
+	}
+
+	for i := range lut {
+		d.sendData([]byte{lut[i]})
```
maruel: Why not:

    return d.sendData(lut)

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+
+	if err := d.sendCommand([]byte{dataEntryModeSetting}); err != nil {
+		return err
+	}
+
+	if err := d.sendData([]byte{byte(0x03)}); err != nil {
+		return err
+	}
+
+	if err := d.setLut(Full); err != nil {
```
maruel: There's never a call to setLut(Partial). Is that intended?

experimental/devices/epd/epd.go
```diff
@@ -0,0 +1,470 @@
...
+// NewSPI returns a Dev object that communicates over SPI to a E-Paper display controller.
+func NewSPI(p spi.Port, dc, cs, rst gpio.PinOut, busy gpio.PinIO, opts *Opts) (*Dev, error) {
+	if dc == gpio.INVALID {
+		return nil, errors.New("epd: use nil for dc to use 3-wire mode, do not use gpio.INVALID")
+	}
+
+	if err := dc.Out(gpio.Low); err != nil {
+		return nil, err
+	}
+
```
maruel: Also call rst.Out()

experimental/devices/epd/example_test.go
```diff
@@ -0,0 +1,123 @@
...
+	// }
+	// drawer.DrawString("Hello from periph!")
+
+	if err := dev.Draw(dev.Bounds(), img, image.Point{}); err != nil {
+		log.Fatal(err)
+	}
+	dev.DisplayFrame() // After drawing on the display, you have to show the frame
+}
+
+func Example_other() {
```
maruel: I'm not sure you can use a lower case for the first word, please confirm how it shows up in godoc.

There's no blockers anymore so I'll integrate but please address the comments in a follow up.

PR #296, opened 2018-10-28
[feat] Add support for pca9685 pwm module

alvarowolfx:
Add support for the PCA9685 16 channel PWM module. I create this for a demo on Gophercon Brazil talking about embedded development with Golang and one of the demos was a robot arm (MeArm) controlled through Google Assistant, Cloud IoT Core and Golang running on an Orange Pi Zero. So I extracted the code for controlling the pca9685 module and I'm sending this PR.

Code for the project: https://github.com/alvarowolfx/golang-voice-iot
Demo of it working: https://youtu.be/Xep6dZ3xOYA

Based on https://github.com/wintersandroid/Android-Things-PCA9685
Datasheets: https://cdn-shop.adafruit.com/datasheets/PCA9685.pdf
Product page: https://www.adafruit.com/product/815

img_4752

### DIFF
--- a/experimental/devices/pca9685/doc.go
+++ b/experimental/devices/pca9685/doc.go
@@ -1,0 +1,17 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package pca9685 includes utilities to controls pca9685 module and servo motors.
+// More details
+//
+// See https://periph.io/device/pca9685/ for more details about the device.
+//
+// Based on https://github.com/wintersandroid/Android-Things-PCA9685
+//
+// Datasheets:
+// https://cdn-shop.adafruit.com/datasheets/PCA9685.pdf
+//
+// Product page:
+// https://www.adafruit.com/product/815
+package pca9685
--- a/experimental/devices/pca9685/doc.go
+++ b/experimental/devices/pca9685/doc.go
@@ -3,15 +3,14 @@
 // that can be found in the LICENSE file.
 
 // Package pca9685 includes utilities to controls pca9685 module and servo motors.
+//
 // More details
 //
-// See https://periph.io/device/pca9685/ for more details about the device.
+// Datasheet
 //
-// Based on https://github.com/wintersandroid/Android-Things-PCA9685
-//
-// Datasheets:
-// https://cdn-shop.adafruit.com/datasheets/PCA9685.pdf
+// https://www.nxp.com/docs/en/data-sheet/PCA9685.pdf
 //
 // Product page:
-// https://www.adafruit.com/product/815
+//
+// https://www.nxp.com/products/analog/interfaces/ic-bus/ic-led-controllers/16-channel-12-bit-pwm-fm-plus-ic-bus-led-controller:PCA9685
 package pca9685
--- a/experimental/devices/pca9685/pca9685.go
+++ b/experimental/devices/pca9685/pca9685.go
@@ -13,32 +13,32 @@ import (
 	"periph.io/x/periph/conn/i2c"
 )
 
-// PCA9685Address i2c default address
-const PCA9685Address uint16 = 0x40
+// I2CAddr i2c default address.
+const I2CAddr uint16 = 0x40
 
 // PCA9685 Commands
 const (
-	Mode1      byte = 0x00
-	Mode2      byte = 0x01
-	SubAdr1    byte = 0x02
-	SubAdr2    byte = 0x03
-	SubAdr3    byte = 0x04
-	Prescale   byte = 0xFE
-	Led0OnL    byte = 0x06
-	Led0OnH    byte = 0x07
-	Led0OffL   byte = 0x08
-	Led0OffH   byte = 0x09
-	AllLedOnL  byte = 0xFA
-	AllLedOnH  byte = 0xFB
-	AllLedOffL byte = 0xFC
-	AllLedOffH byte = 0xFD
+	mode1      byte = 0x00
+	mode2      byte = 0x01
+	subAdr1    byte = 0x02
+	subAdr2    byte = 0x03
+	subAdr3    byte = 0x04
+	prescale   byte = 0xFE
+	led0OnL    byte = 0x06
+	led0OnH    byte = 0x07
+	led0OffL   byte = 0x08
+	led0OffH   byte = 0x09
+	allLedOnL  byte = 0xFA
+	allLedOnH  byte = 0xFB
+	allLedOffL byte = 0xFC
+	allLedOffH byte = 0xFD
 
 	// Bits
-	Restart byte = 0x80
-	Sleep   byte = 0x10
-	AllCall byte = 0x01
-	Invrt   byte = 0x10
-	OutDrv  byte = 0x04
+	restart byte = 0x80
+	sleep   byte = 0x10
+	allCall byte = 0x01
+	invrt   byte = 0x10
+	outDrv  byte = 0x04
 )
 
 // Dev is a handler to pca9685 controller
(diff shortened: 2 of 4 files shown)


experimental/devices/pca9685/doc.go
```diff
@@ -0,0 +1,17 @@
...
+// that can be found in the LICENSE file.
+
+// Package pca9685 includes utilities to controls pca9685 module and servo motors.
+// More details
+//
+// See https://periph.io/device/pca9685/ for more details about the device.
+//
+// Based on https://github.com/wintersandroid/Android-Things-PCA9685
+//
+// Datasheets:
```
maruel: For godoc to process this correctly, you need to keep an empty line between and no semi colon, e.g.:

```
// Datasheet
//
// https://www.nxp.com/docs/en/data-sheet/PCA9685.pdf
```

Use the original datasheet, not the copy from Adafruit.

experimental/devices/pca9685/doc.go
```diff
@@ -0,0 +1,17 @@
...
+// https://cdn-shop.adafruit.com/datasheets/PCA9685.pdf
+//
+// Product page:
+// https://www.adafruit.com/product/815
```
maruel: The product page is https://www.nxp.com/products/analog/interfaces/ic-bus/ic-led-controllers/16-channel-12-bit-pwm-fm-plus-ic-bus-led-controller:PCA9685

I do link to adafruit on the website in the "where to buy" section but not in the documentation.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,127 @@
...
+	"periph.io/x/periph/conn/physic"
+
+	"periph.io/x/periph/conn/i2c"
+)
+
+// PCA9685Address i2c default address
+const PCA9685Address uint16 = 0x40
+
+// PCA9685 Commands
+const (
```
maruel: Please do not export these.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,127 @@
...
+	Invrt   byte = 0x10
+	OutDrv  byte = 0x04
+)
+
+// Dev is a handler to pca9685 controller
+type Dev struct {
+	dev *i2c.Dev
+}
+
+// NewI2CAddress returns a Dev object that communicates over I2C on a alternate address
```
maruel: Merge these two functions, have the caller specify `pca9685.I2CAddr` every time and comment that it's what should be normally passed.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"math"
+	"time"
+
+	"periph.io/x/periph/conn/physic"
+
+	"periph.io/x/periph/conn/i2c"
+)
+
+// PCA9685Address i2c default address
```
maruel: There's a bit of inconsistency in the drivers for default I2C addresses, I think I'd prefer `I2CAddr` here, since PCA9685Address stutters.

https://blog.golang.org/package-names talks a bit about this.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (d *Dev) init() error {
+	d.SetAllPwm(0, 0)
+	d.dev.Write([]byte{Mode2, OutDrv})
+	d.dev.Write([]byte{Mode1, AllCall})
+
+	time.Sleep(100 * time.Millisecond)
+
+	var mode1 byte
+	err := d.dev.Tx([]byte{Mode1}, []byte{mode1})
```
maruel: mode1 => 0 ?

```
if err := d.dev.Tx([]byte{Mode1}, []byte{0}); err != nil {
  return err
}
```

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+
+	mode1 = mode1 & ^Sleep
+	d.dev.Write([]byte{Mode1, mode1 & 0xFF})
+
+	time.Sleep(5 * time.Millisecond)
+
+	err = d.SetPwmFreq(50)
```
maruel: return d.SetPwmFreq(50*physic.Hertz)

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,127 @@
...
+	return err
+}
+
+// SetPwmFreq set the pwm frequency
+func (d *Dev) SetPwmFreq(freqHz float32) error {
```
maruel: Do not use floating point, use physic.Frequency instead.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,127 @@
...
+	Restart byte = 0x80
+	Sleep   byte = 0x10
+	AllCall byte = 0x01
+	Invrt   byte = 0x10
+	OutDrv  byte = 0x04
+)
+
+// Dev is a handler to pca9685 controller
+type Dev struct {
+	dev *i2c.Dev
```
maruel: I think it'd be nice for the Dev to have 16 gpio.PinIO members.
Make all the PinIn functions fail but still implement it as this is potentially useful compared to implementing only gpio.PinOut.

experimental/devices/pca9685/servo.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Servo individual servo from a group of servos connected to a pca9685 module
+type Servo struct {
+	group    *ServoGroup
+	channel  int
+	minAngle int
+	maxAngle int
+}
+
+// NewServoGroup returns a servo group connected throught the pca9685 module
```
maruel: Please use https://periph.io/x/periph/conn/physic#Angle

That said, the angle-to-period logic is fairly device agnostic so the mapValue() code should probably live elsewhere. I started package https://periph.io/x/periph/experimental/devices/gpioutil for this kind of things. I plan to move it under conn/gpio/ at the end.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// NewI2CAddress returns a Dev object that communicates over I2C on a alternate address
+func NewI2CAddress(bus i2c.Bus, address uint16) (*Dev, error) {
+	dev := &Dev{
+		dev: &i2c.Dev{Bus: bus, Addr: address},
+	}
+	err := dev.init()
+	return dev, err
```
maruel: Return nil for *Dev if err != nil

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// NewI2C returns a Dev object that communicates over I2C on the default address
+func NewI2C(bus i2c.Bus) (*Dev, error) {
+	return NewI2CAddress(bus, PCA9685Address)
+}
+
+func (d *Dev) init() error {
+	d.SetAllPwm(0, 0)
```
maruel: Please check errors thoroughly.

experimental/devices/pca9685/pca9685.go
alvarowolfx: The code was supposed to read the mode from the device, I changed it to pass a slice and read back that.

experimental/devices/pca9685/pca9685.go
alvarowolfx: I didn't make that change yet, will need a bit more time to come back and improve that. But would be a really good addition to the driver.

experimental/devices/pca9685/servo.go
alvarowolfx: The problem is: how to implement a generic `mapValue` function? Because was initially implemented for `int` values and now is a mix of ints and `physic.Angle`, so I don't know how can we make a more generic function that can be used in other places.

alvarowolfx:
Thanks for the comments, already pushed some updates and need to work more on some ideas.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,166 @@
...
+		return err
+	}
+	if _, err := d.dev.Write([]byte{allLedOffH, byte(off >> 8)}); err != nil {
+		return err
+	}
+	return nil
+}
+
+// SetPwm set a pwm value for given pca9685 channel
+func (d *Dev) SetPwm(channel int, on uint16, off uint16) error {
```
maruel: Ultimately, I'll ask you to use `gpio.Duty` for PWM duty cycle values. This can be done in a follow up.
https://periph.io/x/periph/conn/gpio#Duty

maruel:
gohci

maruel:
Please fix the test, thanks.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,166 @@
...
+	modeRead = make([]byte, 1)
+	err := d.dev.Tx([]byte{mode1}, modeRead)
+
+	if err != nil {
+		return err
+	}
+
+	oldmode := modeRead[0]
+	newmode := (byte)((oldmode & 0x7F) | 0x10)                     // sleep
+	if _, err := d.dev.Write([]byte{mode1, newmode}); err != nil { // go to sleep;
```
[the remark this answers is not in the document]
alvarowolfx: I tried sending commands altogether instead of doing 3 call `to dev.Write` and unfortunately didn't work on the device.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,150 @@
...
+	led0OnL    byte = 0x06
+	led0OnH    byte = 0x07
+	led0OffL   byte = 0x08
+	led0OffH   byte = 0x09
+	allLedOnL  byte = 0xFA
+	allLedOnH  byte = 0xFB
+	allLedOffL byte = 0xFC
+	allLedOffH byte = 0xFD
+
+	// Bits
```
maruel: In a follow up, I think it's worth describing what they are for. This is mixing bitmasks for mode2 (ourDrv, invrt), mode1 (allCall, sleep) and restart is unused and I couldn't find it in the datasheet.

So remove restart for now, we'll handle the rest later.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return dev, nil
+}
+
+func (d *Dev) init() error {
+	if err := d.SetAllPwm(0, 0); err != nil {
+		return err
+	}
+
+	if _, err := d.dev.Write([]byte{mode2, outDrv}); err != nil {
```
maruel: I looked at the datasheet and the way it works is via a memory mapped system where the pointer increments at each byte. That's fairly standard with I²C and https://periph.io/x/periph/conn/mmr exists exactly for this.

That said we should do that in a follow up, let's get something working first.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,150 @@
...
+	if _, err := d.dev.Write([]byte{mode1, mode}); err != nil {
+		return err
+	}
+
+	time.Sleep(5 * time.Millisecond)
+
+	return d.SetPwmFreq(50 * physic.Hertz)
+}
+
+// SetPwmFreq set the pwm frequency
```
maruel: In a follow up, please terminate all documentation with a period.

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,150 @@
...
+	allLedOnH  byte = 0xFB
+	allLedOffL byte = 0xFC
+	allLedOffH byte = 0xFD
+
+	// Bits
+	restart byte = 0x80
+	sleep   byte = 0x10
+	allCall byte = 0x01
+	invrt   byte = 0x10
+	outDrv  byte = 0x04
```
maruel: This comment is only FYI.

This device can sink 25mA and in totem-pole can source 10mA. You always put it in totem-pole at the moment.
I started adding `Drive()` to some GPIOs (like https://periph.io/x/periph/host/bcm283x#Pin.Drive) but I haven't defined an interface to gpio yet. I realize I should probably have both Sink and Source instead, so maybe I'll change that.

maruel:
gohci

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+// Dev is a handler to pca9685 controller
+type Dev struct {
```
maruel: In a follow up, please add Halt() that sets sleep bit, so that all PWMs stop. That's the essence of Halt. :)

experimental/devices/pca9685/pca9685.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+// I2CAddr i2c default address.
+const I2CAddr uint16 = 0x40
+
+// PCA9685 Commands
+const (
+	mode1      byte = 0x00
+	mode2      byte = 0x01
+	subAdr1    byte = 0x02
```
maruel: These are bitmasks on mode1. I'd vote to remove them since they are not used and you are mixing register addresses and bitmasks here, which is super confusing when looking up the datasheet.

Well, there *are* subadr1~3 registers, but they are 0x19 to 01A.

alvarowolfx:
Thanks for the awesome help @maruel. I'm still fairly new to Golang and I'm learning a lot with your reviews. I'll be fixing the other drivers that I sent and will be probably sending another one later today (bh1750 light sensor driver). Them I can revisit those drivers. You are completely right, I should stop following blindly and porting from other languages and start looking more into the datasheet and understanding more how it works.

maruel:
No worries, you did well. Reading datasheets is an art in itself so there's a ramp up curve there too.

PR #297, opened 2018-10-28
A few changes for mfrc522

vkorn:
+ Antenna Gain
* Terminating Wait if Halt received
* Renamed SetOperationtimeout to SetOperationTimeout

### DIFF
--- a/experimental/devices/mfrc522/mfrc522.go
+++ b/experimental/devices/mfrc522/mfrc522.go
@@ -119,6 +119,9 @@ func NewSPI(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn) (*Dev, er
 		operationTimeout: 30 * time.Second,
 		irqPin:           irqPin,
 		resetPin:         resetPin,
+		antennaGain:      4,
+		stop:             make(chan interface{}, 1),
+		isStopped:        false,
 	}
 	if err := dev.Init(); err != nil {
 		return nil, err
@@ -128,10 +131,13 @@ func NewSPI(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn) (*Dev, er
 
 // Dev is an handle to an MFRC522 RFID reader.
 type Dev struct {
+	antennaGain      int
 	resetPin         gpio.PinOut
 	irqPin           gpio.PinIn
 	operationTimeout time.Duration
 	spiDev           spi.Conn
+	isStopped        bool
+	stop             chan interface{}
 }
 
 // String implements conn.Resource.
@@ -144,16 +150,19 @@ func (r *Dev) String() string {
 //
 // It soft-stops the chip - PowerDown bit set, command IDLE
 func (r *Dev) Halt() error {
+	r.isStopped = true
+	r.stop <- true
+	close(r.stop)
 	return r.devWrite(commands.CommandReg, 16)
 }
 
-// SetOperationtimeout updates the device timeout for card operations.
+// SetOperationTimeout updates the device timeout for card operations.
 //
 // Effectively that sets the maximum time the RFID device will wait for IRQ
 // from the proximity card detection.
 //
 //	timeout the duration to wait for IRQ strobe.
-func (r *Dev) SetOperationtimeout(timeout time.Duration) {
+func (r *Dev) SetOperationTimeout(timeout time.Duration) {
 	r.operationTimeout = timeout
 }
 
@@ -165,6 +174,11 @@ func (r *Dev) Init() error {
 	if err := r.writeCommandSequence(sequenceCommands.init); err != nil {
 		return err
 	}
+
+	if err := r.devWrite(int(commands.RFCfgReg), byte(r.antennaGain)<<4); err != nil {
+		return err
+	}
+
 	return r.SetAntenna(true)
 }
 
@@ -186,6 +200,15 @@ func (r *Dev) SetAntenna(state bool) error {
 		return r.setBitmask(commands.TxControlReg, 0x03)
 	}
 	return r.clearBitmask(commands.TxControlReg, 0x03)
+}
+
+// SetAntennaGain configures antenna signal strength.
+//
+//	gain - signal strength from 0 to 7.
+func (r *Dev) SetAntennaGain(gain int) {
+	if 0 <= gain && gain <= 7 {
+		r.antennaGain = gain
+	}
 }
 
 // CardWrite the low-level interface to write some raw commands to the card.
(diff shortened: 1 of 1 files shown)


experimental/devices/mfrc522/mfrc522.go
```diff
@@ -144,16 +150,19 @@ func (r *Dev) String() string {
 //
 // It soft-stops the chip - PowerDown bit set, command IDLE
 func (r *Dev) Halt() error {
+	r.isStopped = true
+	r.stop <- true
+	close(r.stop)
```
maruel: Halt is not Close, it should not crash when called repeatedly.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -323,7 +346,12 @@ func (r *Dev) Request() (int, error) {
 func (r *Dev) Wait() error {
 	irqChannel := make(chan bool)
 	go func() {
-		irqChannel <- r.irqPin.WaitForEdge(r.operationTimeout)
+		result := r.irqPin.WaitForEdge(r.operationTimeout)
+		if r.isStopped {
```
maruel: Assuming it's called in a separate thread, this is a memory access without lock. Modification to isStopped should be guarded by `mu sync.Mutex`

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -128,10 +131,13 @@ func NewSPI(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn) (*Dev, er
 
 // Dev is an handle to an MFRC522 RFID reader.
 type Dev struct {
+	antennaGain      int
 	resetPin         gpio.PinOut
 	irqPin           gpio.PinIn
 	operationTimeout time.Duration
 	spiDev           spi.Conn
+	isStopped        bool
```
maruel: move it to the bottom, guarded by a mutex

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -128,10 +131,13 @@ func NewSPI(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn) (*Dev, er
 
 // Dev is an handle to an MFRC522 RFID reader.
 type Dev struct {
+	antennaGain      int
```
maruel: put all mutable members at the bottom together

maruel:
Thanks! A few comments. I'm mainly concerned because I assume the object is used from multiple goroutines.

vkorn:
Ok, I was missing a point about Halt and that it might be called multiple times. 
Now it won't fail with couple consecutive calls.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -119,6 +120,8 @@ func NewSPI(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn) (*Dev, er
 		operationTimeout: 30 * time.Second,
 		irqPin:           irqPin,
 		resetPin:         resetPin,
+		antennaGain:      4,
+		stop:             make(chan interface{}, 1),
```
maruel: Use struct{} instead; it takes 0 byte instead of a pointer.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -132,6 +135,11 @@ type Dev struct {
 	irqPin           gpio.PinIn
 	operationTimeout time.Duration
 	spiDev           spi.Conn
+	stop             chan interface{}
```
maruel: Use struct{} instead; it takes 0 byte instead of a pointer.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -188,6 +208,15 @@ func (r *Dev) SetAntenna(state bool) error {
 	return r.clearBitmask(commands.TxControlReg, 0x03)
 }
 
+// SetAntennaGain configures antenna signal strength.
+//
+//	gain - signal strength from 0 to 7.
+func (r *Dev) SetAntennaGain(gain int) {
+	if 0 <= gain && gain <= 7 {
```
maruel: I think you should to protect it via the mutex then, since I suspect the caller may be in a separate goroutine.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -321,12 +350,28 @@ func (r *Dev) Request() (int, error) {
 
 // Wait wait for IRQ to strobe on the IRQ pin when the card is detected.
 func (r *Dev) Wait() error {
+	r.mu.Lock()
```
maruel: You absolutely need to do:
```
select {
  case <- r.stop:
  default:
}
```

before, but even so there's a small risk of deadlock if multiple Halt() are called concurrently, but that's not really a supported use case so I think you're fine here.

Another option is to use a condition variable; https://golang.org/pkg/sync/#Cond

Actually, the more I look at your goroutines, the more I think a sync.Cond is the right locking object since you really want to pass events.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -132,6 +135,11 @@ type Dev struct {
 	irqPin           gpio.PinIn
 	operationTimeout time.Duration
 	spiDev           spi.Conn
+	stop             chan interface{}
+	antennaGain      int
```
maruel: See my comment below about protecting this variable.

maruel:
gohci

experimental/devices/mfrc522/mfrc522.go
vkorn: I don't think it can actually hurt but added a mutex.

vkorn:
Current checks are covering every crazy multi-routine calls to Halt, Wait I could think of.
But just to make sure we're on the same page: from the beginning, this is not a concurrent-safe implementation. You most likely will lock the device by trying concurrent read/writes. Since signals must be sent in a specific order.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -181,7 +188,11 @@ func (r *Dev) Init() error {
 		return err
 	}
 
-	if err := r.devWrite(int(commands.RFCfgReg), byte(r.antennaGain)<<4); err != nil {
+	r.aMu.Lock()
+	err := r.devWrite(int(commands.RFCfgReg), byte(r.antennaGain)<<4)
```
maruel: it's faster to do:

```
  r.aMu.Lock()
  gain := r.antennaGain
  r.aMu.Unlock()
  if err := r.devWrite(int(commands.RFCfgReg), byte(gain)<<4) {
    return err
  }
```

otherwise the call itself is locking

maruel:
gohci

vkorn:
Potentially we could make low-level methods (Init, Wait) private then it should be fine with concurrent access. Or this should be added to documentation: to not mix high-level (Read/Write) with low-level.

maruel:
Yeah moving more methods private in a follow up is a good idea, if possible.

PR #298, opened 2018-10-29
ht16k33: add support for LED alphanumeric display

alvarowolfx:
Add support for the Pimoroni's Rainbow HAT with a lot of peripherals. I also created the driver for the HT16K33 that drives the Alphanumeric display on the hat.

Product page: https://shop.pimoroni.com/products/rainbow-hat-for-android-things
img_5133 trim 2018-10-28 22_59_36

### DIFF
--- a/experimental/devices/ht16k33/alphanum.go
+++ b/experimental/devices/ht16k33/alphanum.go
@@ -5,9 +5,6 @@
 package ht16k33
 
 import (
-	"fmt"
-	"strconv"
-
 	"periph.io/x/periph/conn/i2c"
 )
 
@@ -122,9 +119,7 @@ func NewAlphaNumericDisplay(bus i2c.Bus, address uint16) (*Display, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	display := &Display{dev: dev}
-	return display, nil
+	return &Display{dev: dev}, nil
 }
 
 // SetDigit at position to provided value.
@@ -136,50 +131,37 @@ func (d *Display) SetDigit(pos int, digit rune, decimal bool) error {
 	return d.dev.WriteColumn(pos, val)
 }
 
-// DisplayString print string of values to the display.
+// WriteString print string of values to the display.
 //
 // Characters in the string should be any ASCII value 32 to 127 (printable ASCII).
-func (d *Display) DisplayString(value string, justifyRight bool) error {
-	err := d.dev.Clear()
-	if err != nil {
-		return err
+func (d *Display) WriteString(s string) (int, error) {
+	if err := d.dev.Halt(); err != nil {
+		return 0, err
 	}
-	// Calculate starting position of digits based on justification.
-	pos := (4 - len(value))
-	if !justifyRight || pos < 0 {
+
+	pos := (4 - len(s))
+	if pos < 0 {
 		pos = 0
 	}
 	// Go through each character and print it on the display.
-	for _, ch := range value {
+	for _, ch := range s {
 		if ch == '.' {
 			// Print decimal points on the previous digit.
-			c := rune(value[pos-1])
+			c := rune(s[pos-1])
 			if err := d.SetDigit(pos-1, c, true); err != nil {
-				return err
+				return pos, err
 			}
 		} else {
 			if err := d.SetDigit(pos, ch, false); err != nil {
-				return err
+				return pos, err
 			}
 			pos++
 		}
 	}
-	return nil
-}
-
-// DisplayInt print a string of numeric values to the display.
-func (d *Display) DisplayInt(value int, justifyRight bool) error {
-	str := strconv.Itoa(value)
-	return d.DisplayString(str, justifyRight)
-}
-
-// DisplayFloat print a string of numeric values to the display.
-func (d *Display) DisplayFloat(value float64, justifyRight bool) error {
-	str := fmt.Sprintf("%5f", value)
-	return d.DisplayString(str, justifyRight)
+	return pos, nil
 }
 
 // Halt clear all the display.
 func (d *Display) Halt() error {
-	return d.dev.Clear()
+	return d.dev.Halt()
 }
--- a/experimental/devices/ht16k33/alphanum.go
+++ b/experimental/devices/ht16k33/alphanum.go
@@ -1,0 +1,185 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package ht16k33
+
+import (
+	"fmt"
+	"strconv"
+
+	"periph.io/x/periph/conn/i2c"
+)
+
+var digitValues = map[rune]uint16{
+	' ':  0x0,
+	'!':  0x6,
+	'"':  0x220,
+	'#':  0x12ce,
+	'$':  0x12ed,
+	'%':  0xc24,
+	'&':  0x235d,
+	'\'': 0x400,
+	'(':  0x2400,
+	')':  0x900,
+	'*':  0x3fc0,
+	'+':  0x12c0,
+	',':  0x800,
+	'-':  0xc0,
+	'.':  0x4000,
+	'/':  0xc00,
+	'0':  0xc3f,
+	'1':  0x6,
+	'2':  0xdb,
+	'3':  0x8f,
+	'4':  0xe6,
+	'5':  0x2069,
+	'6':  0xfd,
+	'7':  0x7,
+	'8':  0xff,
+	'9':  0xef,
+	':':  0x1200,
+	';':  0xa00,
+	'<':  0x2400,
+	'=':  0xc8,
+	'>':  0x900,
+	'?':  0x1083,
+	'@':  0x2bb,
+	'A':  0xf7,
+	'B':  0x128f,
+	'C':  0x39,
+	'D':  0x120f,
+	'E':  0xf9,
+	'F':  0x71,
+	'G':  0xbd,
+	'H':  0xf6,
+	'I':  0x1200,
+	'J':  0x1e,
+	'K':  0x2470,
+	'L':  0x38,
+	'M':  0x536,
+	'N':  0x2136,
+	'O':  0x3f,
+	'P':  0xf3,
+	'Q':  0x203f,
+	'R':  0x20f3,
+	'S':  0xed,
+	'T':  0x1201,
+	'U':  0x3e,
+	'V':  0xc30,
+	'W':  0x2836,
+	'X':  0x2d00,
+	'Y':  0x1500,
+	'Z':  0xc09,
+	'[':  0x39,
+	'\\': 0x2100,
+	']':  0xf,
+	'^':  0xc03,
+	'_':  0x8,
+	'`':  0x100,
+	'a':  0x1058,
+	'b':  0x2078,
+	'c':  0xd8,
+	'd':  0x88e,
+	'e':  0x858,
+	'f':  0x71,
+	'g':  0x48e,
+	'h':  0x1070,
+	'i':  0x1000,
+	'j':  0xe,
+	'k':  0x3600,
+	'l':  0x30,
+	'm':  0x10d4,
+	'n':  0x1050,
+	'o':  0xdc,
+	'p':  0x170,
+	'q':  0x486,
+	'r':  0x50,
+	's':  0x2088,
+	't':  0x78,
+	'u':  0x1c,
+	'v':  0x2004,
+	'w':  0x2814,
+	'x':  0x28c0,
+	'y':  0x200c,
+	'z':  0x848,
+	'{':  0x949,
+	'|':  0x1200,
+	'}':  0x2489,
+	'~':  0x520,
+}
+
+// Display is a handler to control an alphanumeric display based on ht16k33.
+type Display struct {
+	dev *Dev
+}
+
+// NewAlphaNumericDisplay returns a Display object that communicates over I2C to ht16k33.
+//
+// To use on the default address, ht16k33.I2CAddr must be passed as argument.
+func NewAlphaNumericDisplay(bus i2c.Bus, address uint16) (*Display, error) {
+	dev, err := NewI2C(bus, address)
+	if err != nil {
+		return nil, err
+	}
+
+	display := &Display{dev: dev}
+	return display, nil
+}
+
+// SetDigit at position to provided value.
+func (d *Display) SetDigit(pos int, digit rune, decimal bool) error {
+	val := digitValues[digit]
+	if decimal {
+		val |= digitValues['.']
+	}
+	return d.dev.WriteColumn(pos, val)
+}
+
+// DisplayString print string of values to the display.
+//
+// Characters in the string should be any ASCII value 32 to 127 (printable ASCII).
+func (d *Display) DisplayString(value string, justifyRight bool) error {
+	err := d.dev.Clear()
+	if err != nil {
+		return err
+	}
+	// Calculate starting position of digits based on justification.
+	pos := (4 - len(value))
+	if !justifyRight || pos < 0 {
+		pos = 0
+	}
+	// Go through each character and print it on the display.
+	for _, ch := range value {
+		if ch == '.' {
+			// Print decimal points on the previous digit.
+			c := rune(value[pos-1])
+			if err := d.SetDigit(pos-1, c, true); err != nil {
+				return err
+			}
+		} else {
+			if err := d.SetDigit(pos, ch, false); err != nil {
+				return err
+			}
+			pos++
+		}
+	}
+	return nil
+}
+
+// DisplayInt print a string of numeric values to the display.
+func (d *Display) DisplayInt(value int, justifyRight bool) error {
+	str := strconv.Itoa(value)
+	return d.DisplayString(str, justifyRight)
+}
+
+// DisplayFloat print a string of numeric values to the display.
+func (d *Display) DisplayFloat(value float64, justifyRight bool) error {
+	str := fmt.Sprintf("%5f", value)
+	return d.DisplayString(str, justifyRight)
+}
+
+// Halt clear all the display.
+func (d *Display) Halt() error {
+	return d.dev.Clear()
+}
--- a/experimental/devices/ht16k33/example_test.go
+++ b/experimental/devices/ht16k33/example_test.go
@@ -1,0 +1,51 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+package ht16k33_test
+
+import (
+	"log"
+	"time"
+
+	"periph.io/x/periph/conn/i2c/i2creg"
+	"periph.io/x/periph/experimental/devices/ht16k33"
+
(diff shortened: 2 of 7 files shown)


experimental/devices/ht16k33/alphanum.go
```diff
@@ -0,0 +1,185 @@
...
+// NewAlphaNumericDisplay returns a Display object that communicates over I2C to ht16k33.
+//
+// To use on the default address, ht16k33.I2CAddr must be passed as argument.
+func NewAlphaNumericDisplay(bus i2c.Bus, address uint16) (*Display, error) {
+	dev, err := NewI2C(bus, address)
+	if err != nil {
+		return nil, err
+	}
+
+	display := &Display{dev: dev}
```
maruel: You can inline as:
```
return &Display{dev: dev}, nil
```

experimental/devices/ht16k33/alphanum.go
```diff
@@ -0,0 +1,185 @@
...
+		val |= digitValues['.']
+	}
+	return d.dev.WriteColumn(pos, val)
+}
+
+// DisplayString print string of values to the display.
+//
+// Characters in the string should be any ASCII value 32 to 127 (printable ASCII).
+func (d *Display) DisplayString(value string, justifyRight bool) error {
+	err := d.dev.Clear()
```
maruel: Inline, i.e.:
```
if err := d.dev.Clear(); err != nil {
  return err
}

File: experimental/devices/ht16k33/ht16k33.go [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+	systemSetup   = 0x20
+	oscillatorOff = 0x00
+	oscillatorOn  = 0x01
+	cmdBrightness = 0xE0
+)
+
+// BlinkFrequency display frequency must be a value allowed by the HT16K33.
+type BlinkFrequency byte
+
+// Note that frequency must be a value allowed by the HT16K33, specifically one of:
```
maruel: I'd change the comment for:
```
// Blinking frequencies.
```

experimental/devices/ht16k33/ht16k33.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// NewI2C returns a Dev object that communicates over I2C.
+//
+// To use on the default address, ht16k33.I2CAddr must be passed as argument.
+func NewI2C(bus i2c.Bus, address uint16) (*Dev, error) {
+	dev := &Dev{
+		dev: &i2c.Dev{Bus: bus, Addr: address},
+	}
+	err := dev.init()
```
maruel: inline with next line

experimental/devices/rainbowhat/rainbowhat.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+const (
+	numPixels  = 7
+	bmp280Addr = 0x77
+)
+
+// Dev represents a Rainbow HAT  (https://shop.pimoroni.com/products/rainbow-hat-for-android-things)
+type Dev struct {
+	ledstrip *apa102.Dev
```
maruel: That's really nice composition work!

Personally, I'd make these public; this way you don't need the GetFoo() methods and it's easier to use IMO.

experimental/devices/rainbowhat/rainbowhat.go
```diff
@@ -0,0 +1,184 @@
...
+	ledR     gpio.PinOut
+	ledG     gpio.PinOut
+	ledB     gpio.PinOut
+	buzzer   gpio.PinOut
+	servo    gpio.PinOut
+}
+
+// NewRainbowHat returns a rainbowhat driver.
+func NewRainbowHat() (*Dev, error) {
+	i2cPort, err := i2creg.Open("")
```
maruel: Please still have the client pass the I²C and SPI handles explicitly. While we can assume it's fine, there's a risk of using I²C bus # 0 instead of # 1 when it is manually enabled, which is unusual but may happen.

Or

Explicitly request the exact bus, i.e. i1c 1 and spi 0. You can request `/dev/i2c-1` and `/dev/spidev0.0` as printed with spi-list and i2c-list. That said, if an eventual RPi4 would rename the port numbers, this would break this code.

experimental/devices/rainbowhat/rainbowhat.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	display, err := ht16k33.NewAlphaNumericDisplay(i2cPort, ht16k33.I2CAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	opts := apa102.DefaultOpts
```
maruel: Maybe the user would like to pass options there, so I think it'd be worth having the client to pass this explicitly so they can choose the brightness and temperature. Then make a copy and override the number of pixels, i.e.
```
func NewRainbowHat(ao * apa102.Opts) (*Dev, error) {
  ...
  opts := *ao
  opts.NumPixels = 7
  ...
```

That said .. it's not a big deal either, users can update these values afterward. So thinking about it a bit more, I guess you can ignore this comment.

experimental/devices/rainbowhat/rainbowhat.go
```diff
@@ -0,0 +1,184 @@
...
+	"periph.io/x/periph/devices/bmxx80"
+	"periph.io/x/periph/experimental/devices/ht16k33"
+	"periph.io/x/periph/host/rpi"
+
+	"periph.io/x/periph/conn/gpio"
+	"periph.io/x/periph/conn/i2c/i2creg"
+)
+
+const (
+	numPixels  = 7
```
maruel: Inline these two, no need for a constant since it won't change and they are used exactly once.

experimental/devices/rainbowhat/rainbowhat.go
```diff
@@ -0,0 +1,184 @@
...
+		buttonB:  rpi.P1_38, // GPIO20
+		buttonC:  rpi.P1_36, // GPIO16
+		ledR:     rpi.P1_31, // GPIO06
+		ledG:     rpi.P1_35, // GPIO19
+		ledB:     rpi.P1_37, // GPIO26
+		buzzer:   rpi.P1_33, // PWM1
+		servo:    rpi.P1_32, // PWM0
+	}
+
+	if err := dev.buttonA.In(gpio.PullUp, gpio.FallingEdge); err != nil {
```
maruel: I'd say BothEdges, it's needed to efficiently detect long presses.

experimental/devices/rainbowhat/rainbowhat.go
```diff
@@ -0,0 +1,184 @@
...
+func (d *Dev) GetBuzzer() gpio.PinOut {
+	return d.buzzer
+}
+
+// GetServo returns gpio.PinOut corresponding to the servo pin.
+func (d *Dev) GetServo() gpio.PinOut {
+	return d.servo
+}
+
+// Close halt all devices.
```
maruel: Close != Halt.

I recommend implementing Halt which calls Halt on all devices. Halt turns off all displays and unblock any pending WaitForEdge() calls.

experimental/devices/rainbowhat/rainbowhat.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = d.display.Halt()
+	if err != nil {
+		return err
+	}
+
+	err = d.ledR.Out(gpio.Low)
```
maruel: Call Halt() instead. The goal here is to unblock pending WaitForEdge calls.

experimental/devices/rainbowhat/rainbowhat.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// GetServo returns gpio.PinOut corresponding to the servo pin.
+func (d *Dev) GetServo() gpio.PinOut {
+	return d.servo
+}
+
+// Close halt all devices.
+func (d *Dev) Close() error {
+	err := d.bmp280.Halt()
```
maruel: Inline the lines, so this function is more dense.

experimental/devices/ht16k33/ht16k33.go
```diff
@@ -0,0 +1,119 @@
...
+
+// WriteColumn oie
+func (d *Dev) WriteColumn(column int, data uint16) error {
+	if _, err := d.dev.Write([]byte{byte(column * 2), byte(data & 0xFF), byte(data >> 8)}); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Clear contents of display buffer.
```
maruel: You can rename to Halt.

experimental/devices/ht16k33/ht16k33.go
```diff
@@ -0,0 +1,119 @@
...
+const (
+	BlinkOff    = 0x00
+	Blink2Hz    = 0x02
+	Blink1Hz    = 0x04
+	BlinkHalfHz = 0x06
+)
+
+// Dev is a handler to ht16k33 controller
+type Dev struct {
+	dev *i2c.Dev
```
maruel: It doesn't need to be a pointer, you can use:
```
   dev i2c.Dev
```

experimental/devices/ht16k33/ht16k33.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// NewI2C returns a Dev object that communicates over I2C.
+//
+// To use on the default address, ht16k33.I2CAddr must be passed as argument.
+func NewI2C(bus i2c.Bus, address uint16) (*Dev, error) {
+	dev := &Dev{
```
maruel: you can inline as:
```
  dev := &Dev{dev: i2c.Dev{Bus: bus, Addr: address}}
```

experimental/devices/ht16k33/ht16k33.go
```diff
@@ -0,0 +1,119 @@
...
+	err := dev.init()
+	if err != nil {
+		return nil, err
+	}
+
+	return dev, nil
+}
+
+func (d *Dev) init() error {
+	// Turn on the oscillator.
```
maruel: I'm pretty sure you can pass all these commands as a single write, which would be faster.

experimental/devices/ht16k33/ht16k33.go
```diff
@@ -0,0 +1,119 @@
...
+		return err
+	}
+
+	// Turn on display
+	if _, err := d.dev.Write([]byte{displaySetup | displayOn}); err != nil {
+		return err
+	}
+
+	// Set no blinking.
+	if err := d.SetBlink(BlinkOff); err != nil {
```
maruel: I'd recommend to inline the write here with the commands before, instead of the method call. Same for SetBrightness.

experimental/devices/ht16k33/ht16k33.go
```diff
@@ -0,0 +1,119 @@
...
+	// Set display to full brightness.
+	if err := d.SetBrightness(15); err != nil {
+		return err
+	}
+	return nil
+}
+
+// SetBlink Blink display at specified frequency.
+func (d *Dev) SetBlink(freq BlinkFrequency) error {
+	if _, err := d.dev.Write([]byte{displaySetup | displayOn | byte(freq)}); err != nil {
```
maruel: No need for condition, just call then return err.

experimental/devices/ht16k33/ht16k33.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return nil
+}
+
+// SetBrightness of entire display to specified value.
+//
+// Supports 16 levels, from 0 to 15.
+func (d *Dev) SetBrightness(brightness int) error {
+	if brightness < 0 || brightness > 15 {
+		return errors.New("Brightness must be a value of 0 to 15")
```
maruel: Errors do not start with an upper case. Use something like:
```
  return errors.New("ht16k33: brightness must be between 0 and 15")
```

experimental/devices/ht16k33/ht16k33.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	if _, err := d.dev.Write([]byte{cmdBrightness | byte(brightness)}); err != nil {
```
maruel: Remove condition.

experimental/devices/ht16k33/ht16k33.go
```diff
@@ -0,0 +1,119 @@
...
+		return err
+	}
+	return nil
+}
+
+// WriteColumn oie
```
maruel: Why not take a `data []uint16` and write the whole display at once? It'd be much faster.
See for example https://periph.io/x/periph/devices/tm1637#Dev.Write

experimental/devices/rainbowhat/example_test.go
```diff
@@ -0,0 +1,97 @@
+package rainbowhat_test
```
maruel: Add the copyright here.

experimental/devices/rainbowhat/example_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"log"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"periph.io/x/periph/conn/gpio"
+	"periph.io/x/periph/conn/physic"
+	"periph.io/x/periph/experimental/devices/rainbowhat"
+
```
maruel: Remove empty space.

experimental/devices/rainbowhat/example_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"periph.io/x/periph/host"
+)
+
+func main() {
```
maruel: Did you mean to put this into cmd/ ? Because right now this won't show up as an example.

experimental/devices/rainbowhat/example_test.go
```diff
@@ -0,0 +1,97 @@
...
+		log.Fatalf("failed to draw: %v", err)
+	}
+
+	display := hat.GetDisplay()
+	sensor := hat.GetBmp280()
+	count := 0
+	go func() {
+		for {
+			var envi physic.Env
+			sensor.Sense(&envi)
```
maruel: Check for error and abort.

experimental/devices/rainbowhat/example_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+			temp := fmt.Sprintf("%5s", envi.Temperature)
+			fmt.Printf("Pressure %8s \n", envi.Pressure)
+			fmt.Printf("Temperature %8s \n", envi.Temperature)
+
+			display.DisplayString(temp, true)
+			count++
+			time.Sleep(3 * time.Second)
```
maruel: Use time.NewTicker() instead, otherwise the effective delay is 3s + <sensing delay> + <console write delay> + <display delay> + <Sleep jitter>

experimental/devices/rainbowhat/example_test.go
```diff
@@ -0,0 +1,97 @@
...
+	log.Println("Press ctrl+c to stop...")
+	<-done // Wait
+}
+
+func handleButton(btn gpio.PinIn, led gpio.PinOut) {
+	ledState := false
+	led.Out(gpio.Low)
+	for {
+		btn.WaitForEdge(-1)
+		if ledState {
```
maruel: Do an explicit call to Read() and handle changes. I just realized the example at https://periph.io/device/button/ is incorrect. :(

experimental/devices/rainbowhat/rainbowhat.go
```diff
@@ -0,0 +1,184 @@
...
+
+package rainbowhat
+
+import (
+	"periph.io/x/periph/conn/spi/spireg"
+	"periph.io/x/periph/devices/apa102"
+	"periph.io/x/periph/devices/bmxx80"
+	"periph.io/x/periph/experimental/devices/ht16k33"
+	"periph.io/x/periph/host/rpi"
+
```
maruel: Remove empty line.

experimental/devices/rainbowhat/example_test.go
```diff
@@ -0,0 +1,97 @@
...
+		done <- true
+	}()
+
+	log.Println("Press ctrl+c to stop...")
+	<-done // Wait
+}
+
+func handleButton(btn gpio.PinIn, led gpio.PinOut) {
+	ledState := false
+	led.Out(gpio.Low)
```
maruel: Check for error and return.

experimental/devices/ht16k33/alphanum.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return err
+	}
+	// Calculate starting position of digits based on justification.
+	pos := (4 - len(value))
+	if !justifyRight || pos < 0 {
+		pos = 0
+	}
+	// Go through each character and print it on the display.
```
maruel: It'd be better to rasterize this as a []uint16 first and then do a single method call.

experimental/devices/ht16k33/alphanum.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	display := &Display{dev: dev}
+	return display, nil
+}
+
+// SetDigit at position to provided value.
+func (d *Display) SetDigit(pos int, digit rune, decimal bool) error {
```
maruel: I'm not fond of these methods, I prefer a single `WriteString(s string) (int, error)` that parses accordingly. This reduces the API surface and if a user wants to display a floating point, they know how to do it. In particular, this removes the need to import `fmt`, which has other advantages.

WriteString would do the smart period handling automatically.

tm1637 has a separate package function Digits() https://periph.io/x/periph/devices/tm1637#Digits but that's because this module cannot really display letters. Your case is different on this, hence supporting io.stringWriter has value; https://golang.org/src/io/io.go#L282

maruel:
gohci

maruel:
Would you mind splitting this PR in two?
One for ht16k33 and another one for rainbowhat?

alvarowolfx:
But I'll have to send the PR for the rainbowhat without the ht16k33? I'm updating the PR with some of the recommended changes.

experimental/devices/rainbowhat/rainbowhat.go
maruel: Yes there's a risk but this is fine. This is not java, we do trust the developer.

experimental/devices/ht16k33/example_test.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+package ht16k33_test
```
maruel: You should keep a space between the package name and the copyright notice so it cannot be confused as a package documentation.

experimental/devices/ht16k33/alphanum.go
maruel: This should be done as a follow up.

experimental/devices/ht16k33/alphanum.go
maruel: Somehow I was confused; Display is a very generic struct, to the point it could live elsewhere. But it's fine here for now.

maruel:
gohci

experimental/devices/rainbowhat/rainbowhat.go
alvarowolfx: hahaha good point

PR #299, opened 2018-10-29
mfrc522: make concurrent-safe

vkorn:
The device itself is not supporting concurrent access. Locking and waiting for a previous operation to finish could introduce a variety of possible dead-locks. While this is not the ideal solution it will prevent hardware-lock for the device if concurrent Read/Write was performed. 
All low-level methods are made private, only 4 left public: 

1. ReadCard
1. ReadAuth
1. WriteCard
1. WriteSectorTrail

I'm not sure whether this is a good approach, tbh, since we're hiding low-level methods which might come in handy sometimes. @jdevelop would be great to have your feedback.

### DIFF
--- a/experimental/devices/mfrc522/commands/low_level.go
+++ b/experimental/devices/mfrc522/commands/low_level.go
@@ -24,7 +24,7 @@ const (
 	AuthFailure
 )
 
-// NewSPI creates and initializes the RFID card reader attached to SPI.
+// NewLowLevelSPI creates and initializes the RFID card reader attached to SPI.
 //
 // 	spiPort - the SPI device to use.
 // 	resetPin - reset GPIO pin.
--- a/experimental/devices/mfrc522/commands/low_level.go
+++ b/experimental/devices/mfrc522/commands/low_level.go
@@ -140,7 +140,7 @@ func (r *LowLevel) ClearBitmask(address, mask int) error {
 	return r.DevWrite(address, current&^byte(mask))
 }
 
-// stopCrypto stops the crypto chip.
+// StopCrypto stops the crypto chip.
 func (r *LowLevel) StopCrypto() error {
 	return r.ClearBitmask(Status2Reg, 0x08)
 }
--- a/experimental/devices/mfrc522/commands/low_level.go
+++ b/experimental/devices/mfrc522/commands/low_level.go
@@ -1,0 +1,293 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package commands
+
+import (
+	"fmt"
+	"time"
+
+	"periph.io/x/periph/conn/gpio"
+	"periph.io/x/periph/conn/physic"
+	"periph.io/x/periph/conn/spi"
+)
+
+// AuthStatus indicates the authentication response, could be one of AuthOk,
+// AuthReadFailure or AuthFailure
+type AuthStatus byte
+
+// Card authentication status enum.
+const (
+	AuthOk AuthStatus = iota
+	AuthReadFailure
+	AuthFailure
+)
+
+// NewSPI creates and initializes the RFID card reader attached to SPI.
+//
+// 	spiPort - the SPI device to use.
+// 	resetPin - reset GPIO pin.
+// 	irqPin - irq GPIO pin.
+func NewLowLevelSPI(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn) (*LowLevel, error) {
+	if resetPin == nil {
+		return nil, wrapf("reset pin is not set")
+	}
+	if irqPin == nil {
+		return nil, wrapf("IRQ pin is not set")
+	}
+	spiDev, err := spiPort.Connect(10*physic.MegaHertz, spi.Mode0, 8)
+	if err != nil {
+		return nil, err
+	}
+	if err := resetPin.Out(gpio.High); err != nil {
+		return nil, err
+	}
+	if err := irqPin.In(gpio.PullUp, gpio.FallingEdge); err != nil {
+		return nil, err
+	}
+
+	dev := &LowLevel{
+		spiDev:   spiDev,
+		irqPin:   irqPin,
+		resetPin: resetPin,
+	}
+
+	return dev, nil
+}
+
+// LowLevel is a low-level handler of a MFRC522 RFID reader.
+type LowLevel struct {
+	resetPin gpio.PinOut
+	irqPin   gpio.PinIn
+	spiDev   spi.Conn
+}
+
+// String implements conn.Resource.
+func (r *LowLevel) String() string {
+	return fmt.Sprintf("Mifare MFRC522 [bus: %v, reset pin: %s, irq pin: %s]",
+		r.spiDev, r.resetPin.Name(), r.irqPin.Name())
+}
+
+// DevWrite sends data to a device.
+func (r *LowLevel) DevWrite(address int, data byte) error {
+	newData := []byte{(byte(address) << 1) & 0x7E, data}
+	return r.spiDev.Tx(newData, nil)
+}
+
+// DevRead gets data from a device.
+func (r *LowLevel) DevRead(address int) (byte, error) {
+	data := []byte{((byte(address) << 1) & 0x7E) | 0x80, 0}
+	out := make([]byte, len(data))
+	if err := r.spiDev.Tx(data, out); err != nil {
+		return 0, err
+	}
+	return out[1], nil
+}
+
+// CRC calculates the CRC of the data using the card chip.
+func (r *LowLevel) CRC(inData []byte) ([]byte, error) {
+	if err := r.ClearBitmask(DivIrqReg, 0x04); err != nil {
+		return nil, err
+	}
+	if err := r.SetBitmask(FIFOLevelReg, 0x80); err != nil {
+		return nil, err
+	}
+	for _, v := range inData {
+		if err := r.DevWrite(FIFODataReg, v); err != nil {
+			return nil, err
+		}
+	}
+	if err := r.DevWrite(CommandReg, PCD_CALCCRC); err != nil {
+		return nil, err
+	}
+	for i := byte(0xFF); i > 0; i-- {
+		n, err := r.DevRead(DivIrqReg)
+		if err != nil {
+			return nil, err
+		}
+		if n&0x04 > 0 {
+			break
+		}
+	}
+	lsb, err := r.DevRead(CRCResultRegL)
+	if err != nil {
+		return nil, err
+	}
+
+	msb, err := r.DevRead(CRCResultRegM)
+	if err != nil {
+		return nil, err
+	}
+	return []byte{lsb, msb}, nil
+}
+
+// SetBitmask sets register bit.
+func (r *LowLevel) SetBitmask(address, mask int) error {
+	current, err := r.DevRead(address)
+	if err != nil {
+		return err
+	}
+	return r.DevWrite(address, current|byte(mask))
+}
+
+// ClearBitmask clears register bit.
+func (r *LowLevel) ClearBitmask(address, mask int) error {
+	current, err := r.DevRead(address)
+	if err != nil {
+		return err
+	}
+	return r.DevWrite(address, current&^byte(mask))
+}
+
+// stopCrypto stops the crypto chip.
+func (r *LowLevel) StopCrypto() error {
+	return r.ClearBitmask(Status2Reg, 0x08)
+}
+
+// WaitForEdge waits for an IRQ pin to strobe.
+func (r *LowLevel) WaitForEdge(timeout time.Duration) bool {
+	return r.irqPin.WaitForEdge(timeout)
+}
+
+// Auth authenticate the card fof the sector/block using the provided data.
+//
+// 	mode - the authentication mode.
+// 	sector - the sector to authenticate on.
+// 	block - the block within sector to authenticate.
+// 	sectorKey - the key to be used for accessing the sector data.
+// 	serial - the serial of the card.
+func (r *LowLevel) Auth(mode byte, blockAddress byte, sectorKey [6]byte, serial []byte) (AuthStatus, error) {
+	buffer := make([]byte, 2)
+	buffer[0] = mode
+	buffer[1] = blockAddress
+	buffer = append(buffer, sectorKey[:]...)
+	buffer = append(buffer, serial[:4]...)
+	_, _, err := r.CardWrite(PCD_AUTHENT, buffer)
+	if err != nil {
+		return AuthReadFailure, err
+	}
+	if n, err := r.DevRead(Status2Reg); err != nil || n&0x08 == 0 {
+		return AuthFailure, err
+	}
+	return AuthOk, nil
+}
+
+// CardWrite the low-level interface to write some raw commands to the card.
+//
+// 	command - the command register
+// 	data - the data to write out to the card using the authenticated sector.
+func (r *LowLevel) CardWrite(command byte, data []byte) ([]byte, int, error) {
+	var backData []byte
+	backLength := -1
+	irqEn := byte(0x00)
+	irqWait := byte(0x00)
+
+	switch command {
+	case PCD_AUTHENT:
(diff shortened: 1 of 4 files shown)


experimental/devices/mfrc522/mfrc522.go
```diff
@@ -123,25 +123,30 @@ func NewSPI(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn) (*Dev, er
...
 // Dev is an handle to an MFRC522 RFID reader.
 type Dev struct {
-	resetPin         gpio.PinOut
-	irqPin           gpio.PinIn
+	resetPin gpio.PinOut
+	irqPin   gpio.PinIn
+	spiDev   spi.Conn
+	stop     chan struct{}
+
+	oMu              sync.Mutex
```
maruel: Please use the same mutex for both antennaGain and operationTimeout, there's no need to use separate mutex.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -176,66 +181,187 @@ func (r *Dev) Halt() error {
...
-	if err := r.Reset(); err != nil {
-		return err
+// SetAntennaGain configures antenna signal strength.
+//
+//	gain - signal strength from 0 to 7.
+func (r *Dev) SetAntennaGain(gain int) {
+	r.aMu.Lock()
+	defer r.aMu.Unlock()
+
+	if 0 <= gain && gain <= 7 {
```
maruel: You can skip the mutex if the value is out of range. Preferably an error would be returned instead of a silent failure. So:
```
func (r *Dev) SetAntennaGain(gain int) error {
  if gain < 0 || gain > 7 {
    return errors.New("mfrc522: invalid gain")
  }
  r.aMu.Lock()
  r.antennaGain = gain
  r.aMu.Unlock()
  return nil
}

```

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -176,66 +181,188 @@ func (r *Dev) Halt() error {
...
+		r.antennaGain = gain
 	}
-	if err := r.writeCommandSequence(sequenceCommands.init); err != nil {
-		return err
+}
+
+// ReadCard reads the card sector/block.
+//
+// 	auth - the authentication mode.
```
maruel: I don't think this will come up nice in godoc, you can look locally with:
`godoc -http=localhost:6060`
then browse to http://localhost:6060/pkg/periph.io/x/periph/experimental/devices/mfrc522/

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -176,66 +181,188 @@ func (r *Dev) Halt() error {
...
+	if state != AuthOk {
+		err = wrapf("authentication failed")
+		return
+	}
+
+	return r.write(calcBlockAddress(sector, block%3), data[:])
 }
 
-// SetAntennaGain configures antenna signal strength.
+// WriteSectorTrail writes the sector trait with sector access bits.
```
maruel: trail / trait?

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -176,66 +181,188 @@ func (r *Dev) Halt() error {
...
 //
-//	gain - signal strength from 0 to 7.
-func (r *Dev) SetAntennaGain(gain int) {
-	r.aMu.Lock()
-	defer r.aMu.Unlock()
+// 	auth - authentication mode.
+// 	sector - sector to set authentication.
+// 	keyA - the key used for AuthA authentication scheme.
+// 	keyB - the key used for AuthB authentication schemd.
```
maruel: scheme

maruel:
gohci

vkorn:
Fixed typos in comments and added an example_test as suggested.

jdevelop:
> All low-level methods are made private, only 4 left public:

I would prefer to at least have an opportunity to use low-level API in the specific cases ( like setting up the antenna gain or using the internal CRC calculator ). I already have some of the dependencies on that, hiding the methods would make it less useful for me.

Can we maybe split the API in 2, the simplified one with locks and "general" methods, and the one with more fine-grained access?

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -174,68 +177,192 @@ func (r *Dev) Halt() error {
...
+	copy(data[10:], keyB[:])
+	return r.write(calcBlockAddress(sector&0xFF, 3), data[:])
+}
 
-	if 0 <= gain && gain <= 7 {
-		r.antennaGain = gain
+//		MFRC522 SPI Dev private/helper functions
+
+// Checks whether device is already in use.
+func (r *Dev) accessStarted() error {
```
jdevelop: Is it possible to take a function as an argument to and wrap the execution with lock/unlock?

this way it will be more clean to see the boundaries of the invocations, and less error-prone to forget to unlock the mutex.

Something like 
```
func withDeviceAccess(f func() error) error {
  r.xMu.Lock()
  defer r.Xmu.Unlock()
  return f()
}
```

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -514,156 +684,19 @@ func (r *Dev) SelectTag(serial []byte) (byte, error) {
 	return blocks, nil
 }
 
-// StopCrypto stops the crypto chip.
-func (r *Dev) StopCrypto() error {
+// stopCrypto stops the crypto chip.
+func (r *Dev) stopCrypto() error {
```
jdevelop: I would move this method (and alike) to the `lowlevel` package rather than making them private.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -514,156 +684,19 @@ func (r *Dev) SelectTag(serial []byte) (byte, error) {
...
-}
-
-// Auth authenticate the card fof the sector/block using the provided data.
-//
-// 	mode - the authentication mode.
-// 	sector - the sector to authenticate on.
-// 	block - the block within sector to authenticate.
-// 	sectorKey - the key to be used for accessing the sector data.
-// 	serial - the serial of the card.
-func (r *Dev) Auth(mode byte, sector, block int, sectorKey [6]byte, serial []byte) (AuthStatus, error) {
```
jdevelop: This function may is required to allow testing the authentication for the sector without reading it, would be nice to keep it,

experimental/devices/mfrc522/mfrc522.go
maruel: That's a fair comment and I'm fine with this idea. I used something similar with lepton, moving all commands to a subpackage cci https://periph.io/x/periph/devices/lepton/cci.

Here I feel this would supersede commands .

experimental/devices/mfrc522/mfrc522.go
jdevelop: It's more about the style, I would say. Most of the commands that anyone would need to implement are atomic, e.g - a single thread gets access to the device and performs a sequence of the commands on it. So it is better to have a function that wraps that sequence in "transaction" with locking the library, and pass a function that does something useful.

I could be wrong, but the current implementation you have to call
- accessStarted
- do something
- accessStopped

Here we have 3 blocks and it is fairly easy to panic in the code and have accessStopped lock the device forever.

So with the "pass the function and wrap the context" we are guaranteed to have the library lock managed in the single place, and we don't need an auxiliary `bool` flag to represent the device state.

Hope that it makes sense.

maruel: Yes. That said I'm also fine with:

```
r.lockDevice()
defer r.unlockDevice()
```

which essentially boils down to the same thing (unless mistaken?) and I think readers can relate to this as a lock. Basically rename the function and remove the error.

maruel:
Note: I pushed a gofmt fix so you'll have to rebase.

experimental/devices/mfrc522/mfrc522.go
vkorn: The reason I'm not using explicit lock is that Read could be a blocking operation, for example, if timeout set for indefinite wait second thread will be locked until a card is detected in a first thread.

maruel: Ok, let's leave it as that, please still fix the issues in example_test.go and it's fine to go.

vkorn:
As discussed, moved a few most important low-level operations (CRC, auth, r/w) into commands.LowLevel and exposed it under Dev.LowLevel.
Fixed example_test

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -0,0 +1,293 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package commands
```
maruel: What about have the package name imply that it's low level, then drop the "LowLevel" from the package's symbols? Then you can use New, Dev, etc. I'm not particularly fond of 'commands'.

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -0,0 +1,293 @@
...
+
+// DevWrite sends data to a device.
+func (r *LowLevel) DevWrite(address int, data byte) error {
+	newData := []byte{(byte(address) << 1) & 0x7E, data}
+	return r.spiDev.Tx(newData, nil)
+}
+
+// DevRead gets data from a device.
+func (r *LowLevel) DevRead(address int) (byte, error) {
+	data := []byte{((byte(address) << 1) & 0x7E) | 0x80, 0}
```
maruel: This reminds me I should fix package mmr to work with full duplex connection. But that's out of scope here.

If you use [...] you create an array instead a slice. But I'm not sure it changes much w.r.t. allocating on stack vs on heap.

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -0,0 +1,293 @@
...
+	msb, err := r.DevRead(CRCResultRegM)
+	if err != nil {
+		return nil, err
+	}
+	return []byte{lsb, msb}, nil
+}
+
+// SetBitmask sets register bit.
+func (r *LowLevel) SetBitmask(address, mask int) error {
+	current, err := r.DevRead(address)
```
maruel: Can we cache the bit in some circumstances for performance?

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -0,0 +1,293 @@
...
+		spiDev:   spiDev,
+		irqPin:   irqPin,
+		resetPin: resetPin,
+	}
+
+	return dev, nil
+}
+
+// LowLevel is a low-level handler of a MFRC522 RFID reader.
+type LowLevel struct {
```
maruel: Please implement Halt() and have it call irqPin.Halt(), to unblock WaitForEdge().

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -0,0 +1,293 @@
...
+	lsb, err := r.DevRead(CRCResultRegL)
+	if err != nil {
+		return nil, err
+	}
+
+	msb, err := r.DevRead(CRCResultRegM)
+	if err != nil {
+		return nil, err
+	}
+	return []byte{lsb, msb}, nil
```
maruel: I think returning a uint16 would be more sensible here. You could rename CRC to CRC16.

experimental/devices/mfrc522/example_test.go
```diff
@@ -0,0 +1,104 @@
...
+	rfid, err := mfrc522.NewSPI(p, rpi.P1_13, rpi.P1_11)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Idling device on exit.
+	defer rfid.Halt()
+
+	// Setting the antenna signal strength.
+	rfid.SetAntennaGain(5)
```
maruel: Please check for error everywhere, so that people can copy paste the code safely and this encourages good code.

experimental/devices/mfrc522/example_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+			// If main thread timed out just exiting.
+			if timedOut {
+				return
+			}
+
+			// Some devices tend to send wrong data while RFID chip is already detected
+			// but still "too far" from a receiver.
+			// Especially some cheap CN clones which you can find on GearBest, AliExpress, etc.
+			// This will suppress such errors.
```
maruel: What about having a clear distinction in the error returned by ReadCard() to distinguish between bus error and logical error? Is it possible to distinguish?

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -0,0 +1,293 @@
...
+
+// Auth authenticate the card fof the sector/block using the provided data.
+//
+// 	mode - the authentication mode.
+// 	sector - the sector to authenticate on.
+// 	block - the block within sector to authenticate.
+// 	sectorKey - the key to be used for accessing the sector data.
+// 	serial - the serial of the card.
+func (r *LowLevel) Auth(mode byte, blockAddress byte, sectorKey [6]byte, serial []byte) (AuthStatus, error) {
+	buffer := make([]byte, 2)
```
maruel: Here's a way to not allocate memory on the heap, but that's really a micro optimization so you can ignore. :)
```
var b [32]byte{mode, blockAddress}
buf := append(b[:2], sectorKey[:]...)
buf = append(buf, serial...)
```

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	return AuthOk, nil
+}
+
+// CardWrite the low-level interface to write some raw commands to the card.
+//
+// 	command - the command register
+// 	data - the data to write out to the card using the authenticated sector.
+func (r *LowLevel) CardWrite(command byte, data []byte) ([]byte, int, error) {
+	var backData []byte
```
maruel: You can move these two at line 252.

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -0,0 +1,293 @@
...
+	if command == PCD_TRANSCEIVE {
+		if err := r.SetBitmask(BitFramingReg, 0x80); err != nil {
+			return nil, -1, err
+		}
+	}
+
+	i := 2000
+	n := byte(0)
+
+	for ; i > 0; i-- {
```
maruel: What about a time based check? This code is literally dependent on the I²C bus speed...

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -0,0 +1,293 @@
...
+		} else {
+			backLength = int(n) * 8
+		}
+
+		if n == 0 {
+			n = 1
+		}
+
+		if n > 16 {
+			n = 16
```
maruel: Can you document why cap at 16 bytes?

maruel:
gohci

experimental/devices/mfrc522/commands/low_level.go
vkorn: What about "raw"? And make raw.Dev to implement Resource (with Halt, as you've mentioned, calling irqPin.Halt).

experimental/devices/mfrc522/example_test.go
vkorn: Unfortunately, you can get a whole variety of errors. I have devices from 3 different manufacturers and those two which are from China are sending different signals every time when a card is on the edge of antenna radius :(

experimental/devices/mfrc522/commands/low_level.go
maruel: I'd prefer "mfrc522raw" or "mfrc522low". This is because when a symbol is referred to in client code, mfrc522raw.Dev makes more sense that raw.Dev, which is not informative.

experimental/devices/mfrc522/example_test.go
maruel: I specifically mean that I would expect in this case the error to be one of:
- AuthReadFailure / "can not authenticate"
- In LowLevel.Auth() differentiate when `n&0x08 == 0` triggers vs DevRead() returns an error.
- "expected 16 bytes, actual %d"
- "can't read data after 2000 loops"
- "IRQ error"

vs sysfs SPI driver reporting an error.

PR #301, opened 2018-10-31
as7262: Remove requirement for go1.7

NeuralSpaz:
Refactor to remove context.Context

### DIFF
--- a/experimental/cmd/as7262/main.go
+++ b/experimental/cmd/as7262/main.go
@@ -9,8 +9,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"os/signal"
-	"syscall"
 	"time"
 
 	"periph.io/x/periph/conn/i2c/i2creg"
@@ -37,19 +35,15 @@ func mainImpl() error {
 	}
 	defer bus.Close()
 
-	// Create a new power sensor a sense with default options of 100 mΩ, 3.2A at
-	// address of 0x40 if no other address supplied with command line option.
+	// Create a spectrum sensor.
 	sensor, err := as7262.New(bus, &as7262.Opts{Gain: as7262.G16x})
 	if err != nil {
 		return fmt.Errorf("failed to open new sensor: %v", err)
 	}
 
-	// Read values from sensor every second.
+	// Create a ticker to read values from sensor every second.
 	everySecond := time.NewTicker(time.Second).C
-	var quit = make(chan os.Signal)
 	defer sensor.Halt()
-	signal.Notify(quit, syscall.SIGTERM)
-	signal.Notify(quit, syscall.SIGINT)
 
 	fmt.Println("ctrl+c to quit")
 
@@ -57,10 +51,8 @@ func mainImpl() error {
 
 	for {
 		select {
-		case <-quit:
-			return nil
 		case <-everySecond:
-			spectrum, err := sensor.Sense(100*physic.MilliAmpere, senseTime)
+			spectrum, err := sensor.Sense(12500*physic.MicroAmpere, senseTime)
 			if err != nil {
 				return fmt.Errorf("sensor reading error: %v", err)
 			}
--- a/experimental/cmd/as7262/main.go
+++ b/experimental/cmd/as7262/main.go
@@ -3,9 +3,6 @@
 // that can be found in the LICENSE file.
 
 // as7262 communicates with an as7262 continually reading the spectrum.
-
-// +build go1.7
-
 package main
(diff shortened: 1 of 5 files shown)


experimental/cmd/as7262/main.go
```diff
@@ -40,30 +35,24 @@ func mainImpl() error {
...
 
 	senseTime := time.Millisecond * 300
 
 	for {
 		select {
-		case <-quit:
-			return nil
 		case <-everySecond:
-			spectrum, err := sensor.Sense(100*physic.MilliAmpere, senseTime)
+			spectrum, err := sensor.Sense(12500*physic.MicroAmpere, senseTime)
```
maruel: As a side note, I'd like to have all types in physic implement [flag.Value](https://golang.org/pkg/flag/#Value), which means implementing Set(), so CLI tools could accept values as strings more easily.

experimental/devices/as7262/as7262.go
```diff
@@ -118,8 +125,13 @@ func (d *Dev) Sense(ledDrive physic.ElectricCurrent, senseTime time.Duration) (S
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
-	d.ctx, d.cancel = context.WithCancel(context.Background())
-	defer d.cancel()
+	d.done = make(chan struct{}, 1)
```
maruel: I think this needs to be rewritten as:
```
done := make(chan struct{}, 1)
d.canceledMu.Lock()
d.canceled = false
d.done = done
d.canceledMu.Unlock()
defer func() {
  done <- struct{}{}
}()
```
and then use done instead of d.done below, as d.done is updated only with d.canceledMu lock but not d.mu in Halt().

experimental/devices/as7262/as7262.go
```diff
@@ -238,16 +254,44 @@ const (
...
-	// a client could cast any int as Gain.
+	if gain != G1x && gain != G4x && gain != G16x && gain != G64x {
+		return errGainValue
+	}
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
-	d.ctx, d.cancel = context.WithCancel(context.Background())
-	defer d.cancel()
+	d.canceledMu.Lock()
```
maruel: By the fact that d.mu is held, you know that Sense() is not running. Thus ines 288-294 are not needed.

NeuralSpaz: I think that is still needed in the case where Gain() is called after a Halt().

experimental/devices/as7262/as7262.go
NeuralSpaz: I've made the changes, and I think its right now.

experimental/cmd/as7262/main.go
NeuralSpaz: I can definitely see value in that, it would also make for some more pipe-able tools.

experimental/devices/as7262/as7262.go
```diff
@@ -35,42 +32,48 @@ var DefaultOpts = Opts{
...
-	order  binary.ByteOrder
+
+	// Mutable
+	mu   sync.Mutex
+	gain Gain
+	done chan struct{}
+
+	// Guards canceled
+	canceledMu sync.Mutex
+	canceled   bool
```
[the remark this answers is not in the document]
NeuralSpaz: Canceled is so that calls to Halt() do not send a done if a Halt() has already been called. I think canceled may not be the best name (cancelable is better). I think it is simpler to know that a done had been sent rather than the other way around ie: when a done had been received. In the case when there is something running like Sense() I can see how to respond that done was received, but starts to get a bit complicated if there is nothing to halt. Maybe I just need to think about it more, maybe read some more of context.Context.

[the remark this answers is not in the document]
maruel: What about:
- Remove canceled, rename done to cancel.
- By default, d.cancel is nil
- Sense() does:
```
d.cancel := make(chan struct{})
d.cancelMu.Lock()
d.d.cancel = d.cancel
d.cancelMu.Unlock()
defer func() {
  d.cancelMu.Lock()
  d.d.cancel = nil
  d.cancelMu.Unlock()
}()
... uses d.cancel
```
- Halt does:
```
d.cancelMu.Lock()
if d.cancel != nil {
  select {
    case d.cancel <- struct{}{}:
    default:
  }
  d.cancel = nil
}
d.cancelMu.Unlock()
```

maruel:
gohci

PR #302, opened 2018-10-31
bh1750: Add support for ambient light sensor

alvarowolfx:
Add support to the BH1750 ambient light sensor.

Datasheet: https://www.waveshare.com/wiki/2.13inch_e-Paper_HAT

### DIFF
--- a/experimental/devices/bh1750/bh1750.go
+++ b/experimental/devices/bh1750/bh1750.go
@@ -1,0 +1,140 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package bh1750
+
+import (
+	"encoding/binary"
+	"time"
+
+	"periph.io/x/periph/conn/i2c"
+	"periph.io/x/periph/conn/physic"
+)
+
+// I2CAddr i2c default address.
+const I2CAddr uint16 = 0x23
+
+// AlternativeI2CAddr i2c alternative address.
+const AlternativeI2CAddr uint16 = 0x5c
+
+// Resolution represents the measurements modes.
+type Resolution uint8
+
+const (
+	// ContinuousHighResMode start measurement at 1lx resolution. Measurement time is approx 120ms.
+	ContinuousHighResMode Resolution = 0x10
+
+	// ContinuousHighResMode2 start measurement at 0.5lx resolution. Measurement time is approx 120ms.
+	ContinuousHighResMode2 Resolution = 0x11
+
+	// ContinuousLowResMode start measurement at 4lx resolution. Measurement time is approx 16ms.
+	ContinuousLowResMode Resolution = 0x13
+
+	// OneTimeHighResMode measurement at 1lx resolution and the device is automatically set to Power Down after measurement.
+	OneTimeHighResMode Resolution = 0x20
+
+	// OneTimeHighResMode2 measurement at 0.5lx resolution and the device is automatically set to Power Down after measurement.
+	OneTimeHighResMode2 Resolution = 0x21
+
+	// OneTimeLowResMode measurement at 4lx resolution and the device is automatically set to Power Down after measurement.
+	OneTimeLowResMode Resolution = 0x23
+
+	safetyTimeout = 20 * time.Millisecond
+)
+
+// Map of timeouts for measurement type.
+var timeout = map[Resolution]time.Duration{
+	ContinuousHighResMode:  120*time.Millisecond + safetyTimeout,
+	ContinuousHighResMode2: 120*time.Millisecond + safetyTimeout,
+	ContinuousLowResMode:   16*time.Millisecond + safetyTimeout,
+	OneTimeHighResMode:     120*time.Millisecond + safetyTimeout,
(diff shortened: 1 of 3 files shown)


experimental/devices/bh1750/bh1750.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if err := d.SetResolution(ContinuousHighResMode); err != nil {
+		return err
+	}
+
+	_, err := d.dev.Write([]byte{reset})
+	return err
+}
+
+// SetMode set the sleep mode of the sensor.
```
maruel: I think this is not needed: it starts when you Sense(), it stops when you call Halt(). This removes the need for Mode, which further simplifies the API.

experimental/devices/bh1750/bh1750.go
```diff
@@ -0,0 +1,140 @@
...
+// SetResolution set the resolution mode of the sensor.
+func (d *Dev) SetResolution(r Resolution) error {
+	d.res = r
+	_, err := d.dev.Write([]byte{byte(r)})
+	return err
+}
+
+// Sense reads the light value from the bh1750 sensor.
+func (d *Dev) Sense() (physic.LuminousFlux, error) {
+	if err := d.SetResolution(d.res); err != nil {
```
maruel: Are you sure this is needed?

experimental/devices/bh1750/bh1750.go
```diff
@@ -0,0 +1,140 @@
...
+	"periph.io/x/periph/conn/physic"
+)
+
+// I2CAddr i2c default address.
+const I2CAddr uint16 = 0x23
+
+// AlternativeI2CAddr i2c alternative address.
+const AlternativeI2CAddr uint16 = 0x5c
+
+// Resolution represents the measurements modes.
```
maruel: In [bmxx80](https://periph.io/x/periph/devices/bmxx80), I was able to hide this by having the client either use:
- Use [SenseContinous()](https://periph.io/x/periph/conn/physic#SenseEnv) for continuous sensing.
- Use Sense() for one off operation.

Then you can expose the resolutions 0.5lx, 1lx, and 4lx and then have the code underneath use the right mode based on the function used.

https://github.com/google/periph/blob/master/devices/bmxx80/bmxx80.go#L431 shows how it's implemented.

experimental/devices/bh1750/bh1750.go
```diff
@@ -0,0 +1,140 @@
...
+	PowerOn Mode = 0x01
+)
+
+const (
+	dataRegister            uint8 = 0x00
+	measurementTimeRegister uint8 = 0x45
+	reset                   uint8 = 0x07
+)
+
+// Dev is a handler to bh1750 controller
```
maruel: Ends doc with a period.

experimental/devices/bh1750/bh1750.go
```diff
@@ -0,0 +1,140 @@
...
+func NewI2C(bus i2c.Bus, address uint16) (*Dev, error) {
+	dev := &Dev{dev: i2c.Dev{Bus: bus, Addr: address}}
+
+	if err := dev.init(); err != nil {
+		return nil, err
+	}
+	return dev, nil
+}
+
+func (d *Dev) init() error {
```
maruel: optional: I think you can inline init() since it's fairly simple.

experimental/devices/bh1750/bh1750.go
```diff
@@ -0,0 +1,140 @@
...
+	return err
+}
+
+// Sense reads the light value from the bh1750 sensor.
+func (d *Dev) Sense() (physic.LuminousFlux, error) {
+	if err := d.SetResolution(d.res); err != nil {
+		return 0, err
+	}
+
+	time.Sleep(timeout[d.res])
```
maruel: optional: cap1xxx uses a [nice way to skip sleeping](https://github.com/google/periph/blob/master/devices/cap1xxx/cap1xxx.go#L547) between a command and the read.

That said, I think you want to change SetPrecision() to *not* write to the device, it's not really needed if you implement SenseContinous().

experimental/devices/bh1750/bh1750.go
```diff
@@ -0,0 +1,140 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package bh1750
```
maruel: Please add package documentation.

datasheet:
http://rohmfs.rohm.com/en/products/databook/datasheet/ic/sensor/light/bh1721fvc-e.pdf

experimental/devices/bh1750/bh1750.go
```diff
@@ -0,0 +1,140 @@
...
+
+	time.Sleep(timeout[d.res])
+
+	buf := [2]byte{}
+	if err := d.dev.Tx([]byte{byte(d.res)}, buf[:]); err != nil {
+		return 0, err
+	}
+
+	rawValue := binary.BigEndian.Uint16(buf[:])
+	lux := physic.LuminousFlux(float64(rawValue) / 1.2)
```
maruel: I think you can do this calculation without floating point. The doc (page 15) is unclear but I think this would look like:

    physic.LuminousFlux(rawValue) * 833333333 * physic.NanoLumen

maruel:
gohci

PR #304, opened 2018-11-02
Added support for ADS1x15

bezineb5:
Hello,

This is the first attempt at a driver for the Texas Instrument ADS1015/ADS1115 Analog-Digital Converters. They are widely available independently (https://www.adafruit.com/product/1085) or within hats (https://www.adafruit.com/product/3018).

I tried to mimic the pin system as offered by the host so that the ADS1x15 device allows to retrieve "Analog Pins" which in turn let the user read the electric potential at the input. I could not make them implement analog.Pin, as the "Read" method returns int32, whereas I prefer to return the actual physical value as physic.ElectricPotential. Do you have any suggestion to improve that?
Also, I tried to limit the exposure of the user to the internal working of the ADC by hiding the specific knowledge about the gain and data rate behind physically meaningful values, respectively the maximum voltage to be read and the polling frequency.
In order to achieve good performance, some precomputed data are cached into the AnalogPins. 

Do you think the direction conforms to your vision of periph.io?

Thanks!
Benjamin


experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+
+func (p *ads1x15AnalogPin) Number() int {
+	return -1
+}
+
+func (p *ads1x15AnalogPin) Function() string {
+	return "DEPRECATED"
+}
+
+func (p *ads1x15AnalogPin) Halt() (err error) {
```
maruel: Please do not used named return variable unless necessary. It is necessary only when you want to access the variable from a deferred function.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+		option(l)
+	}
+
+	return
+}
+
+// Halt returns true if devices is halted successfully
+func (d *Dev) Halt() (err error) { return }
+
+// WithI2CAddress option sets the ADS1x15Driver I2C address, if it is not the default.
```
maruel: Please use an explicit Opts struct like in the other drivers, for consistency, instead of a callback pattern.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
+package ads1x15
```
maruel: Please add copyright.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	"encoding/binary"
+	"errors"
+	"math"
+	"sync"
+	"time"
+
+	"periph.io/x/periph/conn/physic"
+	"periph.io/x/periph/conn/pin"
+
+	"fmt"
```
maruel: Please use `goimports -w` instead of managing imports yourself, this means grouping the stdlib imports first (including fmt), then all periph.io imports in a second group.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+
+	"periph.io/x/periph/conn/i2c"
+)
+
+const (
+
+	// ADS1x15DefaultAddress is the default I2C address for the ADS1x15 components
+	ADS1x15DefaultAddress = 0x48
```
maruel: It's fine to use I2CAddr, so that at usage it looks like ads1x15.I2CAddr. Right now it stutters.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+
+	ads1x15PointerConversion    = 0x00
```
maruel: Don't mix private and exported symbols in a single const group, move the private symbols down.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	ads1x15ConfigModeContinuous = 0x0000
+	//Single shoot mode
+	ads1x15ConfigModeSingle = 0x0100
+
+	ads1x15ConfigCompWindow      = 0x0010
+	ads1x15ConfigCompAactiveHigh = 0x0008
+	ads1x15ConfigCompLatching    = 0x0004
+	ads1x15ConfigCompQueDisable  = 0x0003
+
+	Channel0 = 0
```
maruel: IMO these constants are not useful, remove.

*Unless*

You do:
```
// Channel is the DAC sampling to return.
//
// It is either the raw value of one pin, or the delta between two pins.
type Channel int

const (
  // Reads an individual DAC channel.
  Channel0 = 4
  Channel1 = 5
  Channel2 = 6
  Channel0 = 7
  // Reads the difference between two channels.
  Channel0Minus1 = 0
  Channel1Minus3 = 1
  Channel1Minus3 = 2
  Channel1Minus3 = 3
)
```

And then you can remove the function PinForDifferenceOfChannels() and inline prepareQuery().

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	if currentBestDataRate < 0 {
+		err = fmt.Errorf("The maximum frequency which can be read is %s", max.String())
+		return
+	}
+
+	bestDataRate = currentBestDataRate
+	return
+}
+
+func (d *Dev) checkChannel(channel int) (err error) {
```
maruel: Please inline. You can make the error a global variable if you don't want to copy-paste the sting.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+func (p *ads1x15AnalogPin) Read() (physic.ElectricPotential, error) {
+	return p.adc.executePreparedQuery(p.preparedQuery, p.waitTime, p.voltageMultiplier)
+}
+
+func (p *ads1x15AnalogPin) Name() string {
+	return fmt.Sprintf("%s pin", p.adc.name)
+}
+
+func (p *ads1x15AnalogPin) Number() int {
+	return -1
```
maruel: Return 0~3 for the non-differential ones.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+}
+
+func (p *ads1x15AnalogPin) Function() string {
+	return "DEPRECATED"
```
maruel: It's going to be something from https://periph.io/x/periph/experimental/conn/analog but I haven't started this yet. In the meantime you can use "DAC"

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	name string
+
+	gainConfig      map[int]uint16
+	dataRates       map[int]uint16
+	gainVoltage     map[int]physic.ElectricPotential
+	DefaultDataRate int
+	mutex           *sync.Mutex
+}
+
+// AnalogPin represents a pin which is able to read an electric potential
```
maruel: Historical note: https://periph.io/x/periph/experimental/conn/analog#DAC predates https://periph.io/x/periph/conn/physic. This interface assumes we'll never get to support hardware that can extract more than 32 bits of raw precision (which would be a feat in practice).

Also I wasn't sure if I should always use ElectricPotential for the returned value, as in some case the calibration may be non-trivial.

For now this interface is a good idea. Feel free to refactor the package analog in a later PR.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+type AnalogPin interface {
+	pin.Pin
+	Read() (physic.ElectricPotential, error)
+}
+
+type ads1x15AnalogPin struct {
```
maruel: I'd like the 4 pins to be stored inside Dev as:

```
type Dev struct {
   ...
   pins [4] pin
}
```

No need to specify the whole " ads1x15Analog" prefix IMHO.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	adc               *Dev
+	preparedQuery     []byte
+	voltageMultiplier physic.ElectricPotential
```
maruel: Isn't that "resolution" ?
You may want to include dcOffset too.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	waitTime          time.Duration
+}
+
+// NewADS1015 creates a new driver for the ADS1015 (12-bit ADC)
```
maruel: I prefer a single function, which an Opts struct that specify which device it is in practice.
That said there's trade off on each side and I'm starting to think maybe it's easier for users to have multiple explicit functions.
So leave it as that for now, I'll think about it a bit more.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+// Largely inspired by: https://github.com/adafruit/Adafruit_Python_ADS1x15
+func NewADS1015(i i2c.Bus, options ...func(*Dev)) (l *Dev, err error) {
+	l, err = newADS1x15(i, options...)
+
+	l.dataRates = map[int]uint16{
```
maruel: Make these two maps (this one and the one at line 99) global variables.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	l.name = "ADS1115"
+
+	return
+}
+
+func newADS1x15(i i2c.Bus, options ...func(*Dev)) (l *Dev, err error) {
+	l = &Dev{
+		c: i2c.Dev{Bus: i, Addr: ADS1x15DefaultAddress},
+		// Mapping of gain values to config register values.
+		gainConfig: map[int]uint16{
```
maruel: These (gainConfig, gainVoltage) do not need to be members. So move these as global.
Also this is not floating point, so 2 / 3 == 0. You'll have to adjust your logic accordingly.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	// I2C Communication
+	c i2c.Dev
+
+	name string
+
+	gainConfig      map[int]uint16
+	dataRates       map[int]uint16
+	gainVoltage     map[int]physic.ElectricPotential
+	DefaultDataRate int
+	mutex           *sync.Mutex
```
maruel: Do not make it a pointer, just a member.
Please rename it "mu".

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+type Dev struct {
+	// I2C Communication
+	c i2c.Dev
+
+	name string
+
+	gainConfig      map[int]uint16
+	dataRates       map[int]uint16
+	gainVoltage     map[int]physic.ElectricPotential
+	DefaultDataRate int
```
maruel: It's not "Default" here, it's the DataRate.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	if l.DefaultDataRate == 0 {
+		l.DefaultDataRate = 1600
+	}
+
+	l.name = "ADS1015"
+
+	return
+}
+
+// NewADS1115 creates a new driver for the ADS1115 (16-bit ADC)
```
maruel: Please consistently terminate all sentences, especially the ones in documentation, with a period.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+func (d *Dev) Halt() (err error) { return }
+
+// WithI2CAddress option sets the ADS1x15Driver I2C address, if it is not the default.
+func WithI2CAddress(i2cAddress uint16) func(*Dev) {
+	return func(d *Dev) {
+		d.c.Addr = i2cAddress
+	}
+}
+
+func (d *Dev) PinForChannel(channel int, maxVoltage physic.ElectricPotential, minimumFrequency physic.Frequency) (pin AnalogPin, err error) {
```
maruel: Please document.
I'd rename minimumFrequency to f. :)

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	return
+}
+
+// bestDataRateForFrequency returns the gain the most data rate to read samples at least at the requested frequency.
+func (d *Dev) bestDataRateForFrequency(minimumFrequency physic.Frequency) (bestDataRate int, err error) {
+	var max physic.Frequency
+	difference := physic.Frequency(math.MaxInt64)
+	currentBestDataRate := -1
+
+	for key := range d.dataRates {
```
maruel: Issues:
- You are not going through the map in order, this is problematic
- So I recommend to convert the map into an ordered slice of struct.

```
type dataRate struct {
  f physic.Frequency
  v uint16
}
 
var ads1015DataRates = [...]dataRate{
  {128*physic.Hertz, 0},
  {250*physic.Hertz, 0x20},
  ...
}
```

Then you can do:
```
for i := d.dataRates {
  if f <= d.dataRates[i].f {
    return d.dataRates[i].v
  }
}
return d.dataRates[len(d.dataRates)-1].v
```
This is much simpler. You could return an error if the datarate is too high.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	return
+}
+
+// bestGainForElectricPotential returns the gain the most adapted to read up to the specified difference of potential.
+func (d *Dev) bestGainForElectricPotential(voltage physic.ElectricPotential) (bestGain int, err error) {
+	var max physic.ElectricPotential
+	difference := physic.ElectricPotential(math.MaxInt64)
+	currentBestGain := -1
+
+	for key, value := range d.gainVoltage {
```
maruel: Similar to bestDataRateForFrequency() comment below, use an array (and not a slice) instead of a map.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	config |= ads1x15ConfigCompQueDisable // Disable comparator mode.
+
+	// Build the query to the ADC
+	configBytes := make([]byte, 2)
+	binary.BigEndian.PutUint16(configBytes, config)
+	preparedQuery := append([]byte{ads1x15PointerConfig}, configBytes...)
+
+	// The wait for the ADC sample to finish is based on the sample rate plus a
+	// small offset to be sure (0.1 millisecond).
+	waitTime := time.Duration(1000000/dataRate+100) * time.Microsecond
```
maruel: Please rewrite as:
```
waitTime := (time.Second + dataRate/2) / time.Duration(dataRate) + 100* time.Microsecond
```

So you use constants instead of numbers, especially that everything is scaled in Microsecond, which makes it harder to read. My version includes better integer rounding.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+		},
+		gainVoltage: map[int]physic.ElectricPotential{
+			2 / 3: 6144 * physic.MilliVolt,
+			1:     4096 * physic.MilliVolt,
+			2:     2048 * physic.MilliVolt,
+			4:     1024 * physic.MilliVolt,
+			8:     512 * physic.MilliVolt,
+			16:    256 * physic.MilliVolt,
+		},
+		mutex: &sync.Mutex{},
```
maruel: You can now remove this line.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+
+// AnalogPin represents a pin which is able to read an electric potential
+type AnalogPin interface {
+	pin.Pin
+	Read() (physic.ElectricPotential, error)
+}
+
+type ads1x15AnalogPin struct {
+	adc               *Dev
+	preparedQuery     []byte
```
maruel: You use both prepareQuery as a member and as a method in the same file, please use a different name for either the member or the method, to help "glanceability".

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+		preparedQuery:     preparedQuery,
+		voltageMultiplier: voltageMultiplier,
+		waitTime:          waitTime,
+	}
+
+	return
+}
+
+func (d *Dev) executePreparedQuery(query []byte, waitTime time.Duration, voltageMultiplier physic.ElectricPotential) (value physic.ElectricPotential, err error) {
+	// Lock the ADC converter to avoid multiple simultaneous readings.
```
maruel: Actually, are you sure it's a good idea? I mean maybe someone wants to read from multiple channels from different goroutines?

As a follow up: having a way to read all values at once may be useful for the user.
As a follow up: https://periph.io/x/periph/conn/physic#SenseEnv SenseContinuous is a nice way to have a stable continuous sampling, where the driver can optimize the configured sampling rate based on the frequency the user requested. This has to be stoppable via Halt().

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -0,0 +1,392 @@
...
+	time.Sleep(waitTime)
+
+	// Retrieve the result.
+	data := []byte{0, 0}
+	if err = d.c.Tx([]byte{ads1x15PointerConversion}, data); err != nil {
+		return
+	}
+
+	// Convert the raw data into physical value.
+	result := physic.ElectricPotential(binary.BigEndian.Uint16(data))
```
maruel: I'd prefer something like:
```
raw := int16(binary.BigEndian.Uint16(data))
return physic.ElectricPotential(raw) * mult / physic.ElectricPotential(1<<15)
```

Especially that I see value to eventually just return the raw reading.
Or better (?) maybe the package analog should define an analog reading as:

```
// Reading is the result of DAC.Read()  (obviously not the case right now but this could be)
type Reading struct {
  V   physic.ElectricPotential
  Raw int32
}
```

then users are free to use the relevant value.

maruel:
gohci

experimental/devices/ads1x15/ads1x15.go
bezineb5: Unfortunately, those ADS1x15 devices are made of 1 ADC bundled with a multiplexer. So it's actually not possible to measure multiple channels simultaneously ([diagram](http://www.ti.com/general/docs/datasheetdiagram.tsp?genericPartNumber=ADS1115&diagramId=SBAS444D)).
If you want to do that, you need to have 2 physical ADC components (it's a good thing it's easy to change the I2C address to have them chained).

Follow-up #1: that's interesting, but we have to keep in mind that each pin can have different rate/gain settings. But we could imagine having a simplified version for that case, with shared settings. Do you have a particular use-case in mind?

Follow-up #2: actually, the ADS1x15 have a continuous mode where the ADC never sleeps and samples continuously at one of the fixed rates. However, when using that mode, it's not possible to use multiple pins, as the multiplexer is in a fixed state. So doing that in software could be a better idea, as it would be more flexible and would avoid having the user deal too much with the internals of the device. I'll try to do it, as it totally makes sense from a user perspective.

experimental/devices/ads1x15/ads1x15.go
bezineb5: Excellent idea, I've implemented it.

experimental/devices/ads1x15/ads1x15.go
bezineb5: Done

experimental/devices/ads1x15/ads1x15.go
bezineb5: Oups, sorry

experimental/devices/ads1x15/ads1x15.go
bezineb5: Done

experimental/devices/ads1x15/ads1x15.go
bezineb5: Done

experimental/devices/ads1x15/ads1x15.go
bezineb5: Done

experimental/devices/ads1x15/ads1x15.go
bezineb5: Done

experimental/devices/ads1x15/ads1x15.go
bezineb5: I have to keep it, as the device can only measure 1 input at a time

bezineb5:
I decided not to update the analog package, as it would be a breaking change. However, I tried to adapt the "AnalogPin" interface to what would be an updated version of the analog package.

maruel:
gohci

PR #305, opened 2018-11-03
hx711: add experimental driver.

davidsansome:
HX711 is a 24-bit ADC commonly used for weigh scales.

### DIFF
--- a/experimental/cmd/hx711/main.go
+++ b/experimental/cmd/hx711/main.go
@@ -31,16 +31,11 @@ func mainFunc() error {
 		return err
 	}
 
-	const pinPattern = "no %s pin specified. Please provide the pin via '%s' flag, for example '%s'"
-
 	if *clkPin == "" {
-		return fmt.Errorf(pinPattern, "clock", "-clk", "-clk 25")
+		return fmt.Errorf("-clk is required")
 	}
 	if *dataPin == "" {
-		return fmt.Errorf(pinPattern, "data", "-data", "-data 26")
-	}
-	if *gain != 128 && *gain != 64 && *gain != 32 {
-		return fmt.Errorf("invalid gain '%d', must be either 128, 64 or 32", *gain)
+		return fmt.Errorf("-data is required")
 	}
 	if *cont && *samples != 0 {
 		return fmt.Errorf("-cont and -samples can't be used together")
@@ -48,24 +43,27 @@ func mainFunc() error {
 
 	clkPinReg := gpioreg.ByName(*clkPin)
 	if clkPinReg == nil {
-		return fmt.Errorf("Clock pin %s can not be found", *clkPin)
+		return fmt.Errorf("clock pin %s can not be found", *clkPin)
 	}
 	dataPinReg := gpioreg.ByName(*dataPin)
 	if dataPin == nil {
-		return fmt.Errorf("Data pin %s can not be found", *dataPin)
+		return fmt.Errorf("data pin %s can not be found", *dataPin)
 	}
 
 	dev, err := hx711.New(clkPinReg, dataPinReg)
 	if err != nil {
 		return err
 	}
+
 	switch *gain {
 	case 128:
-		dev.InputMode = hx711.CHANNEL_A_GAIN_128
+		dev.SetInputMode(hx711.CHANNEL_A_GAIN_128)
 	case 64:
-		dev.InputMode = hx711.CHANNEL_A_GAIN_64
+		dev.SetInputMode(hx711.CHANNEL_A_GAIN_64)
 	case 32:
-		dev.InputMode = hx711.CHANNEL_B_GAIN_32
+		dev.SetInputMode(hx711.CHANNEL_B_GAIN_32)
+	default:
+		return fmt.Errorf("invalid gain '%d', must be either 128, 64 or 32", *gain)
 	}
 
 	if *cont {
--- a/experimental/cmd/hx711/main.go
+++ b/experimental/cmd/hx711/main.go
@@ -1,0 +1,97 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"time"
+
+	"periph.io/x/periph/conn/gpio/gpioreg"
+	"periph.io/x/periph/experimental/devices/hx711"
+	"periph.io/x/periph/host"
+)
+
+const timeout = time.Second
+
+func mainFunc() error {
+	clkPin := flag.String("clk", "", "Clock pin")
+	dataPin := flag.String("data", "", "Data pin")
+	gain := flag.Int("gain", 128,
+		"Voltage gain. Must be one of 128, 64 or 32. Using 32 selects Channel B")
+	cont := flag.Bool("cont", false, "Reads continuously from the ADC")
+	samples := flag.Int("samples", 0,
+		"Reads several samples from the ADC and outputs the average value")
+	flag.Parse()
+
+	if _, err := host.Init(); err != nil {
+		return err
+	}
+
+	const pinPattern = "no %s pin specified. Please provide the pin via '%s' flag, for example '%s'"
+
+	if *clkPin == "" {
+		return fmt.Errorf(pinPattern, "clock", "-clk", "-clk 25")
+	}
+	if *dataPin == "" {
+		return fmt.Errorf(pinPattern, "data", "-data", "-data 26")
+	}
+	if *gain != 128 && *gain != 64 && *gain != 32 {
+		return fmt.Errorf("invalid gain '%d', must be either 128, 64 or 32")
+	}
+	if *cont && *samples != 0 {
+		return fmt.Errorf("-cont and -samples can't be used together")
+	}
+
+	clkPinReg := gpioreg.ByName(*clkPin)
+	if clkPinReg == nil {
+		return fmt.Errorf("Clock pin %s can not be found", *clkPin)
+	}
+	dataPinReg := gpioreg.ByName(*dataPin)
+	if dataPin == nil {
+		return fmt.Errorf("Data pin %s can not be found", *dataPin)
+	}
+
+	dev, err := hx711.New(clkPinReg, dataPinReg)
+	if err != nil {
+		return err
+	}
+	switch *gain {
+	case 128:
+		dev.InputMode = hx711.CHANNEL_A_GAIN_128
+	case 64:
+		dev.InputMode = hx711.CHANNEL_A_GAIN_64
+	case 32:
+		dev.InputMode = hx711.CHANNEL_B_GAIN_32
+	}
+
+	if *cont {
+		ch := dev.StartContinuousRead()
+		for {
+			fmt.Println(<-ch)
+		}
+	} else if *samples != 0 {
+		value, err := dev.ReadAveraged(timeout, *samples)
+		if err != nil {
+			return err
+		}
+		fmt.Println(value)
+	} else {
+		value, err := dev.Read(timeout)
+		if err != nil {
+			return err
+		}
+		fmt.Println(value)
+	}
+	return nil
+}
+
+func main() {
+	if err := mainFunc(); err != nil {
+		fmt.Fprintf(os.Stderr, "hx711: %s.\n", err)
+		os.Exit(1)
+	}
+}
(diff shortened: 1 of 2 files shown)


experimental/cmd/hx711/main.go
```diff
@@ -0,0 +1,97 @@
...
+		return err
+	}
+	switch *gain {
+	case 128:
+		dev.InputMode = hx711.CHANNEL_A_GAIN_128
+	case 64:
+		dev.InputMode = hx711.CHANNEL_A_GAIN_64
+	case 32:
+		dev.InputMode = hx711.CHANNEL_B_GAIN_32
+	}
```
maruel: you can use default: return error here instead of lines 42-44, I'd prefer that.

experimental/cmd/hx711/main.go
```diff
@@ -0,0 +1,97 @@
...
+	cont := flag.Bool("cont", false, "Reads continuously from the ADC")
+	samples := flag.Int("samples", 0,
+		"Reads several samples from the ADC and outputs the average value")
+	flag.Parse()
+
+	if _, err := host.Init(); err != nil {
+		return err
+	}
+
+	const pinPattern = "no %s pin specified. Please provide the pin via '%s' flag, for example '%s'"
```
maruel: I'd just say "-clk is required" / "-data is required"

experimental/cmd/hx711/main.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	dataPinReg := gpioreg.ByName(*dataPin)
+	if dataPin == nil {
+		return fmt.Errorf("Data pin %s can not be found", *dataPin)
+	}
+
+	dev, err := hx711.New(clkPinReg, dataPinReg)
+	if err != nil {
+		return err
+	}
```
maruel: defer dev.Halt()

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,180 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package hx711 implements an interface to the HX711 analog to digital converter.
+//
+// Datasheet: https://www.mouser.com/ds/2/813/hx711_english-1022875.pdf
```
maruel: I like using this form so one can do #hdr-Datasheet to refer to it

```
//
// Datasheet
//
// https://www.mouser.com/ds/2/813/hx711_english-1022875.pdf
```

Example: https://periph.io/x/periph/devices/bmxx80#hdr-Datasheets

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,180 @@
...
+const (
+	CHANNEL_A_GAIN_128 InputMode = 1
+	CHANNEL_A_GAIN_64  InputMode = 3
+	CHANNEL_B_GAIN_32  InputMode = 2
+
+	dataBits         = 24
+	readPollInterval = 50 * time.Millisecond
+)
+
+type HX711 struct {
```
maruel: I like using Dev for all packages, hx711.HX711 stutters.

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,180 @@
...
+	useEdge bool
+	done    chan struct{}
+}
+
+// New creates a new HX711 device.
+func New(clk gpio.PinOut, data gpio.PinIn) (*HX711, error) {
+	// Try enabling edge detection on the data pin.
+	var useEdge bool
+	if err := data.In(gpio.PullDown, gpio.FallingEdge); err != nil {
+		if err := data.In(gpio.PullDown, gpio.NoEdge); err != nil {
```
maruel: I'm in the process of adding gpioutil.PollEdge() for software based polling for edge. It's currently at an incorrect location, the end location would be conn/gpio/gpioutil.
https://periph.io/x/periph/experimental/devices/gpioutil#PollEdge

Can you use it and confirm that the code works properly? I didn't promote to stable yet because I need folks to test it and find bugs. :) It's actually more tricky than it looks like; https://github.com/google/periph/blob/master/experimental/devices/gpioutil/polledge.go

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,180 @@
...
+				return 0, TimeoutError
+			}
+			time.Sleep(readPollInterval)
+		}
+	}
+
+	// Shift the 24-bit 2's compliment value.
+	var value uint32
+	for i := 0; i < dataBits; i++ {
+		d.clk.Out(gpio.High)
```
maruel: Please check for errors.

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,180 @@
...
+
+	d.done = done
+	return ret
+}
+
+// StopContinuousRead stops a continuous read that was started with
+// StartContinuousRead.  This will close the channel that was returned by
+// StartContinuousRead.
+func (d *HX711) StopContinuousRead() {
+	close(d.done)
```
maruel: Calling this function twice will panic. I'd be nice to recycle d.done.

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,180 @@
...
+	// TimeoutError is returned from Read and ReadAveraged when the ADC took too
+	// long to indicate data was available.
+	TimeoutError = errors.New("timed out waiting for HX711 to become ready")
+)
+
+// InputMode controls the voltage gain and the channel multiplexer on the HX711.
+// Channel A can be used with a gain of 128 or 64, and Channel B can be used
+// with a gain of 32.
+//
+// Changing the InputMode on an HX711 will only take effect *after* the next
```
maruel: Then maybe worth making it a method `SetGain()` so that the user doesn't have to care about this, as the function could do a wasted Read() implicitly. I feel it'd be simpler for users.

maruel:
gohci

maruel:
Warning: I moved gpioutil to experimental/conn/gpio/gpioutil in a7c4fbd25ecbb8adc43ff77af492ac558df94ade because being under devices/ didn't feel right.

experimental/devices/hx711/hx711.go
davidsansome: Is it enough to just check for errors from the initial call to .Out?  I've added a check to New.

experimental/cmd/hx711/main.go
davidsansome: It doesn't have a Halt :)
Maybe I should add one though?  The datasheet mentions you can set the CLK pin high for 60us to power down the chip.

experimental/devices/hx711/hx711.go
davidsansome: Cool!  I'll give this a try tonight!

davidsansome: Done!  It seems to work!
There's a race condition though.  I want to wait for the pin to be low, regardless of the current level.  But if I do:

    if pin.Read() != gpio.Low {
      pin.WaitForEdge(timeout)
    }

and it goes low between my call to pin.Read() and WaitForEdge's curr := p.PinIO.Read(), then it'll miss it and wait for the timeout.
I think I really wait a WaitForLevel()!

experimental/cmd/hx711/main.go
maruel: Please make sure it implements [conn.Resource](https://periph.io/x/periph/conn#Resource). It's fine to implement proper shutdown later, or not if it's not seen as useful.

experimental/devices/hx711/hx711.go
maruel: That's a good question. For device drivers I prefer to check unconditionally. In practice the only case where this can happen in practice is when using a FTDI device over USB, e.g. https://periph.io/x/extra/hostextra/d2xx

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,173 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package hx711 implements an interface to the HX711 analog to digital converter.
+//
+// Datasheet
+//
+// https://www.mouser.com/ds/2/813/hx711_english-1022875.pdf
```
maruel: The original source is http://www.aviaic.com/Download/hx711F_EN.pdf.pdf (yes with two .pdf) so please use this even if _iffy_.

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,173 @@
...
+// Channel A can be used with a gain of 128 or 64, and Channel B can be used
+// with a gain of 32.
+type InputMode int
+
+const (
+	CHANNEL_A_GAIN_128 InputMode = 1
+	CHANNEL_A_GAIN_64  InputMode = 3
+	CHANNEL_B_GAIN_32  InputMode = 2
+
+	dataBits         = 24
```
maruel: nit: Personally I'd just inline this variable, it is used exactly once.

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,173 @@
...
+	readPollInterval = 50 * time.Millisecond
```
maruel: This variable is not used, so it should be removed.

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,173 @@
...
+	d.inputMode = inputMode
+	d.readImmediately()
+}
+
+// IsReady returns true if there is data ready to be read from the ADC.
+func (d *Dev) IsReady() bool {
+	return d.data.Read() == gpio.Low
+}
+
+// Read reads a single value from the ADC.  It blocks until the ADC indicates
```
maruel: FYI, @bezineb5 added two days ago a very interesting interface in https://periph.io/x/periph/experimental/devices/ads1x15#AnalogPin. We discussed that it should live in https://periph.io/x/periph/experimental/conn/analog once we're happy with the design.

Since the interface is still inside a device driver you cannot use it here yet but I'd like your input on this and how we could settle for something good. @bezineb5 also proposed #307 for continuous reading.

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (d *Dev) readImmediately() int32 {
+	// Shift the 24-bit 2's compliment value.
+	var value uint32
+	for i := 0; i < dataBits; i++ {
+		d.clk.Out(gpio.High)
+		level := d.data.Read()
+		d.clk.Out(gpio.Low)
+
+		if level {
```
maruel:
```
value <<= 1
if level {
  value |= 1
}
```

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,173 @@
...
+
+	return d.readImmediately(), nil
+}
+
+func (d *Dev) readImmediately() int32 {
+	// Shift the 24-bit 2's compliment value.
+	var value uint32
+	for i := 0; i < dataBits; i++ {
+		d.clk.Out(gpio.High)
+		level := d.data.Read()
```
maruel: In some case, the delay between out and read can be *very* short. What's the minimum delay the device needs?

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,173 @@
...
+
+		if level {
+			value = (value << 1) | 1
+		} else {
+			value = (value << 1)
+		}
+	}
+
+	// Convert the 24-bit 2's compliment value to a 32-bit signed value.
+	var signedValue int32 = int32(value<<8) >> 8
```
maruel: you can inline at line 116.

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,173 @@
...
+	}()
+
+	d.done = done
+	return ret
+}
+
+// StopContinuousRead stops a continuous read that was started with
+// StartContinuousRead.  This will close the channel that was returned by
+// StartContinuousRead.
+func (d *Dev) StopContinuousRead() {
```
maruel: This should be Halt(). That's the design used by gpio, but I just realized this is not documented properly. Oops :( Filed #311 for this.

experimental/devices/hx711/hx711.go
```diff
@@ -0,0 +1,173 @@
...
+	if d.done != nil {
+		close(d.done)
+		d.done = nil
+	}
+}
+
+// ReadAveraged reads several samples from the ADC and returns the mean value.
```
maruel: This feels like a pure compute function that doesn't need to be a method, and (ultimately) could be extracted to an hypothetical analog/analogutil package. For the moment I'd vote to unexport it unless you need it right now, we can migrate later.

experimental/devices/hx711/hx711.go
maruel: I understand that there's no room for error here, it can be left as-is for now and we can revisit this later.

maruel:
gohci

maruel:
Merged right away since nothing what a showstopper. Please address comments in a follow up as it makes sense to youé

PR #307, opened 2018-11-03
ads1x15: add continuous reading, add quality bias. Fix min/max values.

bezineb5:
I fixed some remarks, and added 2 new features:
* Continuous reading
* A flag to indicate if the priority should be on power consumption or conversion quality. That will bias the conversion rate in order to turn the ADC off as long as possible.

### DIFF
--- a/experimental/devices/ads1x15/ads1x15.go
+++ b/experimental/devices/ads1x15/ads1x15.go
@@ -44,6 +44,16 @@ const (
 	Channel3 = 3
 )
 
+// ConversionQuality represents a request for a compromise between energy saving versus conversion quality.
+type ConversionQuality int
+
+const (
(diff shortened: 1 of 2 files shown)


experimental/devices/ads1x15/ads1x15.go
```diff
@@ -384,6 +423,35 @@ func (p *ads1x15AnalogPin) Read() (Reading, error) {
 	return p.adc.executePreparedQuery(p.query, p.waitTime, p.voltageMultiplier)
 }
 
+func (p *ads1x15AnalogPin) ReadContinuous() <-chan Reading {
+	p.stopContinous()
+	reading := make(chan Reading)
+	p.stop = make(chan struct{})
```
maruel: you cannot do this without a form of locking, i.e. via a mutex.

maruel:
gohci

PR #309, opened 2018-11-04
Add decimal type.

NeuralSpaz:
This adds the decimal type which is exact fixed point representation of a decimal number.

Add atod: ascii to decimal.
Add dtoi: decimal to int.
Add parseSIPrefix.


NeuralSpaz:
Thinking about it maybe it would be better to have dtoi have the signature
```golang
func dtoi(d decimal, scale) (int, error)
```
rather than being a method?

NeuralSpaz:
changed dtoi signature, I like it more, as it is closer to std lib form.

conn/physic/units.go
```diff
@@ -733,3 +734,247 @@ func picoAsString(v int64) string {
...
+
+// Converts from decimal to int64, using the decimal.digits and converting to a
+// unit64. Scale is combined with the decimal exponent to maximise the
+// resolution and is in powers of ten.
+func dtoi(d decimal, scale int) (int64, error) {
+	// Maximum value for a int64.
+	const max = (1<<63 - 1)
+	// Use uint till the last as it allows checks for overflows.
+	var u uint64
+	for _, c := range []byte(d.digits) {
```
maruel: Use a for i < len() for otherwise this code is causing a memory allocation.

conn/physic/units.go
```diff
@@ -733,3 +734,247 @@ func picoAsString(v int64) string {
...
+		}
+	}
+
+	if end == 0 {
+		end = len(s)
+	}
+	last := end
+	seenDigit = false
+	exp := 0
+	// Strip non significant zeros.
```
maruel: .. to find the base exponent.

conn/physic/units.go
```diff
@@ -733,3 +734,247 @@ func picoAsString(v int64) string {
...
+				}
+			}
+		default:
+			last--
+			end--
+		}
+	}
+
+	if dp > start && dp < end {
+		//Concatenate with out decimal point
```
maruel: Space after //, period at the end. (I'm more strict with non experimental code)

NeuralSpaz:
glad you like it. :)

conn/physic/units.go
```diff
@@ -733,3 +734,250 @@ func picoAsString(v int64) string {
...
+	// digits hold the string representation of the significant decimal digits.
+	digits string
+	// exponent is the left or right decimal shift. (powers of ten).
+	exp int
+	// neg it true if the number is negative.
+	neg bool
+}
+
+// Positive powers of 10 in the form such that powerOF10[index] = 10^index.
+var powerOf10 = [19]uint64{
```
maruel: Still think "..." is better here than hardcoding the length.

conn/physic/units.go
```diff
@@ -733,3 +734,250 @@ func picoAsString(v int64) string {
...
+	mag := d.exp + scale
+	if mag > 18 {
+		return 0, errors.New("exceeds maximum exponent")
+	}
+	if mag < 0 {
+		mag *= -1
+	}
+	// Divide is = 10^(-mag)
+	if d.exp+scale < 0 {
+		u /= powerOf10[mag]
```
maruel: Improve the rounding with:
```
u = (u + powerOf10[mag]/2) / powerOf10[mag]
```

conn/physic/units.go
```diff
@@ -733,3 +734,250 @@ func picoAsString(v int64) string {
...
+		if d.digits[i] >= '0' && d.digits[i] <= '9' {
+			// '0' = 0x30 '1' = 0x31 ...etc.
+			digit := d.digits[i] - '0'
+			// *10 is decimal shift left.
+			u *= 10
+			check := u + uint64(digit)
+			// Check should always be larger than u unless we have overflowed.
+			// Similarly if check > max it will overflow when converted to int64.
+			if check < u || check > max {
+				return int64(u), &parseError{err: errors.New("overflows maximum is"), s: strconv.FormatUint(max, 10)}
```
maruel: I'd prefer if the FormatUint(max) was a constant. Maybe a global. So basically an option is:

```
const max = (1<<63-1)
var maxStr = "9223372036854775807" 
```

then in test do
```
func TestingMax(t* testing.T) {
  if strconv.FormatUint(max, 10) != maxStr {
    t.Fatal("unexpected text representation of max")
  }
}
```

I know this is optimizing breadcrumbs but I really try to reduce the number of runtime calls.

conn/physic/units.go
```diff
@@ -733,3 +734,250 @@ func picoAsString(v int64) string {
...
+// Converts a string to a decimal form. The return int is how many bytes of the
+// string are numeric. The string may contain +-0 prefixes and arbitrary
+// suffixes as trailing non number characters are ignored.
+// Significant digits are stored without leading or trailing zeros, rather an
+// exponet is used.
+func atod(s string) (decimal, int, error) {
+	var d decimal
+	start := 0
+	dp := 0
+	end := 0
```
```
end := len(s)
```
removes the need for lines 877 to 879.

conn/physic/units.go
```diff
@@ -733,3 +734,250 @@ func picoAsString(v int64) string {
...
+	kilo  prefix = 3
+	mega  prefix = 6
+	giga  prefix = 9
+	tera  prefix = 12
+)
+
+func parseSIPrefix(r rune) (prefix, int) {
+	switch r {
+	case 'p':
+		return pico, len(string(r))
```
maruel: Because you are using r here, I think the compile will not be able to inline this into a constant. So while less safe, I'd prefer len("p"), etc, for performance.

conn/physic/units_test.go
```diff
@@ -536,3 +537,166 @@ func BenchmarkCelsiusFloatg(b *testing.B) {
...
 	}
 }
+
+func TestAtod(t *testing.T) {
+	const (
+		negative = true
+		postive  = false
+	)
+	tests := []struct {
+		s    string
```
maruel: I'd prefer:
```
in       string
expected decimal
used     int
```

Then make one table for cases that succeeds, and one for a table that fails, preferably in two different test case. It'll be more readable.

conn/physic/units_test.go
```diff
@@ -536,3 +537,166 @@ func BenchmarkCelsiusFloatg(b *testing.B) {
 		buf.Reset()
 	}
 }
+
+func TestAtod(t *testing.T) {
+	const (
+		negative = true
+		postive  = false
```
maruel: Use "positive".

conn/physic/units_test.go
```diff
@@ -536,3 +537,166 @@ func BenchmarkCelsiusFloatg(b *testing.B) {
...
+
+		if n != tt.used {
+			t.Errorf("expected to consume %d char but used %d", tt.used, n)
+		}
+	}
+}
+
+func TestDoti(t *testing.T) {
+	tests := []struct {
+		name string
```
maruel: Similar,
```
name     string
in       decimal
expected int64
```

maruel:
gohci

conn/physic/units_test.go
```diff
@@ -536,3 +537,166 @@ func BenchmarkCelsiusFloatg(b *testing.B) {
...
+	for i := 0; i < b.N; i++ {
+		if d, n, err = atod("337.2m"); err != nil {
+			b.Fatal(err)
+		}
+		if v, err = dtoi(d, 0); err != nil {
+			b.Fatal(err)
+		}
+	}
+	b.StopTimer()
+	fmt.Sprintf("%d %d", v, n)
```
maruel: Replace with
```
_, _ = fmt.Sprintf("%d %d", v, n)
```
to fix
https://gist.github.com/gohci-bot/acbffc2a9dcccd452303976f2ce63a2c#file-cmd3-failed-in-8-747s

conn/physic/units.go
NeuralSpaz: no I like it, cleaner too. I made it maxInt64 and maxInt64str for a bit of extra clarity since it is global.

conn/physic/units_test.go
NeuralSpaz: While rewriting the test noticed I was not returning the correct sign if it was minimum value. So now its more like:
```golang
if check < u || check > maxInt64 {
	if d.neg {
		return (-maxInt64), &parseError{err: errors.New("overflows minimum is"), s: "-" + maxUint64Str}
	}
	return maxInt64, &parseError{err: errors.New("overflows maximum is"), s: maxUint64Str}
}
```

conn/physic/units_test.go
NeuralSpaz: also was not checking for large -ve exponent.
so moved mag check bellow.
now:
```golang
mag := d.exp + scale
if mag < 0 {
	mag *= -1
}
if mag > 18 {
	return 0, errors.New("exponent exceeds int64")
}
```

conn/physic/units.go
NeuralSpaz: Mhm, that is better.

conn/physic/units.go
NeuralSpaz: That is better idea.

conn/physic/units_test.go
```diff
@@ -536,3 +538,255 @@ func BenchmarkCelsiusFloatg(b *testing.B) {
...
+	}{
+		{"123456789", decimal{"123456789", 0, positive}, 9},
+		{"1nM", decimal{"1", 0, positive}, 1},
+		{"2.2nM", decimal{"22", -1, positive}, 3},
+		{"12.5mA", decimal{"125", -1, positive}, 4},
+		{"-12.5mA", decimal{"125", -1, negative}, 5},
+		{"1ma1", decimal{"1", 0, positive}, 1},
+		{"-0.00001%rH", decimal{"1", -5, negative}, 8},
+		{"0.00001%rH", decimal{"1", -5, positive}, 7},
+		{"--1ma1", decimal{"1", 0, negative}, 3},
```
maruel: Thinking about it, I don't think these two should be valid.
https://play.golang.org/p/IgZrjBjeoRa

conn/physic/units_test.go
```diff
@@ -536,3 +538,255 @@ func BenchmarkCelsiusFloatg(b *testing.B) {
...
+	for _, tt := range tests {
+		got := tt.err.Error()
+
+		if got != tt.want {
+			t.Errorf("wanted err string:\n%s but got:\n%s", tt.want, got)
+		}
+
+	}
+}
+func TestMaxInt64(t *testing.T) {
```
maruel: include an empty line between functions

conn/physic/units_test.go
```diff
@@ -536,3 +538,255 @@ func BenchmarkCelsiusFloatg(b *testing.B) {
...
+		{"empty", &parseError{s: "", err: errors.New("test")}, "parse error: test: \"" + "\""},
+		{"noUnits", noUnits("someunit"), "parse error: no units provided, need: \"someunit\""},
+	}
+	for _, tt := range tests {
+		got := tt.err.Error()
+
+		if got != tt.want {
+			t.Errorf("wanted err string:\n%s but got:\n%s", tt.want, got)
+		}
+
```
maruel: Style nit: I'm not sure why you occasionally add an empty line before a closing bracket.

conn/physic/units_test.go
```diff
@@ -536,3 +538,255 @@ func BenchmarkCelsiusFloatg(b *testing.B) {
...
+		if d, n, err = atod("-337.2m"); err != nil {
+			b.Fatal(err)
+		}
+	}
+	b.StopTimer()
+	_ = fmt.Sprintf("%v %d", d, n)
+}
+
+func BenchmarkString2Decimal2IntNeg(b *testing.B) {
+
```
maruel: Similar here, I'm not used to an empty line after an opening bracket. In practice I mostly favor consistency, so have it consistent across the functions.

maruel:
.. style nits and the ++ and -- ones should be fixed. I think it's preferable to deny both.

conn/physic/units.go
```diff
@@ -733,3 +734,255 @@ func picoAsString(v int64) string {
...
+		n *= -1
+	}
+	return n, nil
+}
+
+// Converts a string to a decimal form. The return int is how many bytes of the
+// string are numeric. The string may contain +-0 prefixes and arbitrary
+// suffixes as trailing non number characters are ignored.
+// Significant digits are stored without leading or trailing zeros, rather an
+// exponet is used.
```
maruel: exponent

NeuralSpaz: Oh the joys of dyslexia; the gift that keeps on giving.

conn/physic/units_test.go
NeuralSpaz: agreed, also "-+" and "+-".

conn/physic/units_test.go
NeuralSpaz: fwiw: It's simply does not catch my eye as an error, I had to read bottom to top to see any of them.
I'll keep a closer eye on my formatting style, as I can imagine it is a pain to nit them.

conn/physic/units.go
```diff
@@ -848,18 +860,47 @@ func atod(s string) (decimal, int, error) {
 	seenDigit := false
 	seenZero := false
 	isPoint := false
+	seenPlus := false
 
 	// Strip leading zeros, +/- and mark DP.
 	for i := 0; i < len(s); i++ {
 		switch {
 		case s[i] == '-':
+			if seenPlus {
```
maruel: I just realized; I think this string would pass: "1-2" because you don't deny if either seenDigit or seenZero is true. Same for case '+'. Please a test case to confirm.

NeuralSpaz: Yes. It is a bug. nicely spotted!
The question is should return a decimal 1 or error?
I'm thinking error. I'm having a hard time thinking of a case where it would not be.

maruel: I'm ambivalent but somewhat leaning towards not an error;
- If it is an error, then it's a one-off compared to the letters, "10-" is invalid, but "10n" is valid.
- if it is parsed properly, parseSIUnits() will then error out properly which I think is better.

conn/physic/units_test.go
```diff
@@ -536,3 +538,257 @@ func BenchmarkCelsiusFloatg(b *testing.B) {
...
+		{"1ma1", decimal{"1", 0, positive}, 1},
+		{"+1ma1", decimal{"1", 0, positive}, 2},
+		{"-0.00001%rH", decimal{"1", -5, negative}, 8},
+		{"0.00001%rH", decimal{"1", -5, positive}, 7},
+		{"1.0", decimal{"1", 0, positive}, 3},
+		{"0.10001", decimal{"10001", -5, positive}, 7},
+		{"-0.10001", decimal{"10001", -5, negative}, 8},
+		{"1n", decimal{"1", 0, positive}, 1},
+		{"200n", decimal{"2", 2, positive}, 3},
+		{".01", decimal{"1", -2, positive}, 3},
```
maruel: For completeness, can you add "+.01" and "-.01". It's really more as a documentation than anything else.

NeuralSpaz: Np. Found another bug introduced by last fix.

conn/physic/units_test.go
```diff
@@ -536,3 +538,269 @@ func BenchmarkCelsiusFloatg(b *testing.B) {
...
+		{"010", decimal{"1", 1, positive}, 3},
+		{"001", decimal{"1", 0, positive}, 3},
+	}
+
+	fails := []struct {
+		in       string
+		expected decimal
+		n        int
+	}{
+		{"1.1.1", decimal{}, 0},
```
maruel: I'm not ambivalent that it should fail; it could succeed and return "1.1" and ".1" as the trailing string.
But at this point let's commit since this code is already superb.

conn/physic/units.go
```diff
@@ -733,3 +734,307 @@ func picoAsString(v int64) string {
...
+
+// Maximum value for a int64.
+const maxInt64 = (1<<63 - 1)
+
+var maxUint64Str = "9223372036854775807"
+
+// Converts from decimal to int64, using the decimal.digits character values and
+// converting to a intermediate unit64.
+// Scale is combined with the decimal exponent to maximise the resolution and is
+// in powers of ten.
```
maruel: We should eventually add a comment that "it tries to go as far as it can until it finds a character that cannot be parsed as a number".

maruel:
gohci

maruel:
Seriously an amazing change, thanks a lot! I don't think I could have written something as good.

NeuralSpaz:
:) Np, thank you for all your help.

PR #312, opened 2018-11-06
hx711: tidy up a few remaining PR comments

davidsansome:
- Remove ReadAveraged - it's not a good fit here.
- Use the proper datasheet URL.
- Remove some constants.
- Rename StopContinuousRead to Halt.


maruel:
gohci

maruel:
Thanks a lot for following through!

PR #315, opened 2018-11-07
ads1x15: style improvement

maruel:
No functional change.

- Move private globals at the bottom of the file.
- Move ads1x15AnalogPin close to its methods.
- Stop using named return variable.
- Inline newADS1x15.
- Move gainConfig and gainVoltage as global variables.
- Wraps comments.
- Rename mutex to mu.
- Use -math.MaxInt16 instead of math.MinInt16 because of the voltage multiplier.
- Change error strings to start with a lower case.

### DIFF
--- a/experimental/devices/ads1x15/ads1x15.go
+++ b/experimental/devices/ads1x15/ads1x15.go
@@ -18,25 +18,8 @@ import (
 )
 
 const (
-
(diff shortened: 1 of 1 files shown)


maruel:
@bezineb5 for your information.

Nothing huge but if you had local changes this will interfere.

PR #316, opened 2018-11-07
physic: Add flag.Value implementation for Frequency 

NeuralSpaz:
Enable physic.Frequency to use the flag.Value interface.

### DIFF
--- a/conn/physic/example_test.go
+++ b/conn/physic/example_test.go
@@ -85,13 +85,19 @@ func ExampleFrequency() {
 func ExampleFrequency_Set() {
 	var f physic.Frequency
 
-	f.Set("10MHz")
+	if err := f.Set("10MHz"); err != nil {
+		fmt.Println(err)
+	}
 	fmt.Println(f)
 
-	f.Set("10mHz")
+	if err := f.Set("10mHz"); err != nil {
+		fmt.Println(err)
+	}
 	fmt.Println(f)
 
-	f.Set("1kHz")
+	if err := f.Set("1kHz"); err != nil {
+		fmt.Println(err)
+	}
 	fmt.Println(f)
 	// Output:
 	// 10MHz
--- a/conn/physic/example_test.go
+++ b/conn/physic/example_test.go
@@ -5,6 +5,7 @@
 package physic_test
 
 import (
+	"flag"
 	"fmt"
 	"time"
 
@@ -79,6 +80,30 @@ func ExampleFrequency() {
 	// 10mHz
 	// 101.010Hz
 	// 10MHz
+}
+
+func ExampleFrequency_Set() {
+	var f physic.Frequency
+
+	f.Set("10MHz")
+	fmt.Println(f)
+
+	f.Set("10mHz")
+	fmt.Println(f)
+
+	f.Set("1kHz")
+	fmt.Println(f)
+	// Output:
+	// 10MHz
+	// 10mHz
+	// 1kHz
+}
+
+func ExampleFrequency_flag() {
+	var pwm physic.Frequency
+
+	flag.Var(&pwm, "pwm", "pwm frequency")
+	flag.Parse()
 }
 
 func ExampleFrequency_Duration() {
--- a/conn/physic/example_test.go
+++ b/conn/physic/example_test.go
@@ -7,6 +7,7 @@ package physic_test
 import (
 	"flag"
 	"fmt"
+	"log"
 	"time"
 
 	"periph.io/x/periph/conn/physic"
@@ -86,17 +87,17 @@ func ExampleFrequency_Set() {
 	var f physic.Frequency
 
 	if err := f.Set("10MHz"); err != nil {
-		fmt.Println(err)
+		log.Fatal(err)
 	}
 	fmt.Println(f)
 
 	if err := f.Set("10mHz"); err != nil {
-		fmt.Println(err)
+		log.Fatal(err)
 	}
 	fmt.Println(f)
 
 	if err := f.Set("1kHz"); err != nil {
-		fmt.Println(err)
+		log.Fatal(err)
 	}
 	fmt.Println(f)
 	// Output:
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -209,6 +209,11 @@ func (f Frequency) String() string {
 	return microAsString(int64(f)) + "Hz"
 }
 
+const (
+	maxFrequency = 9223372036854775807 * MicroHertz
+	minFrequency = -9223372036854775807 * MicroHertz
+)
+
 // Set sets the Frequency to the value represented by s. Units are to be
 // provided in hertz with optional SI prefixes.
 func (f *Frequency) Set(s string) error {
@@ -217,17 +222,15 @@ func (f *Frequency) Set(s string) error {
 		if e, ok := err.(*parseError); ok {
 			switch e.err {
 			case errOverflowsInt64:
-				found, _ := containsUnitString(s[n:], "Hertz", "Hz")
-				return &parseError{
-					msg: "maximum value for \"" + found + "\" is " + e.msg + "n" + found,
-					err: errors.New(""),
-				}
+				return errors.New("maximum value is " + maxFrequency.String())
 			case errUnderflowsInt64:
-				found, _ := containsUnitString(s[n:], "Hertz", "Hz")
-				return &parseError{
-					msg: "minimum value for \"" + found + "\" is " + e.msg + "n" + found,
-					err: errors.New(""),
+				return errors.New("minimum value is " + minFrequency.String())
+			case errNotANumber:
+				found, _ := containsUnitString(s, "Hertz", "Hz")
+				if found != "" {
+					return errors.New("does not contain number")
 				}
+				return errors.New("does not contain number or unit \"Hz\"")
 			}
 		}
 		return err
@@ -817,6 +820,7 @@ var maxUint64Str = "9223372036854775807"
 var (
 	errOverflowsInt64  = errors.New("exceeds maximum")
 	errUnderflowsInt64 = errors.New("exceeds minimum")
+	errNotANumber      = errors.New("not a number")
 )
 
 // Converts from decimal to int64, using the decimal.digits character values and
@@ -851,7 +855,7 @@ func dtoi(d decimal, scale int) (int64, error) {
 			u = check
 		} else {
 			// Should not get here if used atod to generate the decimal.
-			return 0, &parseError{err: errors.New("not a number")}
+			return 0, &parseError{err: errNotANumber}
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


conn/physic/example_test.go
```diff
@@ -81,6 +82,30 @@ func ExampleFrequency() {
 	// 10MHz
 }
 
+func ExampleFrequency_Set() {
+	var f physic.Frequency
+
+	f.Set("10MHz")
```
maruel: Please check for error.

conn/physic/units.go
```diff
@@ -981,6 +999,29 @@ func atod(s string) (decimal, int, error) {
...
+// Set is a helper for converting a string and a prefix in to a physic unit. It
+// can be used when characters of the units do not conflict with any of the SI
+// prefixes.
+func set(s string, base prefix) (int64, int, error) {
+	d, n, err := atod(s)
+	if err != nil {
+		return 0, n, err
+	}
+	si := prefix(none)
+	if !(n == len(s)) {
```
maruel: `if n != len(s)` :)

conn/physic/units_test.go
```diff
@@ -745,6 +745,133 @@ func TestMaxInt64(t *testing.T) {
...
+		{"1THz", 1 * TeraHertz},
+		{"12.345Hz", 12345 * MilliHertz},
+		{"-12.345Hz", -12345 * MilliHertz},
+	}
+
+	fails := []struct {
+		in string
+	}{
+		{"10THz"},
+		{"10"},
```
maruel: That's a bummer for one thing though; when I look at https://github.com/google/periph/blob/master/cmd/apa102/main.go#L122, it's should accept the value without unit.
In the future I intent to replace `-hz` with `-freq` but that's a breaking change.

conn/physic/units_test.go
```diff
@@ -745,6 +745,133 @@ func TestMaxInt64(t *testing.T) {
...
+		{"1T", micro, 1000000000000000000, 2},
+	}
+
+	fails := []struct {
+		in     string
+		prefix prefix
+	}{
+		{"1G", pico},
+		{"1T", pico},
+		{"1T", nano},
```
maruel: please also do one for micro. In practice I'd like it to have the exact "1 too much" value, and have the maximum value in the succeeds above. Having the limits explicitly in the unit tests is informative for the readers.

conn/physic/units.go
```diff
@@ -981,6 +999,29 @@ func atod(s string) (decimal, int, error) {
 	return d, last, nil
 }
 
+// Set is a helper for converting a string and a prefix in to a physic unit. It
```
maruel: s/Set/set/

conn/physic/units.go
```diff
@@ -981,6 +999,29 @@ func atod(s string) (decimal, int, error) {
...
+// can be used when characters of the units do not conflict with any of the SI
+// prefixes.
+func set(s string, base prefix) (int64, int, error) {
+	d, n, err := atod(s)
+	if err != nil {
+		return 0, n, err
+	}
+	si := prefix(none)
```
maruel: I just realized, `none` should be renamed `unit`.

conn/physic/units.go
```diff
@@ -981,6 +999,29 @@ func atod(s string) (decimal, int, error) {
...
+	if !(n == len(s)) {
+		var n1 int
+		si, n1 = parseSIPrefix([]rune(s[n:])[0])
```
maruel: Please use utf8.DecodeRuneInString(), instead of converting the whole string suffix into []rune.

conn/physic/units_test.go
NeuralSpaz: I mean it could.... it just the line 223 
```golang
 return noUnits("Hz")
```
could be 
```golang
*f = (Frequency)(v)
```

NeuralSpaz: just need some check around it to make sure its only a number nothing else.

conn/physic/units.go
```diff
@@ -207,6 +209,23 @@ func (f Frequency) String() string {
...
+func (f *Frequency) Set(s string) error {
+	v, n, err := valueOfUnitString(s, micro)
+	if err != nil {
+		return err
+	}
+
+	switch strings.ToLower(s[n:]) {
+	case "hz", "hertz":
+		*f = (Frequency)(v)
+	default:
```
NeuralSpaz: for naked units we could do something like:
```golang
case "":
    *f = (Frequency)(v)
default:
    incorrectUnits(s[n:], "Hz" )
```

NeuralSpaz: I did this.

My thoughts are if the units are unambiguous, ie: no other units then its fine to allow naked units?

So frequency I think is fine, someone may want rpm but they can do the logic for that themselves.

conn/physic/units.go
```diff
@@ -1001,14 +1050,18 @@ func noUnits(s string) error {
 	return &parseError{s: s, err: errors.New("no units provided, need")}
 }
 
+func incorrectUnit(inputString, want string) error {
+	return &parseError{err: errors.New("\"" + inputString + "\"" + " is not a valid unit for " + want)}
```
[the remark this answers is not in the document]
NeuralSpaz: I'm not particularly happy with the errors are either.
there is a third mode. 
-  The input was not a number at all. 

On the expanded error form I'm not sure how far we could go as far as correct usage of a flag.
I just put together a flagset to see the formatting. This is the output:
`
invalid value "10" for flag -freq: parse error: no units provided, need: "Hz"
`
I think it should look more like:
- The unit was missing:
`invalid value "10" for flag -freq: does not contain any units. Valid unit are Hertz or Hz`
- The unit was wrong:
`invalid value "10rpm" for flag -freq: incorrect usage. Valid unit are Hertz or Hz`
- The SI prefix was wrong:
`invalid value "10fHz" for flag -freq: could not correctly determine SI prefix. Valid prefixes are n, p, u, µ, m, k, M, G or T`
- Was not a number:
`invalid value "ten" for flag -freq: does not appear to be a number`

How does that look?

conn/physic/units.go
NeuralSpaz: ok lets go with that then.

conn/physic/units.go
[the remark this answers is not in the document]
maruel: I like the error messages.
My challenge: how do you distinguish between "unit was wrong" and "SI prefix was wrong" ?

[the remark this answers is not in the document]
NeuralSpaz: If there is a number a valid unit but with an extra Alpha runes that is not a prefix... 
My only thought atm on that is if gets to the default case and the prefix size is unit, there could be a check if any substrings are valid units. At one stage I was using and an acceptUnit(...string) bool that did that. Since it is only for a failing case, the extra cycles spent on the string comparisons would not matter? I'll code something up and see how it looks.

also some other errors need some tweaking for the user side.
`invalid value "+-10" for flag -freq: can't contain both plus and minus symbols`
`invalid value "--10" for flag -freq: can't contain multiple minus symbols`
`invalid value "++10" for flag -freq: can't contain multiple plus symbols`
`invalid value ".1.0" for flag -freq: can't contain multiple decimal points`
`invalid value "1□" for flag -freq: unexpected end of string`
`invalid value "-10THz" for flag -freq: exceeds minimum for unit of -9.23THz`
`invalid value "10THz" for flag -freq: exceeds maximum for unit of 9.23THz`

conn/physic/units.go
```diff
@@ -217,17 +222,15 @@ func (f *Frequency) Set(s string) error {
 		if e, ok := err.(*parseError); ok {
 			switch e.err {
 			case errOverflowsInt64:
-				found, _ := containsUnitString(s[n:], "Hertz", "Hz")
-				return &parseError{
-					msg: "maximum value for \"" + found + "\" is " + e.msg + "n" + found,
-					err: errors.New(""),
-				}
+				return errors.New("maximum value is " + maxFrequency.String())
 			case errUnderflowsInt64:
```
maruel: FTR, underflow is having a number too small, i.e. 1pHz. A negative number is an overflow but in the negative side.

https://en.wikipedia.org/wiki/Arithmetic_underflow

conn/physic/units.go
```diff
@@ -773,6 +817,12 @@ const maxInt64 = (1<<63 - 1)
 
 var maxUint64Str = "9223372036854775807"
 
+var (
+	errOverflowsInt64  = errors.New("exceeds maximum")
+	errUnderflowsInt64 = errors.New("exceeds minimum")
```
maruel: That's OverflowNegative or something like that.
Since this is not exported, it's not a big deal here.

I think it'd be nice for underflow to actually return an error, but that's good for a follow up, since this change dragged on long enough.

conn/physic/units.go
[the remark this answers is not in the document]
maruel: Having slow(er) failing case is fine.

conn/physic/units.go
```diff
@@ -207,6 +209,23 @@ func (f Frequency) String() string {
 	return microAsString(int64(f)) + "Hz"
 }
 
+// Set sets the Frequency to the value represented by s. Units are to be
+// provided in hertz with optional SI prefixes.
```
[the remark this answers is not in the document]
maruel: I still think calling out "hz" and "hertz", in addition to the error message, is a good idea. The "customer" is different, here it's for the programmer.

maruel:
gohci

maruel:
Now I'm trying to decide if I do a v3.3.0 tag before or after. I think I'll do before, as the SPI fix is kinda important.

maruel:
Done; https://periph.io/news/2018/v3.3.0/

conn/physic/units_test.go
```diff
@@ -745,6 +745,133 @@ func TestMaxInt64(t *testing.T) {
...
+		if err != nil {
+			t.Errorf("set(%s,%d) got unexpected error: %v", tt.in, tt.uintbase, err)
+		}
+	}
+
+	for _, tt := range fails {
+		_, _, err := set(tt.in, tt.prefix)
+
+		if err == nil {
+			t.Errorf("set(%s,%d) got expected error but got none", tt.in, tt.prefix)
```
nkovacs: Nitpick: the first got shouldn't be there (set(%s,%d) <s>got</s> expected error but got none)

conn/physic/units_test.go
```diff
@@ -745,6 +745,133 @@ func TestMaxInt64(t *testing.T) {
...
+		}
+	}
+
+	for _, tt := range fails {
+		var got Frequency
+
+		err := got.Set(tt.in)
+
+		if err == nil {
+			t.Errorf("Frequency.Set(%s) got expected error but got none", tt.in)
```
nkovacs: here too

conn/physic/units_test.go
NeuralSpaz: I'll fix on with follow up.

conn/physic/units_test.go
NeuralSpaz: along with this one.

PR #317, opened 2018-11-07
ads1x15: refactor, simplify API

maruel:
- Merge PinForDifferenceOfChannels into PinForChannel.
- Make Channel a type.
- Rename ads1x15AnalogPin to analogPin, it was stuttering.
- Shorten few argument names.
- Remove one memory allocation in analogPin.
- Add a minimalistic unit test.

### DIFF
--- a/experimental/devices/ads1x15/ads1x15.go
+++ b/experimental/devices/ads1x15/ads1x15.go
@@ -17,15 +17,72 @@ import (
 	"periph.io/x/periph/conn/pin"
 )
 
-const (
-	// I2CAddr is the default I2C address for the ADS1x15 components.
-	I2CAddr uint16 = 0x48
+// I2CAddr is the default I2C address for the ADS1x15 components.
+const I2CAddr uint16 = 0x48
 
-	Channel0 = 0
-	Channel1 = 1
-	Channel2 = 2
-	Channel3 = 3
+// Channel is the analog reading to do. It can be either an absolute reading or
+// a differential reading between two pins.
+type Channel int
+
+const (
+	// Absolute reading.
+	Channel0 Channel = 4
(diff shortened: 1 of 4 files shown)


maruel:
@bezineb5  any strong opinion about `Channel`?

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -20,13 +20,72 @@ import (
...
+	case Channel0:
+		return 0
+	case Channel1:
+		return 1
+	case Channel2:
+		return 2
+	case Channel3:
+		return 3
+	case Channel0Minus1:
+		return 4
```
maruel: That's kinda adhoc.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -383,36 +397,37 @@ var (
 	}
 )
 
-type ads1x15AnalogPin struct {
+type analogPin struct {
 	adc                *Dev
-	query              []byte
+	c                  Channel
+	query              [3]byte
```
maruel: This removes one memory allocation.

maruel:
@bezineb5 I would still appreciate your comments on how the API surface changed.

bezineb5:
Hi,

Sorry, I was unexpectedly without internet this week :/.
Your changes are really great! It's much improved now. For the "difference of channel", I agree with the proposal, as the device is anyway too constrained to allow the difference of any pin with any other pin, so the user has to know the device anyway.
I'll do another PR to fix the remaining points (missing a mutex + typo). I'll also add more unit tests to ensure that the decoding is done correctly.

Thanks a lot!

maruel:
Thanks!

PR #318, opened 2018-11-07
analog: improve PinADC and PinDAC

maruel:
Rename the interfaces to be clear about the fact that they represent a single
pin.

Add Reading.

### DIFF
--- a/experimental/conn/analog/analog.go
+++ b/experimental/conn/analog/analog.go
@@ -2,34 +2,46 @@
 // Use of this source code is governed under the Apache License, Version 2.0
 // that can be found in the LICENSE file.
 
-// Package analog defines analog pins, both DAC and ADC.
+// Package analog defines analog pins, both digital to analog converter (DAC)
+// and analog to digital converter (ADC).
 package analog
 
 import (
 	"errors"
 
+	"periph.io/x/periph/conn/physic"
 	"periph.io/x/periph/conn/pin"
 )
 
-// ADC is an analog-to-digital-conversion input.
-type ADC interface {
+// Reading is the result of PinADC.Read().
+type Reading struct {
+	// V is the interpreted electrical level.
+	V physic.ElectricPotential
+	// Raw is the raw measurement.
+	Raw int32
+}
+
+// PinADC is an analog-to-digital-conversion input.
+type PinADC interface {
 	pin.Pin
 	// Range returns the maximum supported range [min, max] of the values.
-	Range() (int32, int32)
+	Range() (Reading, Reading)
 	// Read returns the current pin level.
-	Read() int32
+	Read() (Reading, error)
 }
 
-// DAC is an digital-to-analog-conversion output.
-type DAC interface {
+// PinDAC is an digital-to-analog-conversion output.
+type PinDAC interface {
 	pin.Pin
 	// Range returns the maximum supported range [min, max] of the values.
-	Range() (int32, int32)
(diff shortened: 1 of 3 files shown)


maruel:
Paging @bezineb5 and @davidsansome.

This PR conflicts with #317. I can commit one or the other first, it's not a big deal.

I'm still quite ambivalent about this PR, I'll add comments inline.

experimental/conn/analog/analog.go
```diff
@@ -2,34 +2,46 @@
...
-// DAC is an digital-to-analog-conversion output.
-type DAC interface {
+// PinDAC is an digital-to-analog-conversion output.
+type PinDAC interface {
 	pin.Pin
 	// Range returns the maximum supported range [min, max] of the values.
 	Range() (int32, int32)
 	// Out sets an analog output value.
-	DAC(v int32)
+	Out(v int32) error
```
maruel: I do not see much value in the DAC to accept an ElectricPotential value.
The challenge here is that this makes DAC and ADC not symmetrical. This is especially visible with Range(). This means that a pin cannot implement both interfaces simultaneously.
In practice I wouldn't expect an analog pin to be directional; unlike digital pins (gpio) which are usually bidirectional, analog pins rarely are. That said I would prefer to not have this issue.

One option would be to have Range() here to return Reading, even if it may not make sense, I don't know.

experimental/conn/analog/analog.go
```diff
@@ -2,34 +2,46 @@
...
 }
 
-// INVALID implements both ADC and DAC and fails on all access.
+/*
```
maruel: I commented out this since the current interfaces makes it impossible to implement both interfaces simultaneously.

maruel:
I'm now relatively happy with this change so I'll commit in, it's a net improvement from the status quo. Still looking for comments from @bezineb5 and @davidsansome.

davidsansome:
Sorry for not commenting on this sooner!

This looks awesome.  I'm not sure whether it'll be useful for the DAC's Out to support taking a Reading - if the DAC knows how to convert a voltage to a raw value it might be useful?

maruel:
I think the value is fairly marginal, that's why it takes an int32 as input for now.

bezineb5:
I did have a look at available DACs on the market:
- https://www.adafruit.com/product/935
For that device, the data sent to the DAC is a percentage of the input voltage, which can be anywhere between 3 and 5V. So it means that the actual voltage is not known by the user unless he configures the driver to know it.
- https://www.adafruit.com/product/3678
For that device, it's a fixed maximum voltage, 3.3V. However, it seems to be targeting audio, so the voltage might not be so useful in that case.

=> So I agree with the interface, I don't think the voltage makes as much sense for DAC as for ADC. The only change I could think of is to rename "reading" struct as "sample" (or similar) to be more direction agnostic.

Eventually, we could consider some alternatives:
* introduce a "percentage" unit for them, and the range would specify the max/min voltage if known. But that would make the life of the user a bit harder.
* consider that ADC and DAC are just kinds of "Sensor" and "Generator" (or Reader/Writer) of type physic.ElectricPotential. Unfortunately, as there are no generics in Go, I don't see an easy way to implement that.

maruel:
Thanks for the comment.
- I'll rename Reading to Sample, it makes much more sense.
- Percentage can be calculated from Range() so I'm not fond of doing this yet.

PR #320, opened 2018-11-09
pca9548: i2c multiplexer

NeuralSpaz:
Adds initial experimental support a i2c multiplexer. Although only tested with NXP pca9548 it should be compatible with many i2c multiplexers IC's.

### DIFF
--- a/experimental/cmd/pca9548/main.go
+++ b/experimental/cmd/pca9548/main.go
@@ -33,14 +33,14 @@ func mainImpl() error {
 	}
 	defer bus.Close()
 
-	// Registers a multiplexer with 8 ports at address 0x70 if no other address
+	// Creates a multiplexer with 8 ports at address 0x70 if no other address
 	// supplied with command line option.
 	mux, err := pca9548.New(bus, &pca9548.Opts{Address: *address})
 	if err != nil {
 		return fmt.Errorf("failed to load new mux: %v", err)
 	}
 
-	// Register all 8 ports
+	// Register all 8 ports.
 	busz := make(map[int]string)
 	for i := 0; i < 8; i++ {
 		name := "mux" + strconv.Itoa(i)
@@ -51,7 +51,7 @@ func mainImpl() error {
 		busz[i] = name
 	}
 
-	// Create a place to store the results from the scan
+	// Create a place to store the results from the scan.
 	results := make(map[string][]uint16)
 	fmt.Println("Starting Scan")
 	// Loop through each bus scanning all addresses for a response.
--- a/experimental/cmd/pca9548/main.go
+++ b/experimental/cmd/pca9548/main.go
@@ -9,6 +9,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
 
 	"periph.io/x/periph/conn/i2c/i2creg"
 	"periph.io/x/periph/experimental/devices/pca9548"
@@ -34,30 +35,43 @@ func mainImpl() error {
 
 	// Registers a multiplexer with 8 ports at address 0x70 if no other address
 	// supplied with command line option.
-	_, err = pca9548.Register(bus, &pca9548.Opts{Address: *address})
+	mux, err := pca9548.New(bus, &pca9548.Opts{Address: *address})
 	if err != nil {
 		return fmt.Errorf("failed to load new mux: %v", err)
 	}
 
-	devices := make(map[string][]uint16)
-	fmt.Println("Starting Scan")
+	// Register all 8 ports
+	busz := make(map[int]string)
 	for i := 0; i < 8; i++ {
-		busname := fmt.Sprintf("mux-%x-%d", *address, i)
-		mux, err := i2creg.Open(busname)
-		defer mux.Close()
-		rx := []byte{0x00}
+		name := "mux" + strconv.Itoa(i)
+		err = mux.Register(i, name)
 		if err != nil {
 			return err
 		}
-		for addr := uint16(0); addr < 0x77; addr++ {
-			if err := mux.Tx(addr, nil, rx); err == nil {
-				devices[busname] = append(devices[busname], addr)
+		busz[i] = name
+	}
+
+	// Create a place to store the results from the scan
+	results := make(map[string][]uint16)
+	fmt.Println("Starting Scan")
+	// Loop through each bus scanning all addresses for a response.
+	for i := 0; i < 8; i++ {
+		m, err := i2creg.Open(busz[i])
+		if err != nil {
+			return err
+		}
+		defer m.Close()
+
+		rx := []byte{0x00}
+		for addr := uint16(1); addr < 0x77; addr++ {
+			if err := m.Tx(addr, nil, rx); err == nil {
+				results[busz[i]] = append(results[busz[i]], addr)
 			}
 		}
 	}
 
 	fmt.Println("Scan Results:")
-	for bus, addrs := range devices {
+	for bus, addrs := range results {
 		fmt.Printf("Bus[%s] %d found\n", bus, len(addrs))
 		for _, addr := range addrs {
 			fmt.Printf("\tDevice at %#2x\n", addr)
--- a/experimental/cmd/pca9548/main.go
+++ b/experimental/cmd/pca9548/main.go
@@ -1,0 +1,75 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// pca9548 scans the 8 ports of a pca9548 i2c multiplexer for other i2c devices.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	"periph.io/x/periph/conn/i2c/i2creg"
+	"periph.io/x/periph/experimental/devices/pca9548"
+	"periph.io/x/periph/host"
+)
+
+func mainImpl() error {
+	address := flag.Int("address", 0x70, "I²C address")
+	i2cbus := flag.String("bus", "", "I²C bus (/dev/i2c-1)")
+
+	flag.Parse()
+
+	if _, err := host.Init(); err != nil {
+		return err
+	}
+
+	// Open default I²C bus.
+	bus, err := i2creg.Open(*i2cbus)
+	if err != nil {
+		return fmt.Errorf("failed to open I²C: %v", err)
+	}
+	defer bus.Close()
+
+	// Registers a multiplexer with 8 ports at address 0x70 if no other address
+	// supplied with command line option.
+	_, err = pca9548.Register(bus, &pca9548.Opts{Address: *address})
+	if err != nil {
+		return fmt.Errorf("failed to load new mux: %v", err)
+	}
+
+	devices := make(map[string][]uint16)
+	fmt.Println("Starting Scan")
+	for i := 0; i < 8; i++ {
+		busname := fmt.Sprintf("mux-%x-%d", *address, i)
+		mux, err := i2creg.Open(busname)
+		defer mux.Close()
+		rx := []byte{0x00}
+		if err != nil {
+			return err
+		}
+		for addr := uint16(0); addr < 0x77; addr++ {
+			if err := mux.Tx(addr, nil, rx); err == nil {
+				devices[busname] = append(devices[busname], addr)
+			}
+		}
+	}
+
+	fmt.Println("Scan Results:")
+	for bus, addrs := range devices {
+		fmt.Printf("Bus[%s] %d found\n", bus, len(addrs))
+		for _, addr := range addrs {
+			fmt.Printf("\tDevice at %#2x\n", addr)
+		}
+	}
+
+	return nil
+}
+
+func main() {
+	if err := mainImpl(); err != nil {
+		fmt.Fprintf(os.Stderr, "pca9548: %s.\n", err)
+		return
+	}
+}
--- a/experimental/cmd/pca9548/main.go
+++ b/experimental/cmd/pca9548/main.go
@@ -40,23 +40,18 @@ func mainImpl() error {
 		return fmt.Errorf("failed to load new mux: %v", err)
 	}
 
-	// Register all 8 ports.
-	busz := make(map[int]string)
-	for i := 0; i < 8; i++ {
-		name := "mux" + strconv.Itoa(i)
-		err = mux.Register(i, name)
-		if err != nil {
-			return err
-		}
-		busz[i] = name
+	if err := mux.RegisterPorts("mux"); err != nil {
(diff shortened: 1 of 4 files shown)


experimental/devices/pca9548/doc.go
```diff
@@ -0,0 +1,32 @@
...
+// Package pca9548 is a driver for an 8 port I²C multiplexer that is avaliable
+// from multiple venders. The main features of this multiplexer is that its has
+// 8 channels and is capable of voltage level translation.
+//
+//
+// Use with other I²C multiplexers
+//
+// Most I²C multiplexers work the same way as
+// the pca9548 so it could be possible to use this driver with a range of other
+// I²C multiplexers inclding 2 and 4 port versions.
```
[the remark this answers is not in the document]
NeuralSpaz: I had problems naming the package, as the three difference between the chips is the number of ports, the 2 and 4 ports often have "notify interrupt" on the slave side and the pin names.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// ListPortNames lists all the port names registered for the multiplexer.
+func (d *Dev) ListPortNames() []string {
+	return d.names
+}
+
+// Scan scans every port of the multiplexer and every I²C address for devices,
+// returns a map of I²C multiplexer ports names to slice of device I²C address.
+func (d *Dev) Scan() ScanList {
```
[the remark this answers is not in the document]
NeuralSpaz: I've been searching for a good example that does not depend on anything else.
That is where I can put scan.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,155 @@
...
+
+	for i := uint8(0); i < opts.Ports; i++ {
+		portID := strconv.FormatUint(uint64(i), 10)
+		addrStr := strconv.FormatUint(uint64(opts.Address), 16)
+		name := addrStr + "-" + portID
+		opener := newOpener(d, i)
+		if err := i2creg.Register("i2cmux-"+name, []string{"mux-" + name}, int((opts.Address*10)+uint16(i)), opener); err != nil {
+			return nil, err
+		}
+		d.names = append(d.names, "i2cmux-"+name)
```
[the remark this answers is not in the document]
NeuralSpaz: like "pca9548-70-1" and "mux-70-1"?

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,155 @@
...
+	// Ports number of physical ports on I²C Multiplexer.
+	Ports uint8
+}
+
+// Register creates a new handel to a pca9548 I²C multiplexer, and registers
+// port names with the host. These ports can then be used as any other i2c.Bus.
+// The registered port names are in the form: i2cmux/mux-ADD-I where ADD is the
+// multiplexer I²C address in hex and I is the port number.
+// example: "i2cmux-70-0" and "mux-70-0".
+func Register(bus i2c.Bus, opts *Opts) (*Dev, error) {
```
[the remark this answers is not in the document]
NeuralSpaz: That is a good question. And maybe its a bad way of doing it. But since the user does not interact with the mux directly except for scan() (now removed to an example) and listports() which I'm a little dubious on now. I wanted the function name to reflect what it was doing.

```golang 
	bus, err := i2creg.Open(*i2cbus)
	if err != nil {
		return fmt.Errorf("failed to open I²C: %v", err)
	}
	defer bus.Close()

	_, err := pca9548.Register(bus, &pca9548.DefaultOpts)
	if err != nil {
		t.Errorf("failed to open I²C: %v", err)
	}

	mux0, err := i2creg.Open("mux-70-0")
	if err != nil {
		return fmt.Errorf("failed to open I²C: %v", err)
	}
	defer mux0.Close()
 	....      
```

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// DefaultOpts is the recommended default options.
+var DefaultOpts = Opts{Address: 0x70, Ports: 8}
+
+// Opts is the pca9548 configuration.
+type Opts struct {
+	// Address pca9548 I²C Address.
+	Address uint16
+	// Ports number of physical ports on I²C Multiplexer.
```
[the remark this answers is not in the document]
NeuralSpaz: One control register is all you get with these, so sadly no. It is an extremely simple device. (At least on the version I'm using an others that I remember). If I spit the code into 5 different packages for 2, 4 and 8 port version then number of ports is fixed. 
eg pca9548(8), pca9546(4), pca9540(2). then pca9545(4 with interrupts) pca9543(2 with interrupt)? Its possibly a better solution.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,155 @@
...
+		address: opts.Address,
+		ports:   opts.Ports,
+	}
+
+	for i := uint8(0); i < opts.Ports; i++ {
+		portID := strconv.FormatUint(uint64(i), 10)
+		addrStr := strconv.FormatUint(uint64(opts.Address), 16)
+		name := addrStr + "-" + portID
+		opener := newOpener(d, i)
+		if err := i2creg.Register("i2cmux-"+name, []string{"mux-" + name}, int((opts.Address*10)+uint16(i)), opener); err != nil {
```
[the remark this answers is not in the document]
NeuralSpaz: Just for the behavior described about. I like that they are ready to use. But you make me think that I should check to see the IC is there before registering the ports. That way it can fail right there rather than later in the middle of unrelated code.

experimental/devices/pca9548/doc.go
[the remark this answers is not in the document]
maruel: ok, it's fine to start with something specific and generalize later.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,139 @@
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if p.mux == nil {
+		return errors.New("port" + strconv.FormatUint(uint64(p.number), 10) + "has been closed")
+	}
+	return p.mux.tx(p.number, addr, w, r)
+}
+
+func (p *port) Close() error {
+	p.mux = nil
```
maruel: Take the lock before doing so.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,139 @@
...
+func Register(bus i2c.Bus, opts *Opts) (*Dev, error) {
+	d := &Dev{
+		c:       bus,
+		port:    0xFF,
+		address: uint16(opts.Address),
+	}
+	r := make([]byte, 1)
+	err := bus.Tx(uint16(opts.Address), nil, r)
+	if err != nil {
+		return nil, errors.New("could not communicated with multiplexer: " + err.Error())
```
maruel: communicate

experimental/devices/pca9548/pca9548.go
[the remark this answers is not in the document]
maruel: I'd prefer to split responsibilities:
- the dev enables multiplexing, exposing multiple sub-bus
- registration of such buses
- the scan, as discussed before

There's no way for the user to unregister after.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	for i := uint8(0); i < 8; i++ {
+		portID := strconv.FormatUint(uint64(i), 10)
+		addrStr := strconv.FormatUint(uint64(opts.Address), 16)
+		name := addrStr + "-" + portID
+		opener := newOpener(d, i)
+		if err := i2creg.Register("pca9548-"+name, []string{"mux-" + name}, int((opts.Address*10 + int(i))), opener); err != nil {
```
maruel: In practice, this would fail if an host instantiate this driver twice. There's nothing on that hardware side to do so. Even funnier would be to daisy chain one.

experimental/cmd/pca9548/main.go
```diff
@@ -0,0 +1,75 @@
...
+	// Registers a multiplexer with 8 ports at address 0x70 if no other address
+	// supplied with command line option.
+	_, err = pca9548.Register(bus, &pca9548.Opts{Address: *address})
+	if err != nil {
+		return fmt.Errorf("failed to load new mux: %v", err)
+	}
+
+	devices := make(map[string][]uint16)
+	fmt.Println("Starting Scan")
+	for i := 0; i < 8; i++ {
```
maruel: Could you not hard code it, could the number of ports exposed as a method on Dev? This would open the door to support different chip variation.

experimental/cmd/pca9548/main.go
```diff
@@ -0,0 +1,75 @@
...
+		busname := fmt.Sprintf("mux-%x-%d", *address, i)
+		mux, err := i2creg.Open(busname)
```
maruel: I'd prefer if one could open these as a method on the Dev.

experimental/devices/pca9548/pca9548.go
NeuralSpaz: I'm really not having a good day : (

experimental/devices/pca9548/pca9548.go
NeuralSpaz: tbh, I'm not sure on the difference between the alias and the name, should the alias be provided by the user eg:
```golang
 func (d *Dev) Register(port int, alias string) error
```
might make it a bit more friendly as they would then know what the port is called or does this make more sense?

```golang
 func (d *Dev) Open(port int) (i2c.Bus, error)
```

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Halt does nothing.
+func (d *Dev) Halt() error {
+	// TODO(NeuralSpaz): Find a good way to halt that also can resume.
+	return nil
+}
+
+// String returns the bus base name for multiplexer ports.
+func (d *Dev) String() string {
+	return "pca9548-" + strconv.FormatUint(uint64(d.address), 16)
```
maruel: Since this is going to be called frequently, I'd recommend to compute this value inside New() so here it's just `return d.name` to reduce heap churn.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,149 @@
...
+		err := d.c.Tx(d.address, []byte{uint8(1 << port)}, nil)
+		if err != nil {
+			return errors.New("failed to change active port on multiplexer: " + err.Error())
+		}
+		d.port = port
+	}
+	return d.c.Tx(address, w, r)
+}
+
+// Port is a i2c.Bus on the multiplexer.
```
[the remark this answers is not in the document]
NeuralSpaz: I've been meaning to ask what that syntax of x/x/x is? I'm guessing at intent half the time.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Close closes a port, no futher communication can take place on that port.
+func (p *port) Close() error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.mux = nil
+	return nil
+}
+
+// newOpener is a helper for creating an opener func.
```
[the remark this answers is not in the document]
NeuralSpaz: I still want the user to do i2creg.Open(), as the mux could easily be place in existing code. I think having an pca9548.Open(port int) would be less useful than doing:
```golang 
	mux, err := pca9548.New(bus, &pca9548.DefaultOpts)
	if err != nil {
		return err
	}
	if err := mux.RegisterPorts("mux"); err != nil {
		return err
	}

	m, err := i2creg.Open("mux0")
	defer m.Close()
	....
```
or have I miss understood/missing something? is i2creg.Opener somewhere else.

experimental/cmd/pca9548/main.go
maruel: I'd still prefer `8` to not be hardcoded, and instead retrieved from mux.

experimental/cmd/pca9548/main.go
```diff
@@ -0,0 +1,84 @@
...
+	fmt.Println("Starting Scan")
+
+	// Loop through each bus scanning all addresses for a response.
+	for i := 0; i < 8; i++ {
+		// Open multiplexer port with alias mux0-mux7.
+		m, err := i2creg.Open("mux" + strconv.Itoa(i))
+		if err != nil {
+			return err
+		}
+		defer m.Close()
```
maruel: defer doesn't happen at the end of the loop, but at the end of the function. So I'd recommend not using defer and and instead a real call around line 65.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,158 @@
...
+
+// DefaultOpts is the recommended default options.
+var DefaultOpts = Opts{Address: 0x70}
+
+// Opts is the pca9548 configuration.
+type Opts struct {
+	// Address pca9548 I²C Address. Valid addresses for the NXP pca9548 are 0x70
+	// to 0x77. The address is set by pulling A0~A2 low or high. Please refer
+	// to the datasheet.
+	Address int
```
maruel: I think we should standardize on `Addr` everywhere. I've realized there's also `I2CAddress` used, will update that independently.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, errors.New("could not establish communicate with multiplexer: " + err.Error())
+	}
+	return d, nil
+}
+
+// RegisterPorts registers multiplexer ports with the host. These ports can
+// then be used as any other i2c.Bus. Busses will be named "alias0", "alias1"
+// etc. If using more than one multiplexer note that the alias must be unique.
+func (d *Dev) RegisterPorts(alias string) error {
+	d.mu.Lock()
```
maruel: I don't think this actually requires the lock, since it's only accessing immutable code. And it wouldn't be a big deal if a Tx() call would happen concurrently.

experimental/devices/pca9548/pca9548.go
```diff
@@ -0,0 +1,158 @@
...
+	defer d.mu.Unlock()
+	for i := uint8(0); i < d.numPorts; i++ {
+		portStr := strconv.Itoa(int(i))
+		addrStr := strconv.FormatUint(uint64(d.address), 16)
+		portName := d.c.String() + "-pca9548-" + addrStr + "-" + portStr
+		opener := newOpener(d, i, alias+portStr, portName)
+		err := i2creg.Register(portName, []string{alias + portStr}, -1, opener)
```
maruel: Inline into next line.

maruel:
Please address win18 failure.

PR #324, opened 2018-11-10
physic: Add flag.Value implementation for ElectricResistance

NeuralSpaz:
Add Set().
Add tests.
Add SI Units to docs.
Add Examples.

### DIFF
--- a/conn/physic/doc.go
+++ b/conn/physic/doc.go
@@ -7,18 +7,15 @@
 //
 // This includes temperature, humidity, pressure, tension, current, etc.
 //
-//
 // SI units
 //
 // The supported S.I. units is a subset of the official ones.
-//   T		tera		10¹²		1000000000000
-//   G		giga		10⁹			1000000000
-//   M		mega		10⁶			1000000
-//   k		kilo		10³			1000
-//   m		milli		10⁻³		0.001
-//   µ,u		micro		10⁻⁶		0.000001
-//   n		nano		10⁻⁹		0.000000001
-//   p		pico		10⁻¹²		0.000000000001
-//
-//
+//    T  	tera 	10¹²  	1000000000000
+//    G  	giga 	10⁹   	1000000000
+//    M  	mega 	10⁶   	1000000
+//    k  	kilo 	10³   	1000
+//    m  	milli	10⁻³  	0.001
+//    µ,u	micro	10⁻⁶  	0.000001
+//    n  	nano 	10⁻⁹  	0.000000001
+//    p  	pico 	10⁻¹² 	0.000000000001
 package physic
--- a/conn/physic/doc.go
+++ b/conn/physic/doc.go
@@ -6,4 +6,19 @@
 // units.
 //
 // This includes temperature, humidity, pressure, tension, current, etc.
+//
+//
+// SI units
+//
+// The supported S.I. units is a subset of the official ones.
+//   T		tera		10¹²		1000000000000
+//   G		giga		10⁹			1000000000
+//   M		mega		10⁶			1000000
+//   k		kilo		10³			1000
+//   m		milli		10⁻³		0.001
+//   µ,u		micro		10⁻⁶		0.000001
+//   n		nano		10⁻⁹		0.000000001
+//   p		pico		10⁻¹²		0.000000000001
+//
+//
 package physic
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -155,8 +155,8 @@ func (e ElectricResistance) String() string {
 }
 
 // Set sets the ElectricResistance to the value represented by s. Units are to
-// be provided in "Ohm", "Ohms" or "Ω" with an optional SI prefix: "p", "n", "u"
-// , "µ", "m", "k", "M", "G" or "T".
+// be provided in "Ohm", "Ohms" or "Ω" with an optional SI prefix: "p", "n",
+//  "u", "µ", "m", "k", "M", "G" or "T".
 func (f *ElectricResistance) Set(s string) error {
 	v, n, err := valueOfUnitString(s, nano)
 	if err != nil {
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -156,7 +156,7 @@ func (e ElectricResistance) String() string {
 
 // Set sets the ElectricResistance to the value represented by s. Units are to
 // be provided in "Ohm", "Ohms" or "Ω" with an optional SI prefix: "p", "n",
-//  "u", "µ", "m", "k", "M", "G" or "T".
+// "u", "µ", "m", "k", "M", "G" or "T".
 func (f *ElectricResistance) Set(s string) error {
 	v, n, err := valueOfUnitString(s, nano)
 	if err != nil {
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -155,8 +155,8 @@ func (e ElectricResistance) String() string {
 }
 
 // Set sets the ElectricResistance to the value represented by s. Units are to
-// be provided in Ohm/Ohms or Ω with an optional SI prefix (n,p,u,µ,m,k,M,G or
-// T)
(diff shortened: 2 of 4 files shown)


conn/physic/doc.go
```diff
@@ -6,4 +6,19 @@
 // units.
 //
 // This includes temperature, humidity, pressure, tension, current, etc.
+//
+//
+// SI units
+//
+// The supported S.I. units is a subset of the official ones.
```
NeuralSpaz: If you know trick for to always to produce columns in godoc?

conn/physic/doc.go
```diff
@@ -6,4 +6,19 @@
 // units.
 //
 // This includes temperature, humidity, pressure, tension, current, etc.
+//
```
maruel: one "empty" line is sufficient.

conn/physic/doc.go
maruel: No but you can do a monospace block as I think you did (I forget if +3 is sufficient or if you need +4).

conn/physic/doc.go
```diff
@@ -6,4 +6,19 @@
...
+//
+// SI units
+//
+// The supported S.I. units is a subset of the official ones.
+//   T		tera		10¹²		1000000000000
+//   G		giga		10⁹			1000000000
```
maruel: Use space characters instead of tabs to ensure alignment is 100% consistent.

conn/physic/doc.go
```diff
@@ -6,4 +6,19 @@
...
+//   M		mega		10⁶			1000000
+//   k		kilo		10³			1000
+//   m		milli		10⁻³		0.001
+//   µ,u		micro		10⁻⁶		0.000001
+//   n		nano		10⁻⁹		0.000000001
+//   p		pico		10⁻¹²		0.000000000001
+//
```
maruel: you can remove these two lines

conn/physic/units.go
```diff
@@ -154,6 +154,44 @@ func (e ElectricResistance) String() string {
 	return nanoAsString(int64(e)) + "Ω"
 }
 
+// Set sets the ElectricResistance to the value represented by s. Units are to
+// be provided in Ohm/Ohms or Ω with an optional SI prefix (n,p,u,µ,m,k,M,G or
```
maruel: I'd recommend to use quotes:
// be provided in "Ohm"/"Ohms" or "Ω" with an optional SI prefix: "n", "p", "u" or "µ", "m", "k", "M", "G" or
// "T".

conn/physic/units.go
```diff
@@ -154,6 +154,44 @@ func (e ElectricResistance) String() string {
...
+	v, n, err := valueOfUnitString(s, nano)
+	if err != nil {
+		if e, ok := err.(*parseError); ok {
+			switch e.err {
+			case errOverflowsInt64:
+				return errors.New("maximum value is " + maxElectricResistance.String())
+			case errUnderflowsInt64:
+				return errors.New("minimum value is " + minElectricResistance.String())
+			case errNotANumber:
+				found, _ := containsUnitString(s, "Ohm", "Ohms", "Ω")
```
maruel: you can inline line 170 into line 171.

conn/physic/units.go
```diff
@@ -154,6 +154,44 @@ func (e ElectricResistance) String() string {
...
+					return errors.New("does not contain number")
+				}
+				return errors.New("does not contain number or unit \"Ohm\"")
+			}
+		}
+		return err
+	}
+
+	switch strings.ToLower(s[n:]) {
+	case "ohm", "ohms", "ω": // ω is lowercase Ω
```
maruel: Ah! I think we should not support "ω" as this would be confusing. An option is to first check manually, then do the rest:
```
if rest := s[n:]; rest == "Ω" {
  *f = (ElectricResistance)(v)
} else {
  switch strings.Lower(rest) {
  case "ohm", "ohms":
      *f = (ElectricResistance)(v)
  case "":
    ..
  default:
    ..
  }
}
```

conn/physic/units_test.go
```diff
@@ -857,6 +857,140 @@ func TestValueOfUnitString(t *testing.T) {
...
+			t.Errorf("ElectricResistance.Set(%s) expected: %v(%d) but got: %v(%d)", tt.in, tt.expected, tt.expected, got, got)
+		}
+		if err != nil {
+			t.Errorf("ElectricResistance.Set(%s) got unexpected error: %v", tt.in, err)
+		}
+	}
+
+	for _, tt := range fails {
+		var got ElectricResistance
+
```
maruel: Please densify:
- remove empty lines
- inline line 986 into line 988.

conn/physic/doc.go
NeuralSpaz: I couldn't quite get it right with just spaces in both the editor and godoc, I had to use spaces and tabs. I'm sure its to fragile to survive.

conn/physic/units.go
NeuralSpaz: If == Ω I prefer it it just return right there since the rest of the code has nothing to do with it any more.

conn/physic/units.go
```diff
@@ -154,6 +154,47 @@ func (e ElectricResistance) String() string {
...
+				if found, _ := containsUnitString(s, "Ohm", "Ohms", "Ω"); found != "" {
+					return errors.New("does not contain number")
+				}
+				return errors.New("does not contain number or unit \"Ohm\"")
+			}
+		}
+		return err
+	}
+
+	if s[n:] == "Ω" {
```
maruel: The reason I did `rest := s[n:]` is that I don't think the Go compiler will be smart enough to optimize the two construction, especially with bound checks.

conn/physic/units.go
```diff
@@ -154,6 +154,47 @@ func (e ElectricResistance) String() string {
...
+		*f = (ElectricResistance)(v)
+		return nil
+	}
+	switch strings.ToLower(s[n:]) {
+	case "ohm", "ohms":
+		*f = (ElectricResistance)(v)
+	case "":
+		return noUnits("Ohm")
+	default:
+		found, extra := containsUnitString(s[n:], "Ohm", "Ohm", "Ω")
```
maruel: Please inline with next line

conn/physic/units_test.go
```diff
@@ -857,6 +857,137 @@ func TestValueOfUnitString(t *testing.T) {
...
+		{
+			"1.1.1.1Ohm",
+			"multiple decimal points 1.1.1.1Ohm",
+		},
+	}
+
+	for _, tt := range succeeds {
+		var got ElectricResistance
+		err := got.Set(tt.in)
+		if got != tt.expected {
```
maruel: I'd recommend do:
```
if err := got.Set(tt.in); err != {
   ..
}
```
then check for `got != tt.expected`. The reason is that the error message is likely more informative than the incorrect value. And you save one line. :)

conn/physic/units.go
```diff
@@ -154,6 +154,46 @@ func (e ElectricResistance) String() string {
 	return nanoAsString(int64(e)) + "Ω"
 }
 
+// Set sets the ElectricResistance to the value represented by s. Units are to
+// be provided in "Ohm", "Ohms" or "Ω" with an optional SI prefix: "n", "p", "u"
```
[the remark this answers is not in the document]
maruel: Please fix the order, put p before n.

conn/physic/units.go
```diff
@@ -154,6 +154,46 @@ func (e ElectricResistance) String() string {
 	return nanoAsString(int64(e)) + "Ω"
 }
 
+// Set sets the ElectricResistance to the value represented by s. Units are to
+// be provided in "Ohm", "Ohms" or "Ω" with an optional SI prefix: "p", "n", "u"
+// , "µ", "m", "k", "M", "G" or "T".
```
maruel: That will put a space between `"u"` and `,`

PR #325, opened 2018-11-10
ads1x15: fix synchronization

bezineb5:
I fixed the typo and the lack of mutex when calling Halt() on a pin.
Added unit tests based upon a real device.

Also, I added a convenience String() function on i2ctest.IO, in order to make it easier to write unit tests. Do you think it has value or do you prefer me to remove it?

Last but not least, should I had those devices in the documentation on periph.io? Where is it?

Thanks!

### DIFF
--- a/experimental/devices/ads1x15/ads1x15.go
+++ b/experimental/devices/ads1x15/ads1x15.go
@@ -90,7 +90,7 @@ type ConversionQuality int
 
 const (
 	// SaveEnergy optimizes the power consumption of the ADC, at the expense of
-	// the quality by converting at the gihest rate possible.
+	// the quality by converting at the highest possible rate.
 	SaveEnergy ConversionQuality = 0
 	// BestQuality will use the lowest suitable data rate to reduce the impact of
 	// the noise on the reading.
@@ -394,6 +394,7 @@ type analogPin struct {
 	waitTime           time.Duration
 	requestedFrequency physic.Frequency
 	stop               chan struct{}
+	mu                 sync.Mutex
 }
 
 // Range returns the maximum supported range [min, max] of the values.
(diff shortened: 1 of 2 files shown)


conn/i2c/i2ctest/i2ctest.go
```diff
@@ -22,6 +23,17 @@ type IO struct {
 	R    []byte
 }
 
+// String will return a golang-formatted version of of the IO so that it can be easily copy-pasted into
+// unit tests
+func (i *IO) String() string {
```
maruel: This should be GoString() then. See https://golang.org/pkg/fmt/#GoStringer

conn/i2c/i2ctest/i2ctest.go
```diff
@@ -7,6 +7,7 @@ package i2ctest
 
 import (
```
maruel: I'd prefer this to be a separate PR.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -409,14 +410,23 @@ func (p *analogPin) Read() (analog.Reading, error) {
...
-	reading := make(chan analog.Reading)
+	// We need to lock if there are multiple Halt or ReadContinuous
+	// calls simultaneously.
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	// First release the current continuous reading if there is one
+	if p.stop != nil {
+		p.stop <- struct{}{}
+		close(p.stop)
```
maruel: Technically speaking you don't need to close it. In fact it's probably preferable.

experimental/devices/ads1x15/ads1x15.go
```diff
@@ -450,7 +460,13 @@ func (p *analogPin) Function() string {
...
 
 func (p *analogPin) Halt() error {
+	// We need to lock if there are multiple Halt or ReadContinuous
+	// calls simultaneously.
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
 	if p.stop != nil {
+		p.stop <- struct{}{}
 		close(p.stop)
```
maruel: Same

maruel:
Where would you expect the documentation to find out how to make changes to the website? It's clearly not obvious at the moment by the sheer lack of website contributors but I don't know where I should advertise that.

bezineb5:
For the documentation, I would expect to see it in the CONTRIBUTION file or the equivalent section on the periph.io website. I hope I've not missed it, no?
Something which would be nice would be to have the documentation as markdown in each driver directory so that it's easier to maintain.

maruel:
Good point. I'll revisit https://github.com/periph/website/blob/master/site/content/project/contributing.md to be more to the point.

Issue #303, opened 2018-11-01
feature: physic.Units flag.Value interface


NeuralSpaz:
Probably should have a discussion on what should be parsed.
My thoughts are the format should be compatible with the physic.units stringers.

eg: value|SI Prefix|unit(in ascii or unicode)
2.78µA or as 2.78uA as a concession for type-ability.

[see here](https://github.com/NeuralSpaz/periph/commit/babdbea0dbdca828fe05f0acd8104833a66eac1f)

NeuralSpaz:
#242 could also be done at same time.

maruel:
Yes. Points to take in account:

1. Some quantities have an obvious default unit; ElectricTension is definitely Volts, Frequency is Hz, etc.
1. Some quantities can be specified in multiple base unit; Angle (rad, deg), Temperature (K, C, F), gpio.Duty (%, fraction), etc.
1. Some units can be specified in multiple ways; '°', '°C', 'K', 'C', '°F', 'F', so the code must be tolerant to a reasonable extent, and must nudge the user towards the accepted forms in the error message.

I think we can address this challenge by first addressing 1. for the quantities that are not affected by 2. We could then create a generic function to implement this.

Then we could address 2 based on specific challenges for each quantity.

Oh and I'd prefer to not require floating point calculation, mainly due to the on-going tinygo work to MCUs without FPU.

NeuralSpaz:
I was reasoning that a little FP on start up for a flag might be ok, as the value would still be the unit's type? But I think it can be done without the FP a little more effort. To use them as a physic.unit parser I would be more cautious about the FP.

I agree and think it is important to allow all sensible errors such a "upa" and "µPa" for microPasca etc.
 
I have roughed out working flag.Value implementations and test for all the physic units, there are gaps and needs some work.

I would like to have decimals as part of the api as "-u xx.yUnits" seems more intuitive and forgiving than "-u xxy00Units"?

extract:
```golang
func (p *Power) Set(s string) error {
	r := []rune(s)
	v, n, err := parseNumber(r)
	if err != nil {
		return err
	}
	prefix := prefix(none)
	if !(n == len(r)) {
		prefix = parseSIPrefix(r[n])
	}
	scale := math.Pow10(int(prefix - nano))
	if prefix != none {
		s = string(r[n+1:])
	} else {
		s = string(r[n:])
	}
	switch s {
	case "watt", "watts", "Watt", "Watts", "W", "w":
		fallthrough
	default:
[...]

PR #330, opened 2018-11-11
physic: Add flag.Value for ElectricPotential

NeuralSpaz:
Continuation of #303
This add physic.ElectricPotential to the units that implement the flag.Value interface. 

Add ElectricPotential Set().
Add ElectricPotential max/min consts.
Add ElectricPotential tests.


maruel:
gohci

PR #332, opened 2018-11-11
physic: add 4 new Energy constants

maruel:
These are BTU, WattSecond, WattHour and KiloWattHour.

I thought about adding Calorie but because there's so many variances, it's hard
to choose one. The BTU has similar issue but I choose the ISO value.

### DIFF
--- a/conn/physic/example_test.go
+++ b/conn/physic/example_test.go
@@ -423,8 +423,12 @@ func ExampleLuminousIntensity_flag() {
 
 func ExampleEnergy() {
 	fmt.Println(1 * physic.Joule)
+	fmt.Println(1 * physic.WattSecond)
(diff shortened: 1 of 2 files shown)


maruel:
I haven't added support to parse "kwh", I don't think it's a big deal for now.

conn/physic/example_test.go
```diff
@@ -423,8 +423,10 @@ func ExampleLuminousIntensity_flag() {
 
 func ExampleEnergy() {
 	fmt.Println(1 * physic.Joule)
+	fmt.Println(1 * physic.KiloWattHour)
```
NeuralSpaz: Probably should add examples for the new constants, as they are great for increasing visibility.

maruel: Added one for WattSecond.

PR #333, opened 2018-11-11
physic: Add Frequency RPM

maruel:


conn/physic/units.go
```diff
@@ -388,6 +388,9 @@ const (
 	GigaHertz  Frequency = 1000 * MegaHertz
 	TeraHertz  Frequency = 1000 * GigaHertz
 
+	// RPM is revolutions per minute. It is used to quantify angular frequency.
+	RPM Frequency = 16667 * MicroHertz
```
NeuralSpaz: Since on this can't be exact and while I don't think it will be a problem 99% of the time. I'm just thinking that the should be a doc note to the notion "that for exact rates to use Hz", As unidirectional errors have a tendency to accumulate over time.

maruel: That's a fair point, but the error is 0.006% so I don't think it's worth mentioning. The constant itself is shown in godoc.

PR #337, opened 2018-11-12
physic: Add flag.Value implementation for Distance

NeuralSpaz:
Continuation of #303 
Add max/min Distance consts
Add Tests
Add Examples
Add Benchmarks for Distance and ElectricalCurrent Set() for comparison.
```
BenchmarkElectricCurrentSet-4      	10000000	       211 ns/op	       8 B/op	       2 allocs/op
BenchmarkDistanceSet-4				10000000	       176 ns/op	       6 B/op	       2 allocs/op
```
This looks like a bit more complicated, but its not its just unrolled a little with and extra step in the middle for mm and Mm etc, and one at the end for the conversions.

### DIFF
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -81,6 +81,117 @@ func (d Distance) String() string {
 	return nanoAsString(int64(d)) + "m"
 }
 
+// Set sets the Distance to the value represented by s. Units are to
+// be provided in "Metres", "Metre", "Miles", "Mile", "Yards", "Yard", "Inches"
+// or "Inch" with an optional SI prefix: "p", "n", "u", "µ", "m", "k", "M", "G"
+// or "T".
+func (d *Distance) Set(s string) error {
+	decimal, n, err := atod(s)
+	if err != nil {
+		if e, ok := err.(*parseError); ok {
+			switch e.err {
+			case errNotANumber:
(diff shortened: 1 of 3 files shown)


conn/physic/units.go
```diff
@@ -81,6 +81,117 @@ func (d Distance) String() string {
...
+			return &parseError{
+				err: errors.New("unexpected end of string"),
+			}
+		}
+		var siSize int
+		si, siSize = parseSIPrefix(r)
+		if si == milli || si == mega {
+			switch strings.ToLower(s[n:]) {
+			case "mile", "metre", "miles", "metres", "m":
+				si = unit
```
maruel: That's awkward, but I don't see a better way to do this.

conn/physic/units_test.go
```diff
@@ -2215,3 +2407,29 @@ func BenchmarkString2Decimal2IntNeg(b *testing.B) {
 	b.StopTimer()
 	_ = fmt.Sprintf("%d %d", v, n)
 }
+
+func BenchmarkElectricCurrentSet(b *testing.B) {
+	var err error
+	var d Distance
+	for i := 0; i < b.N; i++ {
+		err = d.Set("1Foot")
```
maruel: you can inline this;
```
if err = d.Set("1Foot"); err != nil {
  b.Fatal(err)
}
```

Use b.Fatal instead of log.Fatal.

maruel:
gohci

PR #338, opened 2018-11-12
fs,sysfs: Fix #323

### LINKED ISSUE #323: sysfs-gpio: assert WaitForEdge() and Halt() work as expected
- Add gpiosmoketest to assert that WaitForEdge() can be unblock by Halt().
- Make it live, #322.
- Document this in package gpio.

mame82:
Another possible solution to fix #323.

Notes:
- observations and testing have been done on a Rapsberry Pi Zero, with a 4.14.71 kernel (custom)
- instead of using epoll_ctl to add an epoll event for the GPIO sysfs value file only once, the code has been modified to delete the epoll event on every call to `Halt` and recreate it on a call to `In`
- before the epoll event gets deleted, it is modified to trigger on `EPOLLOUT`
- this modification ultimately fires an EPOLLOUT event itself and thus unlocks a pending `epoll_wait`. Unlocking the epoll_wait by this methhod could be distinguished from other methods, as the EPOLLOUT flag is set on the resulting event
- as the implementation of this library doesn't preserve EPOLL events, an event with `EPOLLOUT` flag set could be used as abort condition **only once** --> shouldn't work with concurrent calls to WaitForEdge

Another solution (based on loop interrupted by timeout) to force a pending `WaitForEdge` to return after a call to `Halt()` has been presented in the comments of issue #323.

The solution presented here has to be carefully tested. I have neither found any documentation of the observed behavior (calling EPOLL_CTL_MOD to set EPOLLOUT fires an event with the same flag set, for the same epoll fd), nor am I able to test on other hardware than Raspberry Pi0.

### DIFF
--- a/host/fs/fs.go
+++ b/host/fs/fs.go
@@ -97,6 +97,10 @@ func (e *Event) Wait(timeoutms int) (int, error) {
 	return e.event.wait(timeoutms)
 }
 
+func (e *Event) Delete() error {
+	return e.event.deleteEvent()
+}
+
 //
 
 var (
--- a/host/fs/fs_linux.go
+++ b/host/fs/fs_linux.go
@@ -5,6 +5,9 @@
 package fs
 
 import "syscall"
+import "golang.org/x/sys/unix"
+import "time"
+import "errors"
 
 const isLinux = true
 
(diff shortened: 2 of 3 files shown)


host/fs/fs.go
```diff
@@ -97,6 +97,10 @@ func (e *Event) Wait(timeoutms int) (int, error) {
 	return e.event.wait(timeoutms)
 }
 
+func (e *Event) Delete() error {
```
maruel: I think Close() would be appropriate here.

host/fs/fs_linux.go
```diff
@@ -5,6 +5,9 @@
 package fs
 
 import "syscall"
+import "golang.org/x/sys/unix"
```
maruel: Please use `goimports` to update the imports automatically. It'll keep them ordered.
Do not use package unix; periph doesn't have any external imports.

host/fs/fs_linux.go
```diff
@@ -46,12 +49,48 @@ func (e *event) makeEvent(fd uintptr) error {
...
-	e.event[0].Events = epollPRI | epollET
+	e.event[0].Events = unix.EPOLLPRI | unix.EPOLLET
 	e.event[0].Fd = int32(e.fd)
 	return syscall.EpollCtl(e.epollFd, epollCTLAdd, e.fd, &e.event[0])
 }
 
 func (e *event) wait(timeoutms int) (int, error) {
 	// http://man7.org/linux/man-pages/man2/epoll_wait.2.html
-	return syscall.EpollWait(e.epollFd, e.event[:], timeoutms)
+	n,err := syscall.EpollWait(e.epollFd, e.event[:], timeoutms)
```
maruel: Please run `gofmt -s`

host/sysfs/gpio.go
```diff
@@ -365,9 +365,17 @@ func (p *Pin) haltEdge() error {
 		if err := seekWrite(p.fEdge, bNone); err != nil {
 			return p.wrap(err)
 		}
+
+		// delete epoll event
+		err := p.event.Delete()
+		if err == nil {
+			p.fEdge = nil //indicate deletion of epoll event to force recreate on call to In()
```
maruel: That's a leak though.

host/sysfs/gpio.go
```diff
@@ -365,9 +365,17 @@ func (p *Pin) haltEdge() error {
 		if err := seekWrite(p.fEdge, bNone); err != nil {
```
maruel: In hindsight, I wonder if it's a good idea to reset the edge. Maybe we should have a separate function after all...

Basically, I'd like the following to work:
- In()
- WaitForEdge()
- Foo() (from another goroutine)
- WaitForEdge()

Where I had initially imagined that Foo() would be Halt(), but I'm less sure about it now.

host/sysfs/gpio.go
mame82: Close on fd before setting to nil

host/sysfs/gpio.go
mame82: Not clear to me: 

```
In()
WaitForEdge()
go Foo()
WaitForEdge()
```

should work, while 

```
In()
WaitForEdge()
go func() {
  Foo()
  WaitForEdge()
}
```

wouldn't work, unless WaitForEdge is teansformed to a wrapper, calling `wait` internally ONLY ONCE.

Could be achieved with a channel for example: wait is only called if the channel already exists, all succesive calls to WaitForEdge try to read from the channel and when wait returns the channel is closed. This is the simplest approach I could imagine for a "fan out" or "NotifyAll".

mame82:
Sorry for not cleaning up the PR, as already mentioned I'm currently missing the time to do more work on this, but we ended up with a bunch of possible solutions (loop polling/ use EPOLLOUT to indicate abort / use signals to interrupt epoll_wait / toggle pull resistors to return from epoll_wait).

Finding the best solution requires testing, for which I neither have time, nor resources.

Very sorry for this

host/fs/fs.go
mame82: Agree

host/fs/fs_linux.go
maruel: yes that's what the other constants are.

maruel:
Thanks for the ideas. I found a workaround that I pushed in branch new_epoll:
- I use one dedicated goroutine for the epoll_wait loop
- I wake it up with a dummy pipe as needed
- I use channels with timestamp for silencing stale events

I'll try to get this live in the coming weeks (I'm travelling a lot) and the fix should be in v3.5.0.

PR #339, opened 2018-11-13
physic: non exported change: decimal type internal representation.

NeuralSpaz:
Change decimal type internal representation of digits from []byte to uint64. 
Moves validation of digits to atod from doti.
Update tests for new decimal literal.
Add new test cases to atod().
Update Decimal Benchmarks.

This is a ~2x faster atod() and ~5x faster dtoi(), and helps with #303 remaining units.

### DIFF
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -1515,7 +1515,7 @@ func atod(s string) (decimal, int, error) {
 		// Find the exponet based on decimal point distance from left and the
 		// length of the number.
 		d.exp = (dp - start) - (end - start)
-		if (dp - start) <= 0 {
+		if dp <= start {
 			// Account for numbers of the form 1 > n < -1 eg 0.0001.
 			d.exp++
 		}
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -1479,14 +1479,13 @@ func atod(s string) (decimal, int, error) {
 	}
 
 	for i := start; i < end; i++ {
-
+		c := s[i]
 		// Check that is is a digit.
-		if s[i] >= '0' && s[i] <= '9' {
-			// '0' = 0x30 '1' = 0x31 ...etc.
-			digit := s[i] - '0'
+		if c >= '0' && c <= '9' {
 			// *10 is decimal shift left.
 			d.base *= 10
-			check := d.base + uint64(digit)
+			// Convert ascii digit into number.
+			check := d.base + uint64(c-'0')
 			// Check should always be larger than u unless we have overflowed.
 			// Similarly if check > max it will overflow when converted to int64.
 			if check < d.base || check > maxInt64 {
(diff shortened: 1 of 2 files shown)


NeuralSpaz:
gohci

conn/physic/units.go
```diff
@@ -1389,10 +1364,11 @@ func dtoi(d decimal, scale int) (int64, error) {
 }
 
 // Converts a string to a decimal form. The return int is how many bytes of the
-// string are numeric. The string may contain +-0 prefixes and arbitrary
-// suffixes as trailing non number characters are ignored.
-// Significant digits are stored without leading or trailing zeros, rather an
-// exponent is used.
+// string are considered numeric. The string may contain +-0 prefixes and
+// arbitrary suffixes as trailing non number characters are ignored.
+// Significant digits are stored without leading or trailing zeros, rather a
```
maruel: I don't understand what you mean with the first part, as it's stored as an int64 now.

conn/physic/units.go
```diff
@@ -1502,21 +1478,45 @@ func atod(s string) (decimal, int, error) {
 		}
 	}
 
-	if dp > start && dp < end {
-		// Concatenate with out decimal point.
-		d.digits = s[start:dp] + s[dp+1:end]
-	} else {
-		d.digits = s[start:end]
+	for i := start; i < end; i++ {
+
```
maruel: `c := s[i]` and use this instead of continuously indexing.

conn/physic/units.go
```diff
@@ -1502,21 +1478,45 @@ func atod(s string) (decimal, int, error) {
...
+						err: errOverflowsInt64Negative,
+					}
+				}
+				return decimal{}, 0, &parseError{
+					msg: maxUint64Str,
+					err: errOverflowsInt64,
+				}
+			}
+			d.base = check
+		} else {
```
maruel: } else if c != '.' {

conn/physic/units.go
```diff
@@ -1502,21 +1478,45 @@ func atod(s string) (decimal, int, error) {
...
-		d.digits = s[start:end]
+	for i := start; i < end; i++ {
+
+		// Check that is is a digit.
+		if s[i] >= '0' && s[i] <= '9' {
+			// '0' = 0x30 '1' = 0x31 ...etc.
+			digit := s[i] - '0'
+			// *10 is decimal shift left.
+			d.base *= 10
+			check := d.base + uint64(digit)
```
```
// Convert ascii digit into number.
check := d.base + uint64(c - '0')
```

conn/physic/units.go
```diff
@@ -1502,21 +1478,45 @@ func atod(s string) (decimal, int, error) {
...
 	if !isPoint {
 		d.exp = exp
 	} else {
-		ttl := dp - start
-		length := len(d.digits)
-		if ttl > 0 {
-			d.exp = ttl - length
-		} else {
-			d.exp = ttl - length + 1
+		if dp > start && dp < end {
```
maruel: I don't understand this condition. Can you add a comment?

conn/physic/units.go
```diff
@@ -1502,21 +1478,45 @@ func atod(s string) (decimal, int, error) {
...
+			end--
+		}
+		d.exp = (dp - start) - (end - start) + 1
+		if (dp - start) > 0 {
```
maruel: `if dp > start {`
Please comment too.

maruel:
gohci

conn/physic/units.go
NeuralSpaz: I'm trying to say 1000 is stored as base:1, exp:3.
0.001 stored as base: 1, exp: -3, 
and 12.34 as base: 1234 exp: -2

NeuralSpaz: ... and that there is a limit to the number of significant digits (int64) eg you can't have 1234567890123456790.1234567890123456789.
But you can store very large or very small numbers 16* 10^-34 (0.00000000000000000000000000000000016) etc

maruel: ahh now I understand.

conn/physic/units.go
```diff
@@ -1502,21 +1478,46 @@ func atod(s string) (decimal, int, error) {
...
-		} else {
-			d.exp = ttl - length + 1
+		if dp > start && dp < end {
+			// Decimal Point is in the middle of a number.
+			end--
+		}
+		// Find the exponet based on decimal point distance from left and the
+		// length of the number.
+		d.exp = (dp - start) - (end - start)
+		if (dp - start) <= 0 {
```
maruel: dp <= start

NeuralSpaz: oh good one.

PR #340, opened 2018-11-17
physic: decimal overflow fix.

NeuralSpaz:
Fixes bug in overflow detection in dtoi() when the exponent is large enough that overflow check is passed when it has overflowed and wrapped passed base.
Add fail test case for over overflow.


maruel:
gohci

maruel:
(Ignoring the as7262 failure)

PR #341, opened 2018-11-17
physic: performance improvements

NeuralSpaz:
Removed strings.ToLower() from units switches as this was causing an extra allocation. 
This change is ~ 50% faster.

### DIFF
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -91,7 +91,7 @@ func (d *Distance) Set(s string) error {
 		if e, ok := err.(*parseError); ok {
 			switch e.err {
 			case errNotANumber:
-				if found, _ := containsUnitString(s[n:], "Metre", "Metre", "Inch", "Foot", "Yard", "Mile", "m"); found != "" {
+				if found, _ := containsUnitString(s[n:], "Metre", "Inch", "Foot", "Yard", "Mile", "m"); found != "" {
 					return errors.New("does not contain number")
 				}
 				return errors.New("does not contain number or unit \"Metre\"")
@@ -115,8 +115,8 @@ func (d *Distance) Set(s string) error {
 		var siSize int
 		si, siSize = parseSIPrefix(r)
 		if si == milli || si == mega {
-			switch strings.ToLower(s[n:]) {
-			case "mile", "metre", "miles", "metres", "m":
+			switch s[n:] {
+			case "mile", "metre", "miles", "metres", "m", "Mile", "Metre", "Miles", "Metres":
 				si = unit
 			}
 		}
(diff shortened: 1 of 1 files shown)


maruel:
gohci

maruel:
I have a question though, do we really need to support both the acronym and the full form for every units? i.e. just supporting "m" and not "meter" and "meters" would probably be fine.

maruel:
> I have a question though, do we really need to support both the acronym and the full form for every units? i.e. just supporting "m" and not "meter" and "meters" would probably be fine.

time.ParseDuration() is a precedent, it supports "1m", not "1min" or "1 minute"

NeuralSpaz:
no not at all, I guess the only reason is for user experience, it would be simpler not to support longer forms.

maruel:
> no not at all, I guess the only reason is for user experience, it would be simpler not to support longer forms.

Less supported forms is less complex, so I think we should probably only support official suffixes, except for ambivalent ones like "miles".

conn/physic/units.go
```diff
@@ -82,19 +82,18 @@ func (d Distance) String() string {
...
+// be provided in "m", "Mile", "Yard", "in", or "ft" with an optional SI
+// prefix: "p", "n", "u", "µ", "m", "k", "M", "G" or "T".
 func (d *Distance) Set(s string) error {
 	decimal, n, err := atod(s)
 	if err != nil {
 		if e, ok := err.(*parseError); ok {
 			switch e.err {
 			case errNotANumber:
-				if found, _ := containsUnitString(s[n:], "Metre", "Metre", "Inch", "Foot", "Yard", "Mile", "m"); found != "" {
+				if found, _ := containsUnitString(s[n:], "in", "ft", "Yard", "Mile", "m"); found != "" {
```
maruel: Put "m" first, since it's the most sensible unit.

conn/physic/units.go
```diff
@@ -115,8 +114,8 @@ func (d *Distance) Set(s string) error {
 		var siSize int
 		si, siSize = parseSIPrefix(r)
 		if si == milli || si == mega {
-			switch strings.ToLower(s[n:]) {
-			case "mile", "metre", "miles", "metres", "m":
+			switch s[n:] {
+			case "mile", "m", "Mile":
```
maruel: put "m" first for consistency

conn/physic/units.go
```diff
@@ -139,57 +138,57 @@ func (d *Distance) Set(s string) error {
...
 		}
-	case "m", "metre", "metres":
+	case "m":
 		*d = (Distance)(v)
 	case "":
-		return noUnits("m, Metre, Mile, Inch, Foot or Yard")
+		return noUnits("m, Mile, in, ft or Yard")
 	default:
-		if found, extra := containsUnitString(s[n:], "Metre", "Metre", "Inch", "Foot", "Yard", "Mile", "m"); found != "" {
+		if found, extra := containsUnitString(s[n:], "in", "ft", "Yard", "Mile", "m"); found != "" {
```
maruel: "m" first

maruel:
gohci

PR #342, opened 2018-11-18
physic: minor patch for Angle.String(maxInt64)

NeuralSpaz:
Minor patch to a corner of the angle.String() for maximum value represented. 
Previous behavior would overflow int64 and give negative strings.

Helps with Angle.Set() PR wip.


maruel:
gohci

PR #343, opened 2018-11-18
physic: add flag.Value implementation for Force and Angle

NeuralSpaz:
This is an semi experimental method of maintaining dynamic range of units during conversion from one unit base to another using a lossy decimal multiply algorithm. The results of a decimal multiply can lossless but the maximum guaranteed precision is 18-degrade significant digits, This effectively will always give you at least 9 significant digits. The loss is always in the least significant bits.

As more least significant digits are discarded the processing time increases, with the worst case being roughly equivalent to conversion to float for calculation and back to int64 on amd64. float conversion is about %25 faster in worst case.

This PR helps with #303

### DIFF
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -62,18 +62,18 @@ func (a Angle) String() string {
 }
 
 // Set sets the Angle to the value represented by s. Units are to be provided in
-// "Radian", "Radians", "Rad" "Degree", "Degrees", "Deg" or "°" with an optional
-// SI prefix: "p", "n", "u", "µ", "m", "k", "M", "G" or "T".
+// "rad", "deg" or "°" with an optional SI prefix: "p", "n", "u", "µ", "m", "k",
+// "M", "G" or "T".
 func (f *Angle) Set(s string) error {
 	d, n, err := atod(s)
 	if err != nil {
 		if e, ok := err.(*parseError); ok {
 			switch e.err {
 			case errNotANumber:
-				if found, _ := containsUnitString(s[n:], "Radian", "Degree", "Rad", "Deg", "°"); found != "" {
+				if found, _ := containsUnitString(s[n:], "rad", "deg", "°"); found != "" {
 					return errors.New("does not contain number")
 				}
-				return errors.New("does not contain number or unit \"Radian\"")
+				return errors.New("does not contain number or unit \"Rad\"")
 			case errOverflowsInt64:
 				return errors.New("maximum value is " + maxAngle.String())
 			case errOverflowsInt64Negative:
@@ -97,16 +97,15 @@ func (f *Angle) Set(s string) error {
 	}
 
 	switch s[n:] {
-	case "degree", "degrees", "deg", "Degree", "Degrees", "Deg", "°":
+	case "deg", "°", "Deg":
 		degreePerRadian := decimal{
 			base: 17453293,
 			exp:  0,
 			neg:  false,
 		}
-		lbf, err := decimalMulScale(d, degreePerRadian, 7)
-		if err != nil {
-			return errors.New("converting to nano Radian would overflow, consider using nRad for maximum precision")
-		}
+		lbf, _ := decimalMulScale(d, degreePerRadian)
+		// Impossible for precision loss to exceed 9 since the number of
+		// significant figures in degrees per radian is only 8.
 		v, err := dtoi(lbf, int(si))
 		if err != nil {
 			if err != nil {
@@ -122,7 +121,7 @@ func (f *Angle) Set(s string) error {
 			}
 		}
 		*f = (Angle)(v)
-	case "radian", "radians", "rad", "Radian", "Radians", "Rad":
+	case "rad", "Rad":
 		v, err := dtoi(d, int(si-nano))
 		if err != nil {
 			if err != nil {
@@ -139,9 +138,9 @@ func (f *Angle) Set(s string) error {
 		}
 		*f = (Angle)(v)
 	case "":
-		return noUnits("Radian")
+		return noUnits("Rad")
 	default:
-		if found, extra := containsUnitString(s[n:], "Radian", "Degree", "Rad", "Deg", "°"); found != "" {
+		if found, extra := containsUnitString(s[n:], "Rad", "Deg", "°"); found != "" {
 			return unknownUnitPrefix(found, extra, "p,n,u,µ,m,k,M,G or T")
 		}
 		return incorrectUnit(s[n:], "physic.Angle")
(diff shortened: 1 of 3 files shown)


NeuralSpaz:
This is here for just for design consideration, and to see if anyone has a better idea.

conn/physic/units.go
```diff
@@ -51,12 +51,104 @@ func (a Angle) String() string {
...
+		return prefix + strconv.FormatInt(int64(v), 10) + "°"
 	default:
 		v := (a + Degree/2) / Degree
 		return prefix + strconv.FormatInt(int64(v), 10) + "°"
 	}
 }
 
+// Set sets the Angle to the value represented by s. Units are to be provided in
+// "Radian", "Radians", "Rad" "Degree", "Degrees", "Deg" or "°" with an optional
+// SI prefix: "p", "n", "u", "µ", "m", "k", "M", "G" or "T".
```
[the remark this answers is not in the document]
NeuralSpaz: yeah most people use decimal / sexagesimal, 0.5 arc seconds etc.

conn/physic/units.go
```diff
@@ -1552,6 +1740,56 @@ func valueOfUnitString(s string, base prefix) (int64, int, error) {
...
+			}
+		}
+		// fmt.Println(sa, sb)
+		exp++
+	}
+	// Exceeded the limit of precision degradation
+	return decimal{}, errBaseOverflow
+}
+
+var (
```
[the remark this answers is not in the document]
NeuralSpaz: Its only there to keep the code as a block for readability it would move to the other decimal errors var block.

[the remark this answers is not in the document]
maruel: ok no problem

conn/physic/units.go
```diff
@@ -51,12 +51,103 @@ func (a Angle) String() string {
 		i := v / 1000
 		v = v - i*1000
 		return prefix + strconv.FormatInt(int64(i), 10) + "." + prefixZeros(1, int(v)) + "°"
+	case a > (9223372036854775807 - 17453293):
```
maruel: Can you rebase on top of origin/master so this disappears? Thanks.

But I realized, 17453293 should be Degree instead so then the condition would be:
`case a > maxAngle - Degree:`
which would be much more readable.

conn/physic/units.go
```diff
@@ -51,12 +51,103 @@ func (a Angle) String() string {
...
+			base: 17453293,
+			exp:  0,
+			neg:  false,
+		}
+		lbf, _ := decimalMul(d, degreePerRadian)
+		// Impossible for precision loss to exceed 9 since the number of
+		// significant figures in degrees per radian is only 8.
+		v, err := dtoi(lbf, int(si))
+		if err != nil {
+			if err != nil {
```
maruel: ? :)

conn/physic/units.go
```diff
@@ -1552,6 +1739,50 @@ func valueOfUnitString(s string, base prefix) (int64, int, error) {
 	return v, n, nil
 }
 
+// Calcululates the product of two decimals; a and b, keeping the base less than
```
```
// decumalMul calculates the ...
```

Also, s/figures/digits/.

conn/physic/units.go
```diff
@@ -1552,6 +1739,50 @@ func valueOfUnitString(s string, base prefix) (int64, int, error) {
...
+// maxInt64. Degrade controls the maximum limit of the precision loss in the
+// product. Returns an error if the product base exceeds an int64 after losing a
+// degrade number of least significant figures. This function is to aid in the
+// multiplication of numbers that combined have more than 18 significant figures
+// each. The minimum limit of significant figures is 9 figures.
+func decimalMul(a, b decimal) (decimal, uint) {
+	if a.base > 18446744073709551609 || b.base > 18446744073709551609 {
```
maruel: Can you comment where this number comes from?
1<<64 is 18446744073709551616, not 18446744073709551609.
But in practice, anything higher than 1<<63 is game over.

conn/physic/units.go
```diff
@@ -67,6 +158,9 @@ const (
 	Theta  Angle = 6283185307 * NanoRadian
 	Pi     Angle = 3141592653 * NanoRadian
 	Degree Angle = 17453293 * NanoRadian
+
+	maxAngle = 9223372036854775807 * NanoRadian
```
maruel: I wonder if these should be rewritten as:
```
maxAngle = (1<<63)*NanoRadian
```
then in the test with:
- as passing, maxAngle-1
- as failing, maxAngle

conn/physic/units.go
```diff
@@ -443,6 +613,9 @@ const (
 	// Pound is both a unit of mass and weight (force). The suffix Force is added
 	// to disambiguate the measurement it represents.
 	PoundForce Force = 4448221615261 * NanoNewton
+
+	maxForce = 9223372036854775807 * NanoNewton
```
maruel: I'd recommend 1<<63 for these too

conn/physic/units.go
maruel: Can you add this a a comment?

conn/physic/units_test.go
```diff
@@ -35,8 +35,8 @@ func TestAngle_String(t *testing.T) {
 		{-1000 * Degree, "-1000°"},
 		{100000000000 * Degree, "100000000000°"},
 		{-100000000000 * Degree, "-100000000000°"},
-		{(9223372036854775807) * NanoRadian, "528460276055°"},
-		{(-9223372036854775807) * NanoRadian, "-528460276055°"},
+		{maxAngle * NanoRadian, "528460276055°"},
```
maruel: maxAngle is already in NanoRadian, so this is technically an error to multiply by NanoRadian.

conn/physic/units_test.go
```diff
@@ -857,6 +1019,166 @@ func TestValueOfUnitString(t *testing.T) {
...
+		{"1krad", 1000 * Radian},
+		{"10krad", 10000 * Radian},
+		{"100krad", 100000 * Radian},
+		{"1Mrad", 1000000 * Radian},
+		{"10Mrad", 10000000 * Radian},
+		{"100Mrad", 100000000 * Radian},
+		{"1Grad", 1000000000 * Radian},
+		{"12.345rad", 12345 * MilliRadian},
+		{"-12.345rad", -12345 * MilliRadian},
+		{"9.223372036854775807Grad", 9223372036854775807 * NanoRadian},
```
maruel: maxAngle?

conn/physic/units_test.go
```diff
@@ -857,6 +1019,166 @@ func TestValueOfUnitString(t *testing.T) {
...
+		{
+			"10eRadianE",
+			"contains unknown unit prefix \"e\". valid prefixes for \"Rad\" are p,n,u,µ,m,k,M,G or T",
+		},
+		{
+			"10",
+			"no units provided, need Rad",
+		},
+		{
+			"9223372036854775808rad",
```
maruel: fmt.Sprintf("%drad", maxAngle + 1) ?

conn/physic/units_test.go
```diff
@@ -2445,3 +2942,81 @@ func BenchmarkElectricCurrentSet(b *testing.B) {
 	b.StopTimer()
 	_ = fmt.Sprintf("%d", e)
 }
+
+func BenchmarkForceSetMetric(b *testing.B) {
+	var err error
+	var f Force
+	for i := 0; i < b.N; i++ {
+		err = f.Set("123N")
```
maruel: you can inline these as:

```
if err = f.Set("123N"); err != nil {
```

conn/physic/units_test.go
NeuralSpaz: could do `fmt.Sprintf("%drad", uint64(maxAngle)+1)`?

maruel: In tests it's fine to use fmt.

NeuralSpaz:
I'd forgot to add examples, Checking them noticed PoundForce was out. Which brings up another choice. We can keep the precision of the conversion (effective picoNewton) or can make const PoundForce law.
```
"20lbf" == 88964432305 nN (accurate)
20* PoundForce == 88964432300nN (self consistent)
```
I don't really care either way, but I think it should be a decision.

PR #348, opened 2018-11-26
sn3218 - add experimental support for sn3218 LED driver

lnitram:
Add experimental device for SN3218 LED driver.

### DIFF
--- a/experimental/cmd/sn3218/main.go
+++ b/experimental/cmd/sn3218/main.go
@@ -20,23 +20,24 @@ func main() {
 	}
 
 	b, err := i2creg.Open("")
-
+	if err != nil {
+		log.Fatal(err)
+	}
 	defer b.Close()
 
+	d, err := sn3218.New(b)
 	if err != nil {
 		log.Fatal(err)
 	}
-
-	d, err := sn3218.New(b)
-
 	defer d.Halt()
 
-	if err != nil {
-		log.Fatal(err)
+	if err := d.Enable(); err != nil {
+		log.Fatal("Error while enabling device", err)
 	}
 
-	d.Enable()
-	d.SetGlobalBrightness(1)
+	if err := d.SetGlobalBrightness(1); err != nil {
+		log.Fatal("Error while setting brightness", err)
+	}
 
 	// Switch LED 7 on
 	if err := d.Switch(7, true); err != nil {
@@ -58,18 +59,26 @@ func main() {
 	log.Println("State: ", state, " - Brightness: ", brightness)
 
 	// Switch all LEDs on
-	d.SwitchAll(true)
+	if err := d.SwitchAll(true); err != nil {
+		log.Fatal("Error while switching all LEDs", err)
+	}
 	time.Sleep(1000 * time.Millisecond)
 
 	// Increase brightness for all
-	d.SetGlobalBrightness(125)
+	if err := d.SetGlobalBrightness(125); err != nil {
+		log.Fatal("Error while changing globalBrightness", err)
+	}
 	time.Sleep(1000 * time.Millisecond)
 
 	// Disable to save energy, but keep state
-	d.Disable()
+	if err := d.Disable(); err != nil {
+		log.Fatal("Error while disabling device")
+	}
 	time.Sleep(1000 * time.Millisecond)
 
 	// Enable again
-	d.Enable()
+	if err := d.Enable(); err != nil {
+		log.Fatal("Error while enabling device")
+	}
 	time.Sleep(1000 * time.Millisecond)
 }
--- a/experimental/cmd/sn3218/main.go
+++ b/experimental/cmd/sn3218/main.go
@@ -26,7 +26,12 @@ func main() {
 		log.Fatal(err)
 	}
 
-	l, _ := sn3218.New(b)
+	l, err := sn3218.New(b)
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	l.Reset()
 	l.Enable()
 
--- a/experimental/cmd/sn3218/main.go
+++ b/experimental/cmd/sn3218/main.go
@@ -74,5 +74,5 @@ func main() {
 	}
 
 	log.Println("Cleanup: Reset register and switch off")
-	l.Disable()
+	l.Halt()
 }
--- a/experimental/cmd/sn3218/main.go
+++ b/experimental/cmd/sn3218/main.go
@@ -35,7 +35,7 @@ func main() {
 		log.Fatal("Error while enabling device", err)
 	}
 
-	if err := d.SetGlobalBrightness(1); err != nil {
+	if err := d.BrightnessAll(1); err != nil {
 		log.Fatal("Error while setting brightness", err)
 	}
 
@@ -46,7 +46,7 @@ func main() {
 	time.Sleep(1000 * time.Millisecond)
 
 	//Increase brightness for LED 7 to max
-	if err := d.SetBrightness(7, 255); err != nil {
+	if err := d.Brightness(7, 255); err != nil {
 		log.Fatal("Error while changing LED brightness", err)
 	}
 	time.Sleep(1000 * time.Millisecond)
@@ -65,7 +65,7 @@ func main() {
 	time.Sleep(1000 * time.Millisecond)
 
 	// Increase brightness for all
-	if err := d.SetGlobalBrightness(125); err != nil {
+	if err := d.BrightnessAll(125); err != nil {
 		log.Fatal("Error while changing globalBrightness", err)
 	}
 	time.Sleep(1000 * time.Millisecond)
--- a/experimental/cmd/sn3218/main.go
+++ b/experimental/cmd/sn3218/main.go
@@ -32,7 +32,6 @@ func main() {
 		log.Fatal(err)
 	}
 
-	l.Reset()
 	l.Enable()
 
 	log.Println("Switch each LED on one by one")
@@ -75,6 +74,5 @@ func main() {
 	}
 
 	log.Println("Cleanup: Reset register and switch off")
-	l.Reset()
 	l.Disable()
 }
--- a/experimental/cmd/sn3218/main.go
+++ b/experimental/cmd/sn3218/main.go
@@ -39,7 +39,7 @@ func main() {
 	d.SetGlobalBrightness(1)
 
 	// Switch LED 7 on
-	if err := d.SwitchLed(7, true); err != nil {
+	if err := d.Switch(7, true); err != nil {
 		log.Fatal("Error while switching LED", err)
 	}
 	time.Sleep(1000 * time.Millisecond)
@@ -51,14 +51,14 @@ func main() {
 	time.Sleep(1000 * time.Millisecond)
 
 	//Get state of LED 7
-	state, brightness, err := d.GetLedState(7)
+	state, brightness, err := d.GetState(7)
 	if err != nil {
 		log.Fatal("Error while reading LED state", err)
 	}
 	log.Println("State: ", state, " - Brightness: ", brightness)
 
 	// Switch all LEDs on
-	d.SwitchAllLeds(true)
+	d.SwitchAll(true)
 	time.Sleep(1000 * time.Millisecond)
 
 	// Increase brightness for all
--- a/experimental/cmd/sn3218/main.go
+++ b/experimental/cmd/sn3218/main.go
@@ -31,7 +31,7 @@ func main() {
 	}
 	defer d.Halt()
 
-	if err := d.Enable(); err != nil {
+	if err := d.WakeUp(); err != nil {
 		log.Fatal("Error while enabling device", err)
 	}
 
@@ -70,14 +70,14 @@ func main() {
 	}
 	time.Sleep(1000 * time.Millisecond)
 
-	// Disable to save energy, but keep state
-	if err := d.Disable(); err != nil {
+	// Sleep mode to save energy, but keep state
+	if err := d.Sleep(); err != nil {
 		log.Fatal("Error while disabling device")
 	}
 	time.Sleep(1000 * time.Millisecond)
 
-	// Enable again
-	if err := d.Enable(); err != nil {
+	// WakeUp again
+	if err := d.WakeUp(); err != nil {
 		log.Fatal("Error while enabling device")
 	}
 	time.Sleep(1000 * time.Millisecond)
--- a/experimental/cmd/sn3218/main.go
+++ b/experimental/cmd/sn3218/main.go
@@ -14,6 +14,7 @@ import (
 )
 
 func main() {
+
 	if _, err := host.Init(); err != nil {
 		log.Fatal(err)
 	}
@@ -26,53 +27,49 @@ func main() {
 		log.Fatal(err)
 	}
 
-	l, err := sn3218.New(b)
+	d, err := sn3218.New(b)
+
+	defer d.Halt()
 
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	l.Enable()
+	d.Enable()
+	d.SetGlobalBrightness(1)
 
-	log.Println("Switch each LED on one by one")
-	l.SetGlobalBrightness(1)
-	for i := 0; i < 18; i++ {
-		err := l.SwitchLed(i, true)
-		if err != nil {
-			log.Fatal(err)
-		}
-		time.Sleep(100 * time.Millisecond)
+	// Switch LED 7 on
+	if err := d.SwitchLed(7, true); err != nil {
+		log.Fatal("Error while switching LED", err)
 	}
+	time.Sleep(1000 * time.Millisecond)
 
-	log.Println("Set brightness of each LED to 10 one by one")
-	for i := 0; i < 18; i++ {
-		l.SetBrightness(i, 10)
-		time.Sleep(100 * time.Millisecond)
+	//Increase brightness for LED 7 to max
+	if err := d.SetBrightness(7, 255); err != nil {
+		log.Fatal("Error while changing LED brightness", err)
 	}
+	time.Sleep(1000 * time.Millisecond)
 
-	log.Println("Change between even and odd LEDs a couple of times")
-	l.SetGlobalBrightness(1)
-	time.Sleep(100 * time.Millisecond)
-	l.SetGlobalBrightness(1)
-	for x := 0; x < 10; x++ {
-		for i := 0; i < 18; i++ {
-			l.SwitchLed(i, i%2 == 0)
-		}
-		time.Sleep(100 * time.Millisecond)
-		for i := 0; i < 18; i++ {
-			l.SwitchLed(i, i%2 == 1)
-		}
-		time.Sleep(100 * time.Millisecond)
+	//Get state of LED 7
+	state, brightness, err := d.GetLedState(7)
+	if err != nil {
+		log.Fatal("Error while reading LED state", err)
 	}
+	log.Println("State: ", state, " - Brightness: ", brightness)
 
-	log.Println("And now all on and off together")
-	state := true
-	for i := 0; i < 10; i++ {
-		state = !state
-		l.SwitchAllLeds(state)
-		time.Sleep(100 * time.Millisecond)
-	}
+	// Switch all LEDs on
+	d.SwitchAllLeds(true)
+	time.Sleep(1000 * time.Millisecond)
 
-	log.Println("Cleanup: Reset register and switch off")
-	l.Halt()
+	// Increase brightness for all
+	d.SetGlobalBrightness(125)
+	time.Sleep(1000 * time.Millisecond)
+
+	// Disable to save energy, but keep state
+	d.Disable()
+	time.Sleep(1000 * time.Millisecond)
+
+	// Enable again
+	d.Enable()
+	time.Sleep(1000 * time.Millisecond)
 }
--- a/experimental/cmd/sn3218/main.go
+++ b/experimental/cmd/sn3218/main.go
@@ -1,0 +1,71 @@
+package main
+
+import (
+	"log"
+	"time"
+
+	"periph.io/x/periph/conn/i2c/i2creg"
+	"periph.io/x/periph/experimental/devices/sn3218"
+	"periph.io/x/periph/host"
+)
+
+func main() {
+	if _, err := host.Init(); err != nil {
+		log.Fatal(err)
+	}
+
+	b, err := i2creg.Open("")
+
+	defer b.Close()
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	l, _ := sn3218.New(b)
+	l.Reset()
+	l.Enable()
+
+	log.Println("Switch each LED on one by one")
+	l.SetGlobalBrightness(1)
+	for i := 0; i < 18; i++ {
+		err := l.SwitchLed(i, true)
+		if err != nil {
+			log.Fatal(err)
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	log.Println("Set brightness of each LED to 10 one by one")
+	for i := 0; i < 18; i++ {
+		l.SetBrightness(i, 10)
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	log.Println("Change between even and odd LEDs a couple of times")
+	l.SetGlobalBrightness(1)
+	time.Sleep(100 * time.Millisecond)
+	l.SetGlobalBrightness(1)
+	for x := 0; x < 10; x++ {
+		for i := 0; i < 18; i++ {
+			l.SwitchLed(i, i%2 == 0)
+		}
+		time.Sleep(100 * time.Millisecond)
+		for i := 0; i < 18; i++ {
+			l.SwitchLed(i, i%2 == 1)
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	log.Println("And now all on and off together")
+	state := true
+	for i := 0; i < 10; i++ {
+		state = !state
+		l.SwitchAllLeds(state)
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	log.Println("Cleanup: Reset register and switch off")
+	l.Reset()
+	l.Disable()
+}
--- a/experimental/cmd/sn3218/main.go
+++ b/experimental/cmd/sn3218/main.go
@@ -1,3 +1,7 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
 package main
 
 import (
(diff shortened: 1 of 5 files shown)


experimental/cmd/sn3218/main.go
```diff
@@ -0,0 +1,71 @@
...
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	l, _ := sn3218.New(b)
+	l.Reset()
+	l.Enable()
+
+	log.Println("Switch each LED on one by one")
```
maruel: This is more of a smoke test. The executable should provide a simple way to control the LEDs.

experimental/devices/sn3218/doc.go
```diff
@@ -0,0 +1,11 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
```
maruel: Use the current year (in all files).

experimental/devices/sn3218/doc.go
```diff
@@ -0,0 +1,11 @@
...
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+//
+// Package sn3218 controls a SN3218 LED driver with 18 LEDs over an i2c bus.
+// See cmd/sn3218/main.go for a usage example. Make sure to run Reset() after
```
maruel: Do not add this, include a simple example_test.go instead.

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,136 @@
...
+	i2cAddress             = 0x54
+	cmdEnableOutput        = 0x00
+	cmdSetBrightnessValues = 0x01
+	cmdEnableLeds          = 0x13
+	cmdUpdate              = 0x16
+	cmdReset               = 0x17
+)
+
+type dev struct {
+	i2c        *i2c.Dev
```
maruel: This doesn't need to be a pointer, you can use `d i2c.Dev`

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,136 @@
...
+	states     [18]bool
+	brightness [18]byte
+}
+
+// New returns an object that handles communication with a SN3218 LED driver
```
maruel: // New returns an handle to a SN3218 LED driver.

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,136 @@
...
+// The SN3213 only accepts writes, so this object also keeps the state of
```
maruel: This is an implementation detail that you can comment in this file, but it's not worth surfacing in the documentation, the user likely doesn't mind about this.

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,136 @@
...
+// each LED.
+// Important: The device is write only, so it is not possible to read the state
+// of each LED. Therefore it's recommended to run d.Reseet() after this, so that
+// all LEDs are off and have the brightness 0.  Otherwise GetLedState() returns
+// wrong results.
+func New(bus i2c.Bus) (*dev, error) {
+	d := &i2c.Dev{Bus: bus, Addr: i2cAddress}
```
maruel: Since it's a write-only device, I'd vote to reset it. It's annoying in some cases, as some user may want to "keep the previous state as-is" but I'd prefer to err on the safer side here and reset automatically instead of having to comment everywhere that the user has to call Reset().

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,136 @@
...
+	dev := dev{}
+	dev.i2c = d
+	return &dev, nil
+}
+
+// Enable enables the SN3218
```
maruel: What does enable means?

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,136 @@
...
+	return err
+}
+
+// Disable disables the SN3218
+func (d *dev) Disable() error {
+	_, err := d.i2c.Write([]byte{cmdEnableOutput, 0x00})
+	return err
+}
+
+// Reset resets the registers to the default values
```
maruel: I don't think you need to expose this function, inline this code inside New().

experimental/devices/sn3218/doc.go
```diff
@@ -0,0 +1,11 @@
...
+// New(), so that the state of the LEDs match the register of the chip.
+//
+// Datasheet
+// https://github.com/pimoroni/sn3218/blob/master/datasheets/sn3218-datasheet.pdf
```
maruel: Please start `godoc -http=:6060` locally and browse to http://localhost:6060/pkg/periph.io/x/periph/experimental/devices/sn3218.

Then include an empty `//` in between the line Datasheet and the URL and refresh the page. You'll see how Datasheet becomes a title.

The right URL is http://www.si-en.com/uploadpdf/s2011517171720.pdf

experimental/devices/sn3218/doc.go
```diff
@@ -0,0 +1,11 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+//
+// Package sn3218 controls a SN3218 LED driver with 18 LEDs over an i2c bus.
```
maruel: In the end of the day, the sn3218 is really just 18x PWMs with an unspecified(?) frequency. So while the datasheet calls out LEDs use case, people can use it for anything. So I'd prefer not abuse the word "LED" in the function names below.

experimental/devices/sn3218/sn3218_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+func setup() *i2ctest.Record {
+	bus := i2ctest.Record{
+		Bus: nil,
+		Ops: []i2ctest.IO{},
+	}
+	return &bus
+}
+
+func equal(slice1 []byte, slice2 []byte) bool {
```
maruel: https://golang.org/pkg/bytes/#Equal

experimental/devices/sn3218/sn3218_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+	for i := range slice1 {
+		if slice1[i] != slice2[i] {
+			return false
+		}
+	}
+	return true
+}
+
+// func New(bus i2c.Bus) (*dev, error) {
```
maruel: Remove

experimental/cmd/sn3218/main.go
```diff
@@ -0,0 +1,71 @@
+package main
```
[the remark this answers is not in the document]
lnitram: done

experimental/devices/sn3218/doc.go
lnitram: done

experimental/devices/sn3218/sn3218.go
lnitram: You can enable and disable the driver. It keeps the state when you disable it. According to the datasheet you can use this to save energy and switch on and off the LEDs only when needed.

maruel: Enabling could be implicit: it's reenabled as soon as _anything_ is reactivated, and disabled when everything has been turned off (for example via `Halt()`). This limits the API surface.

Basically when sending the shutdown command, also set the other registers to stop everything. This reduces the state management.

maruel: One interesting thing is that if you write 00 to register 0x14 (reset), it also sets shutdown mode. I think that does exactly what is desired. Then when you reactivate you set whatever register to the PWM enabled and then write 1 to register 0x00 (shutdown)

experimental/devices/sn3218/sn3218.go
lnitram: done

experimental/cmd/sn3218/main.go
```diff
@@ -0,0 +1,71 @@
...
+
+	b, err := i2creg.Open("")
+
+	defer b.Close()
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	l, _ := sn3218.New(b)
```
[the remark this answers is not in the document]
lnitram: done

experimental/devices/sn3218/sn3218.go
lnitram: done

experimental/devices/sn3218/sn3218.go
lnitram: done

experimental/devices/sn3218/sn3218_test.go
lnitram: done

experimental/devices/sn3218/sn3218_test.go
lnitram: done

experimental/devices/sn3218/sn3218.go
lnitram: done

experimental/cmd/sn3218/main.go
lnitram: I simplified the main.go 
It's now just showing one call of each public function with a short comment. As it's a write only device it's a bit unclear how to build something that "has purpose" (maybe just blink 42 times ;-))

experimental/devices/sn3218/doc.go
lnitram: done

experimental/devices/sn3218/sn3218.go
lnitram: done

experimental/devices/sn3218/doc.go
lnitram: removed LED from the functions and renamed function parameters from led to channel

experimental/cmd/sn3218/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+	d, err := sn3218.New(b)
+
+	defer d.Halt()
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	d.Enable()
```
maruel: Still check for errors everywhere, even if tedious.

experimental/devices/sn3218/example_test.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package sn3218_test
+
+import (
+	"log"
```
maruel: Please use `goimports -w`. I recommend to setup your text editor to invoke it automatically upon file save.

http://golang.org/x/tools/cmd/goimports

experimental/devices/sn3218/example_test.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func Example() {
+	if _, err := host.Init(); err != nil {
+		log.Fatal(err)
+	}
+
+	b, err := i2creg.Open("")
+
+	defer b.Close()
```
maruel: This ordering would panic if err != nil since b == nil

experimental/devices/sn3218/example_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	d, err := sn3218.New(b)
+
+	// to ensure that all registers are reset and all LEDs are switched off
+	defer d.Halt()
```
maruel: Same

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+type Dev struct {
+	i2c        i2c.Dev
+	states     [18]bool
+	brightness [18]byte
+}
+
+// New returns a handle to a SN3218 LED driver.
+func New(bus i2c.Bus) (Dev, error) {
+	d := i2c.Dev{Bus: bus, Addr: i2cAddress}
+	dev := Dev{}
```
maruel:
```
d := &Dev{
  i2c: i2c.Dev{Bus: bus, Addr: i2cAddress},
}
if err := d.reset(); err != nil {
  return nil, err
}
return d, nil
```

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+	return d.reset()
+}
+
+// Enable enables the SN3218.
+func (d *Dev) Enable() error {
+	_, err := d.i2c.Write([]byte{cmdEnableOutput, 0x01})
+	return err
+}
+
+// Disable disables the SN3218.
```
maruel: It's kinda unclear to the user what 'disable' means in this context.

Sleep / WakeUp would probably be clearer.

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+func (d *Dev) Disable() error {
+	_, err := d.i2c.Write([]byte{cmdEnableOutput, 0x00})
+	return err
+}
+
+// GetState returns the state (on/off) and the brightness (0..255) of the
+// Channel 0..17.
+func (d *Dev) GetState(channel int) (bool, byte, error) {
+	if channel < 0 || channel >= 18 {
+		return false, 0, errors.New("Channel number out of range 0..17")
```
maruel: start the error with a lower case.

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return d.states[channel], d.brightness[channel], nil
+}
+
+// Switch switched the channel (0..18) to state (on/off).
```
maruel: I'm not sure about the word 'Switch' but currently do not have a good suggestion.

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+func (d *Dev) Switch(channel int, state bool) error {
+	if channel < 0 || channel >= 18 {
+		return errors.New("Channel number out of range 0..17")
+	}
+	d.states[channel] = state
+	return d.updateStates()
+}
+
+// SetGlobalBrightness sets the brightness of all channels to the value (0..255).
+func (d *Dev) SetGlobalBrightness(value byte) {
```
maruel: return error

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+func (d *Dev) SetBrightness(channel int, value byte) error {
+	if channel < 0 || channel >= 18 {
+		return errors.New("Chhannel number out of range 0..17")
+	}
+	d.brightness[channel] = value
+	d.updateBrightness()
+	return nil
+}
+
+// SwitchAll switches all channels accoring to the state (on/off).
```
maruel: according
move it close to Switch

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+// SetGlobalBrightness sets the brightness of all channels to the value (0..255).
+func (d *Dev) SetGlobalBrightness(value byte) {
+	for i := 0; i < 18; i++ {
+		d.brightness[i] = value
+	}
+	d.updateBrightness()
+}
+
+// SetBrightness sets the brightness of led (0..17) to value (0..255).
+func (d *Dev) SetBrightness(channel int, value byte) error {
```
maruel: If you use Switch/SwitchAll, should you use similar naming for Brightness?

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Reset resets the registers to the default values.
+func (d *Dev) reset() error {
+	_, err := d.i2c.Write([]byte{cmdReset, 0xFF})
+	d.states = [18]bool{}
+	d.brightness = [18]byte{}
+	return err
+}
+
+func boolArrayToInt(states [18]bool) uint {
```
maruel: That's actually copying 18 bytes.

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+	cmdSetBrightnessValues = 0x01
+	cmdEnableLeds          = 0x13
+	cmdUpdate              = 0x16
+	cmdReset               = 0x17
+)
+
+// Dev is a handler to sn3218 controller.
+type Dev struct {
+	i2c        i2c.Dev
+	states     [18]bool
```
maruel: s/states/on/ ?

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+	for i := 0; i < 18; i++ {
+		d.states[i] = state
+	}
+	d.updateStates()
+}
+
+// Reset resets the registers to the default values.
+func (d *Dev) reset() error {
+	_, err := d.i2c.Write([]byte{cmdReset, 0xFF})
+	d.states = [18]bool{}
```
maruel: I wonder if the compiler is able to optimize that, vs a loop.

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+	d := i2c.Dev{Bus: bus, Addr: i2cAddress}
+	dev := Dev{}
+	dev.i2c = d
+	dev.reset()
+	return dev, nil
+}
+
+// Halt resets the registers and switches the driver off.
+func (d *Dev) Halt() error {
+	if err := d.Disable(); err != nil {
```
maruel: This is not necessary, reset() puts the device in power save mode.

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+	cmdEnableOutput        = 0x00
+	cmdSetBrightnessValues = 0x01
+	cmdEnableLeds          = 0x13
+	cmdUpdate              = 0x16
+	cmdReset               = 0x17
+)
+
+// Dev is a handler to sn3218 controller.
+type Dev struct {
+	i2c        i2c.Dev
```
maruel: you need to track if the device is in sleep mode or now, and transparently wake it up (Enable) on any call enabling any LED.

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (d *Dev) update() error {
+	_, err := d.i2c.Write([]byte{cmdUpdate, 0xFF})
+	return err
+}
+
+func (d *Dev) updateStates() error {
+	mask := boolArrayToInt(d.states)
+	_, err := d.i2c.Write([]byte{cmdEnableLeds, byte(mask & 0x3F), byte((mask >> 6) & 0x3F), byte((mask >> 12) & 0X3F)})
```
maruel:
```
cmd := [...]byte{cmdEnableLeds, byte(mask & 0x3F), byte((mask >> 6) & 0x3F), byte((mask >> 12) & 0X3F)}
if _, err := d.i2c.Write(cmd[:]); err != nil {
  return err
}
```

experimental/devices/sn3218/sn3218.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return err
+	}
+	return d.update()
+}
+
+func (d *Dev) updateBrightness() error {
+	_, err := d.i2c.Write(append([]byte{cmdSetBrightnessValues}, d.brightness[0:len(d.brightness)]...))
```
maruel: This causes memory allocation, instead do

```
cmd := [19]byte{cmdSetBrightnessValues}
copy(cmd[1:], d.brightness)
if _ err := d.i2c.Write(cmd[:]); err != nil {
...
```

experimental/devices/sn3218/sn3218_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"periph.io/x/periph/conn/i2c/i2ctest"
+)
+
+func setup() *i2ctest.Record {
+	bus := i2ctest.Record{
```
maruel: return &i2ctest.Record{

experimental/devices/sn3218/sn3218_test.go
```diff
@@ -0,0 +1,235 @@
...
+		Bus: nil,
+		Ops: []i2ctest.IO{},
+	}
+	return &bus
+}
+
+func TestNew(t *testing.T) {
+	bus := i2ctest.Record{
```
maruel: why not use the function just above?

experimental/devices/sn3218/sn3218_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+	if !bytes.Equal(bus.Ops[0].W, []byte{0x17, 0xFF}) {
+		t.Fatal("Expected: 0x17, 0xFF (reset), got: ", bus.Ops[0].W)
+	}
+}
+
+func TestHalt(t *testing.T) {
+	bus := setup()
+	dev, _ := New(bus)
+	err := dev.Halt()
```
maruel: Inline into next line.

experimental/devices/sn3218/sn3218_test.go
```diff
@@ -0,0 +1,235 @@
...
+func TestNew(t *testing.T) {
+	bus := i2ctest.Record{
+		Bus: nil,
+		Ops: []i2ctest.IO{},
+	}
+	_, err := New(&bus)
+	if err != nil {
+		t.Fatal("New should not return error", err)
+	}
+	if len(bus.Ops) > 1 {
```
maruel: In practice, you should use i2ctest.Playback in all of these, not Record.

experimental/devices/sn3218/sn3218.go
lnitram: Agree, will do, just see a little confusion with time.Sleep.

experimental/devices/sn3218/sn3218.go
lnitram: Idea was according to the sentence "Switch LED 8 on"  - Switch(8, true)
But I am happy to change it. Some ideas:
Out(channel uint8, state bool)
Output(channel uint8, state bool)
Set(channel uint8, state bool)

experimental/devices/sn3218/sn3218.go
lnitram: changed to
Switch / SwitchAll - Brightness / BrightnessAll
still not happy with the names, but no better idea atm

experimental/devices/sn3218/sn3218.go
lnitram: 100 mio x [18]bool{} - Took: 41.262059ms
100 mio x loop - Took: 821.613264ms

```
package main

import (
        "fmt"
        "time"
)

func performance() {
        var on [18]bool
        start := time.Now()
        for i := 0; i < 100000000; i++ {
                on = [18]bool{}
        }
        elapsed := time.Since(start)
        fmt.Printf("100 mio x [18]bool{} - Took: %s\n", elapsed)

        start2 := time.Now()
        for i := 0; i < 100000000; i++ {
                for i := 0; i < 18; i++ {
                        on[i] = false
                }
        }
        elapsed2 := time.Since(start2)
        fmt.Printf("100 mio x loop - Took: %s\n", elapsed2)
        _ = on
}

func main() {
        for i := 0; i < 10; i++ {
                performance()
                fmt.Println("***************")
        }
}
```

experimental/devices/sn3218/sn3218.go
lnitram: Track the state is a good idea. But I disagree with wakeup on change. It should be possible to change the states for different channels and enable everything later at once.

Example use case (What I have in mind for my project)
```
        d.BrightnessAll(1)
        d.Switch(7, true) //Because a sensor mesaured something
        d.Switch(5, true) //Because a sensor measured something else
        d.WakeUp() //User pushed button to show the states
        time.Sleep(1000 * time.Millisecond)
        d.Sleep()  //Save power again
```

experimental/devices/sn3218/sn3218_test.go
lnitram: As the device is writeOnly and does not return anything, I decided to use Record. Happy to change it if really needed, but I would like to save the time here.

experimental/cmd/sn3218/main.go
lnitram: Should be really done now

experimental/devices/sn3218/example_test.go
lnitram: fixed, will have a look at goimports

experimental/devices/sn3218/example_test.go
lnitram: fixed

experimental/devices/sn3218/example_test.go
lnitram: fixed

experimental/devices/sn3218/sn3218.go
lnitram: done

experimental/devices/sn3218/sn3218.go
lnitram: changed to Sleep / WakeUp

experimental/devices/sn3218/sn3218.go
lnitram: fixed

experimental/devices/sn3218/sn3218.go
lnitram: done

experimental/devices/sn3218/sn3218.go
lnitram: done

experimental/devices/sn3218/sn3218.go
lnitram: refactored. It's now a member of Dev and reads from the state []

experimental/devices/sn3218/sn3218.go
lnitram: ok, done

experimental/devices/sn3218/sn3218.go
lnitram: agree, fixed

experimental/devices/sn3218/sn3218.go
lnitram: Track sleepMode still TODO

experimental/devices/sn3218/sn3218.go
lnitram: fixed

experimental/devices/sn3218/sn3218.go
lnitram: fixed

experimental/devices/sn3218/sn3218_test.go
lnitram: done

experimental/devices/sn3218/sn3218_test.go
lnitram: fixed

experimental/devices/sn3218/sn3218_test.go
lnitram: you mean just add an empty line? Then done. Otherwise, please explain

maruel: if err := dev.Halt(); err != nil {

lnitram: fixed, will send another pull requests together with some more improvements.

PR #349, opened 2018-12-05
mfrc522: reworked public/private API.

jdevelop:
Fixed the wait for edge block, removed synchronization code.
Moved private-related access methods to the low-level driver.
Added concurrent execution protection via config.

### DIFF
--- a/experimental/devices/mfrc522/blockaccess.go
+++ b/experimental/devices/mfrc522/blockaccess.go
@@ -31,25 +31,27 @@ const (
 
 // Access bits for the sector trail.
 // Every trail sector has the options for controlling the access to the trailing sector bits.
-// For example : KeyA_R[Key]_W[Key]_BITS_R[Key]_W[Key]_KeyB_R[Key]_W[Key]
+// For example :
 //
-// - KeyA
-//   - could be Read by providing [Key] ( where [Key] could be KeyA or KeyB )
-//   - could be Written by Providing [Key] ( where [Key] is KeyA or KeyB )
-// - access bits for the sector data (see above)
-//   - could be Read by providing [Key] ( where [Key] could be KeyA or KeyB )
-//   - could be Written by Providing [Key] ( where [Key] is KeyA or KeyB )
-// - KeyB
-//   - could be Read by providing [Key] ( where [Key] could be KeyA or KeyB )
-//   - could be Written by Providing [Key] ( where [Key] is KeyA or KeyB )
+//  KeyA_R[Key]_W[Key]_BITS_R[Key]_W[Key]_KeyB_R[Key]_W[Key]
 //
-// example:
+//  - KeyA
+//    - could be Read by providing [Key] ( where [Key] could be KeyA or KeyB )
+//    - could be Written by Providing [Key] ( where [Key] is KeyA or KeyB )
+//  - access bits for the sector data (see above)
+//    - could be Read by providing [Key] ( where [Key] could be KeyA or KeyB )
+//    - could be Written by Providing [Key] ( where [Key] is KeyA or KeyB )
+//  - KeyB
+//    - could be Read by providing [Key] ( where [Key] could be KeyA or KeyB )
+//    - could be Written by Providing [Key] ( where [Key] is KeyA or KeyB )
 //
-//  KeyA_RN_WA_BITS_RA_WA_KeyB_RA_WA means
-//  - KeyA could not be read but could be overwriten if KeyA is provided
-//  - Access bits could be read and overwritten if KeyA is provided during the card authentication
-//  - KeyB could be read and overriten if KeyA is provided during the card authentication
-// more on the matter: https://www.nxp.com/docs/en/data-sheet/MF1S50YYX_V1.pdf
+//  example:
+//
+//   KeyA_RN_WA_BITS_RA_WA_KeyB_RA_WA means
+//   - KeyA could not be read but could be overwriten if KeyA is provided
+//   - Access bits could be read and overwritten if KeyA is provided during the card authentication
+//   - KeyB could be read and overriten if KeyA is provided during the card authentication
+//  more on the matter: https://www.nxp.com/docs/en/data-sheet/MF1S50YYX_V1.pdf
 const (
 	KeyA_RN_WA_BITS_RA_WN_KeyB_RA_WA        SectorTrailerAccess = 0x0
 	KeyA_RN_WN_BITS_RA_WN_KeyB_RA_WN        SectorTrailerAccess = 0x02
@@ -73,7 +75,7 @@ func (ba *BlocksAccess) String() string {
 
 // serialize calculates the block access and stores it into the passed slice, that must be at least 4 bytes wide.
 func (ba *BlocksAccess) serialize(dst []byte) error {
-	if len(dst) < 4 {
+	if len(dst) != 4 {
 		return wrapf("serialized array must be of size at least 4")
 	}
 	dst[0] = ((^ba.getBits(2) & 0x0F) << 4) | (^ba.getBits(1) & 0x0F)
--- a/experimental/devices/mfrc522/blockaccess.go
+++ b/experimental/devices/mfrc522/blockaccess.go
@@ -1,0 +1,104 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package mfrc522 controls a Mifare RFID card reader.
+//
+// Datasheet
+//
+// https://www.nxp.com/docs/en/data-sheet/MFRC522.pdf
+package mfrc522
+
+import "fmt"
+
+// BlockAccess defines the block access bits.
+type BlockAccess byte
+
+// SectorTrailerAccess defines the sector trailing block access bits.
+type SectorTrailerAccess byte
+
+// Access bits for the sector data.
+const (
+	AnyKeyRWID    BlockAccess = 0x0  // Any key (A or B) can read, write, increment and decrement block.
+	RAB_WN_IN_DN  BlockAccess = 0x02 // Read (A or B), Write (None), Increment (None), Decrement (None)
+	RAB_WB_IN_DN  BlockAccess = 0x04 // Read (A orB), Write (B), Increment (None), Decrement (None)
+	RAB_WB_IB_DAB BlockAccess = 0x06 // Read (A or B), Write (B), Icrement (B), Decrement (A or B)
+	RAB_WN_IN_DAB BlockAccess = 0x01 // Read (A or B), Write (None), Increment (None), Decrment (A or B)
+	RB_WB_IN_DN   BlockAccess = 0x03 // Read (B), Write (B), Increment (None), Decrement (None)
+	RB_WN_IN_DN   BlockAccess = 0x05 // Read (B), Write (None), Increment (None), Decrement (None)
+	RN_WN_IN_DN   BlockAccess = 0x07 // Read (None), Write (None), Increment (None), Decrement (None)
+)
+
+// Access bits for the sector trail.
+// Every trail sector has the options for controlling the access to the trailing sector bits.
+// For example : KeyA_R[Key]_W[Key]_BITS_R[Key]_W[Key]_KeyB_R[Key]_W[Key]
+//
+// - KeyA
+//   - could be Read by providing [Key] ( where [Key] could be KeyA or KeyB )
+//   - could be Written by Providing [Key] ( where [Key] is KeyA or KeyB )
+// - access bits for the sector data (see above)
+//   - could be Read by providing [Key] ( where [Key] could be KeyA or KeyB )
+//   - could be Written by Providing [Key] ( where [Key] is KeyA or KeyB )
+// - KeyB
+//   - could be Read by providing [Key] ( where [Key] could be KeyA or KeyB )
+//   - could be Written by Providing [Key] ( where [Key] is KeyA or KeyB )
+//
+// example:
+//
+//  KeyA_RN_WA_BITS_RA_WA_KeyB_RA_WA means
+//  - KeyA could not be read but could be overwriten if KeyA is provided
+//  - Access bits could be read and overwritten if KeyA is provided during the card authentication
+//  - KeyB could be read and overriten if KeyA is provided during the card authentication
+// more on the matter: https://www.nxp.com/docs/en/data-sheet/MF1S50YYX_V1.pdf
+const (
+	KeyA_RN_WA_BITS_RA_WN_KeyB_RA_WA        SectorTrailerAccess = 0x0
+	KeyA_RN_WN_BITS_RA_WN_KeyB_RA_WN        SectorTrailerAccess = 0x02
+	KeyA_RN_WB_BITS_RAB_WN_KeyB_RN_WB       SectorTrailerAccess = 0x04
+	KeyA_RN_WN_BITS_RAB_WN_KeyB_RN_WN       SectorTrailerAccess = 0x06
+	KeyA_RN_WA_BITS_RA_WA_KeyB_RA_WA        SectorTrailerAccess = 0x01
+	KeyA_RN_WB_BITS_RAB_WB_KeyB_RN_WB       SectorTrailerAccess = 0x03
+	KeyA_RN_WN_BITS_RAB_WB_KeyB_RN_WN       SectorTrailerAccess = 0x05
+	KeyA_RN_WN_BITS_RAB_WN_KeyB_RN_WN_EXTRA SectorTrailerAccess = 0x07
+)
+
+// BlocksAccess defines the access structure for first 3 blocks of the sector and the access bits for the sector trail.
+type BlocksAccess struct {
+	B0, B1, B2 BlockAccess
+	B3         SectorTrailerAccess
+}
+
+func (ba *BlocksAccess) String() string {
+	return fmt.Sprintf("B0: %d, B1: %d, B2: %d, B3: %d", ba.B0, ba.B1, ba.B2, ba.B3)
+}
+
+// serialize calculates the block access and stores it into the passed slice, that must be at least 4 bytes wide.
+func (ba *BlocksAccess) serialize(dst []byte) error {
+	if len(dst) < 4 {
+		return wrapf("serialized array must be of size at least 4")
+	}
+	dst[0] = ((^ba.getBits(2) & 0x0F) << 4) | (^ba.getBits(1) & 0x0F)
+	dst[1] = ((ba.getBits(1) & 0x0F) << 4) | (^ba.getBits(3) & 0x0F)
+	dst[2] = ((ba.getBits(3) & 0x0F) << 4) | (ba.getBits(2) & 0x0F)
+	dst[3] = dst[0] ^ dst[1] ^ dst[2]
+	return nil
+}
+
+// Init parses the given byte array into the block access structure.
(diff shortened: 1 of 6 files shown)


jdevelop:
@maruel , @vkorn please have a look

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,73 @@
+package mfrc522
```
maruel: Please add copyright here.

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,73 @@
...
+// BlockAccess defines the block access bits.
+type BlockAccess byte
+
+// SectorTrailerAccess defines the sector trailing block access bits.
+type SectorTrailerAccess byte
+
+// Access bits.
+const (
+	AnyKeyRWID    BlockAccess = iota
+	RAB_WN_IN_DN              = 0x02 // Read (A|B), Write (None), Increment (None), Decrement(None)
```
maruel: To make godoc do the right thing, please repeat the BlockAccess type on each constant, same for SectorTrailerAccess below.

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,73 @@
...
+	B0, B1, B2 BlockAccess
+	B3         SectorTrailerAccess
+}
+
+func (ba *BlocksAccess) String() string {
+	return fmt.Sprintf("B0: %d, B1: %d, B2: %d, B3: %d", ba.B0, ba.B1, ba.B2, ba.B3)
+}
+
+// CalculateBlockAccess calculates the block access.
+func CalculateBlockAccess(ba *BlocksAccess) []byte {
```
maruel: While this was only moved code and you didn't change it, would you mind to change the signature to:
```
// Serializes calculates the serialized block access as 4 bytes.
func (b *BlockAccess) Serialize(dst []byte) {
```

Does this *really* needs to be exported?

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,73 @@
...
+	res := make([]byte, 4)
+	res[0] = ((^ba.getBits(2) & 0x0F) << 4) | (^ba.getBits(1) & 0x0F)
+	res[1] = ((ba.getBits(1) & 0x0F) << 4) | (^ba.getBits(3) & 0x0F)
+	res[2] = ((ba.getBits(3) & 0x0F) << 4) | (ba.getBits(2) & 0x0F)
+	res[3] = res[0] ^ res[1] ^ res[2]
+	return res
+}
+
+// ParseBlockAccess parses the given byte array into the block access structure.
```
maruel: This one could be
```
func (b *BlockAccess) Init(src []byte) {
```

The end goal of this request and the one above is to reduce memory allocations.

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -33,34 +42,79 @@ func NewLowLevelSPI(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn) (
...
-type LowLevel struct {
-	resetPin gpio.PinOut
-	irqPin   gpio.PinIn
-	spiDev   spi.Conn
+// Reset resets the RFID chip to initial state.
+func (r *LowLevel) Reset() error {
+	return r.DevWrite(CommandReg, PCD_RESETPHASE)
+}
+
+func (r *LowLevel) SetAntennaGain(gain int) {
```
maruel: Please add a comment.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -20,107 +20,83 @@ import (
...
 }
 
-// ParseBlockAccess parses the given byte array into the block access structure.
-func ParseBlockAccess(ad []byte) *BlocksAccess {
-	return &BlocksAccess{
-		B0: BlockAccess(((ad[1] & 0x10) >> 2) | ((ad[2] & 0x01) << 1) | ((ad[2] & 0x10) >> 5)),
-		B1: BlockAccess(((ad[1] & 0x20) >> 3) | (ad[2] & 0x02) | ((ad[2] & 0x20) >> 5)),
-		B2: BlockAccess(((ad[1] & 0x40) >> 4) | ((ad[2] & 0x04) >> 1) | ((ad[2] & 0x40) >> 6)),
-		B3: SectorTrailerAccess(((ad[1] & 0x80) >> 5) | ((ad[2] & 0x08) >> 2) | ((ad[2] & 0x80) >> 7)),
+// WithSync	sets the synchronization for the entire device, using internal mutex.
```
maruel: s/\t/ /

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -20,107 +20,83 @@ import (
...
+func WithSync() configF {
+	var mtx sync.Mutex
```
maruel: Standard name for a sync.Mutex instance is `mu`. Please rename.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -20,107 +20,83 @@ import (
...
+	return func(c *config) *config {
+		c.beforeCall = func() {
```
maruel: ?
```
c.beforeCall = mu.Lock
```

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -130,65 +106,50 @@ func (r *Dev) String() string {
...
+//  block 	the block within sector to authenticate.
+//  key 	the key to be used for accessing the sector data.
 func (r *Dev) ReadCard(auth byte, sector int, block int, key [6]byte) (data []byte, err error) {
-	if err = r.accessStarted(); err != nil {
-		return
-	}
+	return r.ReadCardTimed(r.operationTimeout, auth, sector, block, key)
+}
 
+// ReadCardTimed 	reads the card sector/block with IRQ event timeout.
```
maruel: s/\t/ / everywhere in the first line of method docs.

maruel:
gohci

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -20,107 +20,83 @@ import (
...
+	LowLevel         *commands.LowLevel
+	operationTimeout time.Duration
+	beforeCall       func()
+	afterCall        func()
 }
 
-func (ba *BlocksAccess) String() string {
-	return fmt.Sprintf("B0: %d, B1: %d, B2: %d, B3: %d", ba.B0, ba.B1, ba.B2, ba.B3)
+// DefaultKey	provides the default bytes for card authentication for method B.
+var DefaultKey = [...]byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF}
```
maruel: Please create a `type Key [6]byte` and document it.

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,73 @@
...
+
+// SectorTrailerAccess defines the sector trailing block access bits.
+type SectorTrailerAccess byte
+
+// Access bits.
+const (
+	AnyKeyRWID    BlockAccess = iota
+	RAB_WN_IN_DN              = 0x02 // Read (A|B), Write (None), Increment (None), Decrement(None)
+	RAB_WB_IN_DN              = 0x04
+	RAB_WB_IB_DAB             = 0x06
```
maruel: Could the name be a bit more self-standing? Right now I have no clue what to make of these constants.

I mean line 14 is a hint, but it's not super clear. Is it a bitmask or individual values?

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,73 @@
+package mfrc522
+
+import "fmt"
+
+// BlockAccess defines the block access bits.
```
maruel: Can you expand on what this means? Same for SectorTrailerAccess.

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,73 @@
...
+	AnyKeyRWID    BlockAccess = iota
+	RAB_WN_IN_DN              = 0x02 // Read (A|B), Write (None), Increment (None), Decrement(None)
+	RAB_WB_IN_DN              = 0x04
+	RAB_WB_IB_DAB             = 0x06
+	RAB_WN_IN_DAB             = 0x01
+	RB_WB_IN_DN               = 0x03
+	RB_WN_IN_DN               = 0x05
+	RN_WN_IN_DN               = 0x07
+
+	KeyA_RN_WA_BITS_RA_WN_KeyB_RA_WA        SectorTrailerAccess = iota
```
maruel: Why KeyA, is there KeyB?

experimental/devices/mfrc522/blockaccess.go
jdevelop: I tried to add some more or less meaningful explanations in the comments, please have a look.

experimental/devices/mfrc522/blockaccess.go
jdevelop: Yes, there are KeyA and KeyB that could be provided during the card authentication stage.

experimental/devices/mfrc522/blockaccess.go
jdevelop: This is sort of hard to explain w/o reciting much of the datasheet.

maruel:
The only big problem is the race condition I noted. Also double check how some godoc shows up, I'm not sure it'll be well formatted everywhere but that's less critical.

maruel:
The only big problem is the race condition I noted. Also double check how some godoc shows up, I'm not sure it'll be well formatted everywhere but that's less critical.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -130,65 +104,50 @@ func (r *Dev) String() string {
...
 	}
-
-	r.oMu.Lock()
-	defer r.oMu.Unlock()
-	r.antennaGain = gain
+	r.LowLevel.SetAntennaGain(gain)
 	return nil
 }
 
 // ReadCard reads the card sector/block.
```
maruel: I'd prefer a smaller interface; I'm really not a fan of the Foo/FooTimed duplication. I'd recommend to:
- Remove the non timed version
- Rename the Timed one without the Timed suffix

What you can do is that when timeout is 0, use the default value. That will simplify the API.

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,104 @@
...
+//
+// example:
+//
+//  KeyA_RN_WA_BITS_RA_WA_KeyB_RA_WA means
+//  - KeyA could not be read but could be overwriten if KeyA is provided
+//  - Access bits could be read and overwritten if KeyA is provided during the card authentication
+//  - KeyB could be read and overriten if KeyA is provided during the card authentication
+// more on the matter: https://www.nxp.com/docs/en/data-sheet/MF1S50YYX_V1.pdf
+const (
+	KeyA_RN_WA_BITS_RA_WN_KeyB_RA_WA        SectorTrailerAccess = 0x0
```
maruel: Same here, this is exposing internal bits while in practice you could write it in a more meaningful way by not using the acronyms. This can be done later.

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,104 @@
...
+	B3         SectorTrailerAccess
+}
+
+func (ba *BlocksAccess) String() string {
+	return fmt.Sprintf("B0: %d, B1: %d, B2: %d, B3: %d", ba.B0, ba.B1, ba.B2, ba.B3)
+}
+
+// serialize calculates the block access and stores it into the passed slice, that must be at least 4 bytes wide.
+func (ba *BlocksAccess) serialize(dst []byte) error {
+	if len(dst) < 4 {
```
maruel: I'd recommend =

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -592,6 +447,9 @@ func (r *Dev) preAccess(blockAddr byte, cmd byte) ([]byte, int, error) {
 	return r.LowLevel.CardWrite(commands.PCD_TRANSCEIVE, send)
 }
 
+// read	reads the block
```
maruel: Please don't use tabs in the documentation.

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -20,107 +20,81 @@ import (
...
-)
-
-// BlocksAccess defines the access structure for first 3 blocks of the sector
-// and the access bits for the sector trail.
-type BlocksAccess struct {
-	B0, B1, B2 BlockAccess
-	B3         SectorTrailerAccess
+// Dev is an handle to an MFRC522 RFID reader.
+type Dev struct {
+	LowLevel         *commands.LowLevel
```
maruel: One thing you can do is to embed the pointer, i.e. do:
```
  *commands.LowLevel
```

without a name. This will expose all its methods, including its own embedded sync.Mutex. That's what lepton does; https://github.com/google/periph/blob/master/devices/lepton/lepton.go#L110

It is very versatile for the users while keeping the godoc tight.

maruel:
gohci

maruel:
Weird, I think I had many comments I had forgot posting so you'll see a few outdated comment. Sorry about that. :/

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -13,17 +13,26 @@ import (
...
-// AuthReadFailure or AuthFailure
-type AuthStatus byte
-
 // Card authentication status enum.
 const (
-	AuthOk AuthStatus = iota
-	AuthReadFailure
-	AuthFailure
+	AuthOk          AuthStatus = iota
+	AuthReadFailure AuthStatus = iota
```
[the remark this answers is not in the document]
maruel: Thanks! Yes I prefer the second.

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,104 @@
...
+
+// BlockAccess defines the block access bits.
+type BlockAccess byte
+
+// SectorTrailerAccess defines the sector trailing block access bits.
+type SectorTrailerAccess byte
+
+// Access bits for the sector data.
+const (
+	AnyKeyRWID    BlockAccess = 0x0  // Any key (A or B) can read, write, increment and decrement block.
```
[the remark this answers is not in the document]
jdevelop: I couldn't make sense of these bits to organize them in something more readable, like `KeyA | READ | WRITE | INCREMENT`, but it could be reworked to look in that way.

experimental/devices/mfrc522/blockaccess.go
```diff
@@ -0,0 +1,103 @@
...
+	RAB_WB_IB_DAB BlockAccess = 0x06 // Read (A or B), Write (B), Icrement (B), Decrement (A or B)
+	RAB_WN_IN_DAB BlockAccess = 0x01 // Read (A or B), Write (None), Increment (None), Decrment (A or B)
+	RB_WB_IN_DN   BlockAccess = 0x03 // Read (B), Write (B), Increment (None), Decrement (None)
+	RB_WN_IN_DN   BlockAccess = 0x05 // Read (B), Write (None), Increment (None), Decrement (None)
+	RN_WN_IN_DN   BlockAccess = 0x07 // Read (None), Write (None), Increment (None), Decrement (None)
+)
+
+// Access bits for the sector trail.
+// Every trail sector has the options for controlling the access to the trailing sector bits.
+// For example : KeyA_R[Key]_W[Key]_BITS_R[Key]_W[Key]_KeyB_R[Key]_W[Key]
```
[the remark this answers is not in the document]
jdevelop: Yep, indentation was screwed :)

experimental/devices/mfrc522/blockaccess.go
jdevelop: ok

experimental/devices/mfrc522/commands/low_level.go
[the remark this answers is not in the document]
jdevelop: ok :)

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -20,107 +20,81 @@ import (
...
 }
 
-// ParseBlockAccess parses the given byte array into the block access structure.
-func ParseBlockAccess(ad []byte) *BlocksAccess {
-	return &BlocksAccess{
-		B0: BlockAccess(((ad[1] & 0x10) >> 2) | ((ad[2] & 0x01) << 1) | ((ad[2] & 0x10) >> 5)),
-		B1: BlockAccess(((ad[1] & 0x20) >> 3) | (ad[2] & 0x02) | ((ad[2] & 0x20) >> 5)),
-		B2: BlockAccess(((ad[1] & 0x40) >> 4) | ((ad[2] & 0x04) >> 1) | ((ad[2] & 0x40) >> 6)),
-		B3: SectorTrailerAccess(((ad[1] & 0x80) >> 5) | ((ad[2] & 0x08) >> 2) | ((ad[2] & 0x80) >> 7)),
+// WithSyncsets the synchronization for the entire device, using internal mutex.
```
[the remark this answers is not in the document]
jdevelop: That is the way to add the sync over the device as a "trait". If someone wants to use that device with the "locks" - then with this configuration wrapper for before/after all the operations will be synchronized using the private mutex. 
Otherwise for a single-threaded access no mutex is necessary. Hence no need for an embedded mutex at all.

experimental/devices/mfrc522/mfrc522.go
jdevelop: This is also exposes the low-level details / methods into a high-level interface. So I'd rather keep those separated. The whole point of extracting that low-level part was to create a high-level interface with **fewer** methods exposed.

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -145,9 +193,40 @@ func (r *LowLevel) StopCrypto() error {
...
-func (r *LowLevel) WaitForEdge(timeout time.Duration) bool {
-	return r.irqPin.WaitForEdge(timeout)
+// WaitForEdge waits for an IRQ pin to strobe. If IRQ pin is not set, then always returns false immediately.
+func (r *LowLevel) WaitForEdge(timeout time.Duration) error {
+	irqChannel := make(chan bool)
+	go func() {
+		irqChannel <- r.irqPin.WaitForEdge(timeout)
+	}()
+
+	defer func() {
```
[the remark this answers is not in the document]
jdevelop: This should be fine now, I guess.

maruel:
gohci

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -145,9 +193,37 @@ func (r *LowLevel) StopCrypto() error {
...
 }
 
-// WaitForEdge waits for an IRQ pin to strobe.
-func (r *LowLevel) WaitForEdge(timeout time.Duration) bool {
-	return r.irqPin.WaitForEdge(timeout)
+// WaitForEdge waits for an IRQ pin to strobe. If IRQ pin is not set, then always returns false immediately.
+func (r *LowLevel) WaitForEdge(timeout time.Duration) error {
+	irqChannel := make(chan bool)
+	go func() {
+		defer close(irqChannel)
```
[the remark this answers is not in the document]
jdevelop: Perhaps, just wanted to make it clean that the channel is never used out of the goroutine itself

PR #350, opened 2018-12-07
physic: add flag.Value implementation for Mass and Speed 

NeuralSpaz:
This adds support for the flag.Value interface for physics.Mass and physic.Speed units types.
Adds Set and flag doc examples.
Adds Set() for Mass and Speed.
Add Tests for Mass and Speed Set().

I'm not happy with `func (sp *Speed) Set(s string) error` but was blocked on anything better.
I'm also floating (I think) a slightly better error message for non si units overflowing, for Mass and Speed the max/min of unit in error is the same as the unit supplied. If that is better I will follow up in kind for other units.

Helps with #303.

### DIFF
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -768,9 +768,9 @@ func (m *Mass) Set(s string) error {
 			if e, ok := err.(*parseError); ok {
 				switch e.err {
 				case errOverflowsInt64:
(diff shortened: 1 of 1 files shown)


maruel:
lgtm but my Windows box broke, let me fix this.

PR #351, opened 2018-12-10
physic: Adds flag.Value for RelativeHumidity

NeuralSpaz:
This adds support for the flag.Value interface for RelativeHumidity unit type.
Add Set and flag doc examples.
Add Set() for RelativeHumidity.
Add Tests for RelativeHumidity Set().
Add Set() and flag.Value examples for RelativeHumidity.

Helps with #303.

### DIFF
--- a/conn/physic/example_test.go
+++ b/conn/physic/example_test.go
@@ -401,7 +401,7 @@ func ExampleRelativeHumidity_Set() {
 	}
 	fmt.Println(r)
 
-	if err := r.Set("20%rH"); err != nil {
+	if err := r.Set("20%"); err != nil {
 		log.Fatal(err)
 	}
 	fmt.Println(r)
--- a/conn/physic/example_test.go
+++ b/conn/physic/example_test.go
@@ -393,6 +393,30 @@ func ExampleRelativeHumidity() {
 	// 20%rH
 }
 
+func ExampleRelativeHumidity_Set() {
+	var r physic.RelativeHumidity
+
+	if err := r.Set("50.6%rH"); err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println(r)
+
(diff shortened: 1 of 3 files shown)


conn/physic/example_test.go
```diff
@@ -393,6 +393,30 @@ func ExampleRelativeHumidity() {
...
 
+func ExampleRelativeHumidity_Set() {
+	var r physic.RelativeHumidity
+
+	if err := r.Set("50.6%rH"); err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println(r)
+
+	if err := r.Set("20%rH"); err != nil {
```
maruel: You can put "20%" in the example, to show that it works?

conn/physic/units.go
```diff
@@ -920,11 +920,60 @@ func (h RelativeHumidity) String() string {
...
+	return nil
+}
+
 const (
 	TenthMicroRH RelativeHumidity = 1                 // 0.00001%rH
 	MicroRH      RelativeHumidity = 10 * TenthMicroRH // 0.0001%rH
 	MilliRH      RelativeHumidity = 1000 * MicroRH    // 0.1%rH
 	PercentRH    RelativeHumidity = 10 * MilliRH      // 1%rH
+
+	maxRelativeHumidity RelativeHumidity = (1 << 31) - 1
```
maruel: It's between 0% and 100%.
https://periph.io/x/periph/conn/physic#RelativeHumidity

I prefer to deny parsing values outside this range, or at the very least values below 0.

conn/physic/example_test.go
NeuralSpaz: sure can.

maruel:
Humm looks like the beaglebone died again.

PR #353, opened 2018-12-11
sysfs: implement SenseContinuous in ThermalSensor

davidsansome:


maruel:
The code lgtm but please fix the test.

davidsansome:
I fixed the test again :D

cmd/thermal/main.go
```diff
@@ -12,13 +12,16 @@ import (
...
 	"os"
+	"time"
 
 	"periph.io/x/periph/conn/physic"
 	"periph.io/x/periph/host"
 	"periph.io/x/periph/host/sysfs"
 )
 
 func mainImpl() error {
+	cont := flag.String("cont", "", "Reads continuously from the sensor with this name")
```
maruel: I think it should be `-dev` and looks as:
- If not specified, read all devices found
- If specified, read only this device

If you want to simplify your life, you could require `-dev` when using `-interval`.

maruel:
gohci

cmd/thermal/main.go
davidsansome: Done, sorry for the delay again!

maruel:
gohci

PR #354, opened 2018-12-11
physic: Adds flag.Value for physic.Temperature

### LINKED ISSUE #303: feature: physic.Units flag.Value interface

NeuralSpaz:
This adds support for the flag.Value interface for Temperature unit type.
Add Set and flag doc examples.
Add Set() for Temperature.
Add Tests for Temperature Set().
Add Set() and flag.Value examples for Temperature.
Fix Fahrenheit rounding error in Example.
Fix Temperature stringer overflow for temperatures > allowed by Celsius.

This should close #303.

### DIFF
--- a/conn/physic/units.go
+++ b/conn/physic/units.go
@@ -1160,7 +1160,124 @@ type Temperature int64
 
 // String returns the temperature formatted as a string in °Celsius.
 func (t Temperature) String() string {
+	if t < minCelsius || t > maxCelsius {
+		return nanoAsString(int64(t)) + "K"
+	}
 	return nanoAsString(int64(t-ZeroCelsius)) + "°C"
+}
+
+// Set sets the Temperature to the value represented by s. Units are to be
+// provided in "C", "°C", "F" or "K" with an optional SI prefix: "p", "n", "u",
+// "µ", "m", "k", "M", "G" or "T".
+func (t *Temperature) Set(s string) error {
+	d, n, err := atod(s)
+	if err != nil {
+		if e, ok := err.(*parseError); ok {
+			switch e.err {
+			case errNotANumber:
+				if found, _ := containsUnitString(s[n:], "°C", "C", "F", "K"); found != "" {
+					return errors.New("does not contain number")
+				}
+				return errors.New("does not contain number or unit \"°C\"")
+			case errOverflowsInt64:
+				return errors.New("maximum value is " + maxTemperature.String())
+			case errOverflowsInt64Negative:
+				return errors.New("minimum value is " + minTemperature.String())
+			}
+		}
+		return err
+	}
+
+	var si prefix
+	if n != len(s) {
+		r, rsize := utf8.DecodeRuneInString(s[n:])
+		if r <= 1 || rsize == 0 {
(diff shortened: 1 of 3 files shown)


conn/physic/units.go
```diff
@@ -1178,9 +1295,20 @@ const (
 	Celsius      Temperature = Kelvin
 
 	// Conversion between Kelvin and Fahrenheit.
-	ZeroFahrenheit  Temperature = 255372 * MilliKelvin
+	ZeroFahrenheit  Temperature = 255372222222 * NanoKelvin
 	MilliFahrenheit Temperature = 555555 * NanoKelvin
 	Fahrenheit      Temperature = 555555555 * NanoKelvin
+
+	maxTemperature Temperature = (1 << 63) - 1
+	minTemperature Temperature = -((1 << 63) - 1)
```
maruel: I think we should set it to 0. For the same reason as with RelativeHumidity, there's no true negative temperature.

That said there can be "negative rate of change" when looking at a differential of the measure. In this case we should allow the range to exceed, but this is only because the type is used to store a different unit (K/s or %rH/s) than what it is supposed to.

I'm still ambivalent, but both should be coherent (either we support or we don't)

conn/physic/units.go
```diff
@@ -1160,9 +1160,126 @@ type Temperature int64
...
 // String returns the temperature formatted as a string in °Celsius.
 func (t Temperature) String() string {
+	if t < minCelsius || t > maxCelsius {
+		return nanoAsString(int64(t)) + "K"
+	}
 	return nanoAsString(int64(t-ZeroCelsius)) + "°C"
 }
 
+// Set sets the Temperature to the value represented by s. Units are to be
+// provided in "C", "°C", "F" or "K" with an optional SI prefix: "p", "n", "u",
```
maruel: Farenheit is officially "°F" so it should be supported too.

conn/physic/units.go
NeuralSpaz: ok, I was thinking of more of and delta scenario, but that could always be expressed as +ve, so I'll make 0K the limit for all units.

conn/physic/units.go
NeuralSpaz: °F supported now.

maruel:
I'm really grateful you did all this work. Thanks a lot! You can squash and merge yourself. :)

NeuralSpaz:
: ) np.

maruel:
I think we miss only one test: for each type, do a full round trip.

```
x := 123 * physic.Meter
var y Distance
if err := y.Set(x.String()) {
  t.Fatal(err)
}
if x != y {
  t.Fatalf("...")
}
```

NeuralSpaz:
ahh, yes, that is a good set of tests.

PR #356, opened 2018-12-18
mcp9808: add experimental support for digital temperature sensor

NeuralSpaz:
Add initial support for microchip i2c MCP9808 temperature sensor.

### DIFF
--- a/experimental/cmd/mcp9808/main.go
+++ b/experimental/cmd/mcp9808/main.go
@@ -55,7 +55,7 @@ func mainImpl() error {
 	for {
 		select {
 		case <-everySecond:
-			t, err := sensor.Sense()
+			t, err := sensor.SenseTemp()
 			if err != nil {
 				return fmt.Errorf("sensor reading error: %v", err)
 			}
--- a/experimental/cmd/mcp9808/main.go
+++ b/experimental/cmd/mcp9808/main.go
@@ -1,0 +1,74 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// mcp9808 communicates with an mcp9808 sensor reading ambient temperature.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"periph.io/x/periph/conn/i2c/i2creg"
+	"periph.io/x/periph/experimental/devices/mcp9808"
+	"periph.io/x/periph/host"
+)
+
+func mainImpl() error {
+	if _, err := host.Init(); err != nil {
+		return err
+	}
+	address := flag.Int("address", 0x18, "I²C address")
+	i2cbus := flag.String("bus", "", "I²C bus (/dev/i2c-1)")
+
+	flag.Parse()
+
+	fmt.Println("Starting MCP9808 Temperature Sensor")
+	if _, err := host.Init(); err != nil {
+		return err
+	}
+
+	// Open default I²C bus.
+	bus, err := i2creg.Open(*i2cbus)
+	if err != nil {
+		return fmt.Errorf("failed to open I²C: %v", err)
+	}
+	defer bus.Close()
+
+	// Create a new temperature sensor a sense with default options.
+	sensor, err := mcp9808.New(bus, &mcp9808.Opts{Address: *address})
+	if err != nil {
+		return fmt.Errorf("failed to open new sensor: %v", err)
+	}
+
+	// Read values from sensor every second.
+	everySecond := time.NewTicker(time.Second).C
+	var halt = make(chan os.Signal)
+	signal.Notify(halt, syscall.SIGTERM)
+	signal.Notify(halt, syscall.SIGINT)
+
+	fmt.Println("ctrl+c to exit")
+	for {
+		select {
+		case <-everySecond:
+			t, err := sensor.Sense()
+			if err != nil {
+				return fmt.Errorf("sensor reading error: %v", err)
+			}
+			fmt.Println(t)
+		case <-halt:
+			return nil
+		}
+	}
+}
+
+func main() {
+	if err := mainImpl(); err != nil {
+		fmt.Fprintf(os.Stderr, "mcp9808: %s.\n", err)
+		return
+	}
+}
--- a/experimental/cmd/mcp9808/main.go
+++ b/experimental/cmd/mcp9808/main.go
@@ -40,13 +40,13 @@ func mainImpl() error {
 	defer bus.Close()
 
 	// Create a new temperature sensor a sense with default options.
-	sensor, err := mcp9808.New(bus, &mcp9808.Opts{Address: *address})
+	sensor, err := mcp9808.New(bus, &mcp9808.Opts{Addr: *address})
 	if err != nil {
 		return fmt.Errorf("failed to open new sensor: %v", err)
 	}
 
 	// Read values from sensor every second.
-	everySecond := time.NewTicker(time.Second).C
+	everySecond := time.Tick(time.Second)
 	var halt = make(chan os.Signal)
 	signal.Notify(halt, syscall.SIGTERM)
 	signal.Notify(halt, syscall.SIGINT)
--- a/experimental/devices/mcp9808/example_test.go
+++ b/experimental/devices/mcp9808/example_test.go
@@ -8,14 +8,13 @@ import (
 	"fmt"
 	"log"
 
-	"periph.io/x/periph/conn/physic"
-
 	"periph.io/x/periph/conn/i2c/i2creg"
+	"periph.io/x/periph/conn/physic"
 	"periph.io/x/periph/experimental/devices/mcp9808"
 	"periph.io/x/periph/host"
 )
 
-func ExampleDev_Sense() {
+func ExampleDev_SenseTemp() {
 	// Make sure periph is initialized.
 	if _, err := host.Init(); err != nil {
 		log.Fatal(err)
@@ -35,7 +34,7 @@ func ExampleDev_Sense() {
 	}
 
 	// Read values from sensor.
-	measurement, err := sensor.Sense()
+	measurement, err := sensor.SenseTemp()
 
 	if err != nil {
 		log.Fatalln(err)
@@ -74,10 +73,8 @@ func ExampleDev_SenseWithAlerts() {
 		log.Fatalln(err)
 	}
 
-	if len(alerts) > 0 {
-		for _, alert := range alerts {
-			fmt.Println(alert)
-		}
+	for _, alert := range alerts {
+		fmt.Println(alert)
 	}
 
 	fmt.Println(temperature)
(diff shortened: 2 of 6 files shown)


experimental/devices/mcp9808/example_test.go
```diff
@@ -8,12 +8,14 @@ import (
 	"fmt"
 	"log"
 
+	"periph.io/x/periph/conn/physic"
+
```
maruel: Don't include an empty line here, did you add the import manually or did you use `goimports -w`?

experimental/cmd/mcp9808/main.go
```diff
@@ -0,0 +1,74 @@
...
+	defer bus.Close()
+
+	// Create a new temperature sensor a sense with default options.
+	sensor, err := mcp9808.New(bus, &mcp9808.Opts{Address: *address})
+	if err != nil {
+		return fmt.Errorf("failed to open new sensor: %v", err)
+	}
+
+	// Read values from sensor every second.
+	everySecond := time.NewTicker(time.Second).C
```
maruel: https://golang.org/pkg/time/#Tick

experimental/cmd/mcp9808/main.go
```diff
@@ -0,0 +1,74 @@
...
+	var halt = make(chan os.Signal)
+	signal.Notify(halt, syscall.SIGTERM)
+	signal.Notify(halt, syscall.SIGINT)
+
+	fmt.Println("ctrl+c to exit")
+	for {
+		select {
+		case <-everySecond:
+			t, err := sensor.Sense()
```
maruel: One interesting thing here: the first reading will be after 1s, instead of immediately at program startup.

experimental/devices/mcp9808/example_test.go
```diff
@@ -0,0 +1,84 @@
...
+	critical := physic.ZeroCelsius + 32*physic.Celsius
+
+	// Read values from sensor.
+	temperature, alerts, err := sensor.SenseWithAlerts(lower, upper, critical)
+
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	if len(alerts) > 0 {
```
maruel: This is unnecessary.

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -0,0 +1,345 @@
...
+	"periph.io/x/periph/conn/i2c"
+	"periph.io/x/periph/conn/mmr"
+	"periph.io/x/periph/conn/physic"
+)
+
+// Opts holds the configuration options.
+//
+// Slave Address
+//
+// Depending which pins the A0, A1 and A2 pins are connected to will change the
```
maruel: Put this as a comment to `Addr` (I prefer short named, Addr, Res, more go style)

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -0,0 +1,345 @@
...
+
+// DefaultOpts is the recommended default options.
+var DefaultOpts = Opts{
+	Address:    0x18,
+	Resolution: Maximum,
+}
+
+// New opens a handle to an mcp9808 sensor.
+func New(bus i2c.Bus, opts *Opts) (*Dev, error) {
+
```
maruel: I personally do not put an empty line after the bracket.

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -0,0 +1,345 @@
...
+	mu       sync.Mutex
+	critical physic.Temperature
+	upper    physic.Temperature
+	lower    physic.Temperature
+	shutdown bool
+}
+
+// Sense reads the current ambient temperature.
+func (d *Dev) Sense() (physic.Temperature, error) {
```
maruel: I'd prefer if it matched the signature of https://periph.io/x/periph/conn/physic#SenseEnv

That said, I eventually want to move this interface out of physic for v4, and a SenseTemp() function is useful for people who just want to have the temperature. But the general function is useful IMHO.

Here's an example of implementing both; https://periph.io/x/periph/devices/ds18b20#Dev.LastTemp

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -0,0 +1,345 @@
...
+	t := physic.Temperature(tbits&0x0FFF) * 62500 * physic.MicroKelvin
+	if tbits&0x1000 > 0 {
+		// Check for bit sign.
+		t = -t
+	}
+	t += physic.ZeroCelsius
+	return t, nil
+}
+
+// SenseWithAlerts reads the ambient temperature and returns an slice of any
```
maruel: I'm not sure "ambient" is relevant here, the sensor may be strapped on a heater or something. :)

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -0,0 +1,345 @@
...
+		}
+	} else {
+		return 0, nil, errAlertInvalid
+	}
+
+	if err := d.enable(); err != nil {
+		return 0, nil, err
+	}
+
+	tbits, err := d.m.ReadUint16(temperature)
```
maruel: I'd prefer if this block and Sense() were deduped in an utility function.

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -0,0 +1,345 @@
...
+	t := physic.Temperature(b&0x03FF) * 250 * physic.MilliKelvin
+	if b&0x400 > 0 {
+		t = -t
+	}
+	t += physic.ZeroCelsius
+	return t
+}
+
+func alertTemperatureToBits(t physic.Temperature) (uint16, error) {
+	const (
```
maruel: I'd personally prefer:
```
const maxAlert = ..
const minAlert = ...
```

Its 50% the number of lines. :)

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -0,0 +1,345 @@
...
+		maxAlert = 125*physic.Kelvin + physic.ZeroCelsius
+		minAlert = -40*physic.Kelvin + physic.ZeroCelsius
+	)
+	if t >= maxAlert || t <= minAlert {
```
maruel: `>=` or `>`?

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -0,0 +1,345 @@
...
+		return 0, errAlertOutOfRange
+	}
+	t -= physic.ZeroCelsius
+	t /= 250 * physic.MilliKelvin
```
maruel: You could round with:
```
t = (t + 125*physic.MilliKelvin) / (250*physic.MilliKelvin)
```

albeit that doesn't work for negative.

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -0,0 +1,345 @@
...
+type resolution uint8
+
+const (
+	Maximum resolution = 0
+	Low     resolution = 1
+	Medium  resolution = 2
+	High    resolution = 3
+)
+
+var _ conn.Resource = &Dev{}
```
maruel:
```
var _ physic.SenseEnv = &Dev{}
```

experimental/devices/mcp9808/mcp9808smoketest/mcp9808smoketest.go
```diff
@@ -0,0 +1,76 @@
...
+	"flag"
+	"fmt"
+
+	"periph.io/x/periph/conn/i2c/i2creg"
+	"periph.io/x/periph/experimental/devices/mcp9808"
+	"periph.io/x/periph/host"
+)
+
+// SmokeTest is imported by periph-smoketest.
+type SmokeTest struct {
```
maruel: For it to be used you need to modify https://github.com/google/periph/blob/master/cmd/periph-smoketest/main.go

experimental/devices/mcp9808/mcp9808.go
NeuralSpaz: good question. I don't know The possible value in the register is much larger, but the specs on the chip are -40 to 125. so ?

experimental/devices/mcp9808/mcp9808.go
NeuralSpaz: Rounding does not matter as its 0.25C per a bit.

experimental/devices/mcp9808/mcp9808.go
maruel: I'd say to use `>` and `<` then. In practice, the chip can probably use much wider range, but they spec it to this so they can sell the "military range" spec at high multiple markup price.

experimental/devices/mcp9808/mcp9808.go
NeuralSpaz: I had not thought of implementing SenseEnv. I'm not sure about this.

experimental/cmd/mcp9808/main.go
NeuralSpaz: go 1.5 compatible? We'll find out.

experimental/devices/mcp9808/example_test.go
NeuralSpaz: always use go imports, don't think I have ever used the -w switch

experimental/cmd/mcp9808/main.go
NeuralSpaz: Personally I normally do stuff on the tick of the clock. So I don't have any good clear solutions ready to go.

experimental/devices/mcp9808/mcp9808.go
NeuralSpaz: Done and tested.

NeuralSpaz:
what do you think of 
```go
func (d *Dev) SenseWithAlerts(lower, upper, critical physic.Temperature) (physic.Temperature, []Alert, error) 
```
being 
```go
func (d *Dev) SenseWithAlerts(lower, upper, critical Alert) (physic.Temperature, []Alert, error) 
```
As that why people could define their own alerts names and with a suitable stringer could be rather nice?

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -0,0 +1,426 @@
...
+
+// Opts holds the configuration options.
+//
+// Slave Address
+//
+// Depending which pins the A0, A1 and A2 pins are connected to will change the
+// slave address. Default configuration is address 0x18 (Ax pins to GND). For a
+// full address table see datasheet.
+type Opts struct {
+	Addr int
```
maruel: uint16 ?

NeuralSpaz: Oh, the only reason I do that that way is for the flag be it int or uint since there is no uint16 flag. If its int I can bounds check in the New() can convert to uint16 there. Just makes it cleaner user code imo.
I can change it if you think it better.

NeuralSpaz:
```go
address := flag.Int("addr", 0x18, "I²C address")
```

```go
// Create a new temperature sensor a sense with default options.
sensor, err := mcp9808.New(bus, &mcp9808.Opts{Addr: *addr})
if err != nil {
	return fmt.Errorf("failed to open new sensor: %v", err)
}
```
rather than 
```go
sensor, err := mcp9808.New(bus, &mcp9808.Opts{Addr: uint16(*addr)})
if err != nil {
	return fmt.Errorf("failed to open new sensor: %v", err)
}
```

maruel: I started thinking about adding a i2c.Addr type that would implement Set(). Then it'd be easy to use as a flag.

Yes, I prefer uint16 for i2c address.

PR #357, opened 2018-12-19
ads1x15: add cmd for ads1015 AD converter with voltage divider calculation

lnitram:
Add a simple cmd to show the use of the ADS1015 AD converter. I added calculation for a voltage divider which is used for example in the automation-hat. The code is taken mainly from the example_test.go but for my development it is useful to have a main.


maruel:
gohci

Issue #360, opened 2018-12-20
i2c: add type Addr


maruel:
The main rationale is that we could implement `i2c.Addr.Set(s string) error` method, which would implement `flag.Var` interface, which would reduce the amount of boiler plate in the executables in cmd/... In the meantime it's often specified as uint16 but it's annoying to process.

NeuralSpaz:
the code would be straight forward enough for the flag.Value 
```go
// Addr implements flag.Value
type Addr uint16

func (a *Addr) Set(s string) error {
	u, err := strconv.ParseUint(s, 0, strconv.IntSize)
	if err != nil {
		return errI2CSetError
	}
	if u > 0xffff {
		return errI2CSetError
	}
	*a = Addr(u)
	return nil
}
```
I presume we don't want to break api anywhere, so this would have to be an optional connivance, If maintainers wanted to accept and i2c.Addr into the New funcs for example.

NeuralSpaz:
closing since done by #362

PR #362, opened 2018-12-20
i2c: Adds i2c.Addr type and flag.Value interface.

NeuralSpaz:
This is adds a new type i2c.Addr and Set methods that implement the flag.Value interface as a convenience for device maintainers.

Add i2c.Addr
Add i2c.Addr Set()
Add Tests.

helps with #360

### DIFF
--- a/conn/i2c/i2c.go
+++ b/conn/i2c/i2c.go
@@ -110,11 +110,11 @@ func (d *Dev) Duplex() conn.Duplex {
 	return conn.Half
 }
 
-// Addr is an I²C slave address. Addr implements the flag.Value interface.
+// Addr is an I²C slave address.
 type Addr uint16
 
 // Set sets the Addr to a value represented by the string s. Values maybe in
-// decimal or hexadecimal form.
+// decimal or hexadecimal form. Set implements the flag.Value interface.
 func (a *Addr) Set(s string) error {
 	u, err := strconv.ParseUint(s, 0, 16)
 	if err != nil {
--- a/conn/i2c/i2c.go
+++ b/conn/i2c/i2c.go
@@ -116,8 +116,8 @@ type Addr uint16
 // Set sets the Addr to a value represented by the string s. Values maybe in
 // decimal or hexadecimal form. Set implements the flag.Value interface.
 func (a *Addr) Set(s string) error {
-	// Allow for only maximum of 11 bits for i2c addresses.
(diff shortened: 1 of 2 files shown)


conn/i2c/i2c.go
```diff
@@ -109,6 +110,20 @@ func (d *Dev) Duplex() conn.Duplex {
 	return conn.Half
 }
 
-//
+// Addr implements flag.Value.
```
maruel: It should state first what it represents. :)

conn/i2c/i2c.go
```diff
@@ -109,6 +110,20 @@ func (d *Dev) Duplex() conn.Duplex {
...
+type Addr uint16
+
+// Set sets the Addr to a value respresented by the string s. Values maybe in
```
maruel: It is Set that implements flag.Value.

represented*

conn/i2c/i2c.go
```diff
@@ -109,6 +110,20 @@ func (d *Dev) Duplex() conn.Duplex {
...
+// decimal or hexadecimal form.
+func (a *Addr) Set(s string) error {
+	u, err := strconv.ParseUint(s, 0, 16)
+	if err != nil {
+		return errI2CSetError
+	}
+	*a = Addr(u)
```
maruel: Valid addresses are on 11 bits (when including R/W bit); so values above 0x7ff are invalid.
https://www.totalphase.com/support/articles/200349176-7-bit-8-bit-and-10-bit-I2C-Slave-Addressing gives a nice explanation.

NeuralSpaz:
no problem, Iv'e never used or even seen anything that uses 10 bit addressing, its one of those elusive parts of the spec, that most people will never use.

conn/i2c/i2c.go
```diff
@@ -109,6 +110,21 @@ func (d *Dev) Duplex() conn.Duplex {
...
 
-//
+// Addr is an I²C slave address.
+type Addr uint16
+
+// Set sets the Addr to a value represented by the string s. Values maybe in
+// decimal or hexadecimal form. Set implements the flag.Value interface.
+func (a *Addr) Set(s string) error {
+	// Allow for only maximum of 10 bits for i2c addresses.
+	u, err := strconv.ParseUint(s, 0, 10)
```
maruel: I hadn't realized you could use 10 bits for ParseUint!

PR #373, opened 2019-01-12
experimental: new device driver - CCS811

DeziderMesko:


DeziderMesko:
Mismatched repos...

experimental/cmd/ccs811/main.go
```diff
@@ -0,0 +1,158 @@
+package main
```
maruel: Please add consistent copyright headers.

experimental/cmd/ccs811/main.go
```diff
@@ -0,0 +1,158 @@
...
+
+func main() {
+
+	// Make sure periph is initialized.
+	if _, err := host.Init(); err != nil {
+		log.Fatal(err)
+	}
+
+	// Use i2creg I²C bus registry to find the first available I²C bus.
+	b, err := i2creg.Open("")
```
maruel: Please use the flag package, and add flags for the I²C bus selection and verbosity. Don't always open the default i2c bus. See other tools for examples.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	"fmt"
+	"time"
+
+	"periph.io/x/periph/conn"
+	"periph.io/x/periph/conn/i2c"
+)
+
+// Different measurement modes constants
+const (
+	MeasurementModeIdle         byte = 0
```
maruel: Probably worth making a type too.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+)
+
+// SensorErrorID represents error reported by the sensor
+type SensorErrorID byte
+
+/*
+Error constants, applicable if status registers signals error
+0
+WRITE_REG_INVALID
+The CCS811 received an I2C write request addressed to this station
```
maruel: Move this comment aside the constant below.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	ReadRegInvalid  SensorErrorID = 0x2
+	MeasModeInvalid SensorErrorID = 0x4
+	MaxResistance   SensorErrorID = 0x8
+	HeaterFault     SensorErrorID = 0x10
+	HeaterSupply    SensorErrorID = 0x20
+)
+
+// Opts holds the configuration options.
+type Opts struct {
+	Addr               uint16
```
maruel: Please state the valid values here as a comment.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+		return nil, fmt.Errorf("Invalid device address, only 0x5A or 0x5B are allowed")
+	}
+
+	if opts.MeasurementMode > 4 {
+		return nil, fmt.Errorf("Invalid measurement mode")
+	}
+
+	dev := &Dev{
+		c:    &i2c.Dev{Bus: bus, Addr: opts.Addr},
+		opts: opts,
```
maruel: That's dangerous, the user could keep a pointer to the opts and change it without your knowledge. Make line 112 a value, not a pointer.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+
+	// from boot mode to measurement mode
+	err := dev.StartSensorApp()
```
maruel: Embed into next line.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Error transitioning from boot do app mode: %v", err)
+	}
+
+	time.Sleep(20 * time.Millisecond)
```
maruel: Is that a requirement by the device datasheet?

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+
+	dev.SetMeasurementMode(MeasurementModeConstant1000, opts.InterruptWhenReady, opts.UseThreshold)
```
maruel: Why not use the mode requested by the user?

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	environmentReg     byte = 0x05
+	baselineReg        byte = 0x11
+	resetReg           byte = 0xFF
+)
+
+func (d *Dev) String() string {
+	return "CCS811"
+}
+
+// StartSensorApp initializes sensor to application mode
```
maruel: Try to end the comments with a period.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+func (d *Dev) StartSensorApp() error {
+	err := d.c.Tx([]byte{0xf4}, nil)
+	if err != nil {
```
maruel: lines 132-135 are unnecessary.

experimental/devices/ccs811/ccs811.go
maruel: This function returns an error that you ignore.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	if generateInterrupt {
+		mesModeValue = mesModeValue | (0x1 << 3)
+	}
+	if useThreshold {
+		mesModeValue = mesModeValue | (0x1 << 2)
+	}
+
+	// set measurement mode
+	err := d.c.Tx([]byte{measurementModeReg, mesModeValue}, nil)
+	if err != nil {
```
maruel: same

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+		return err
+	}
+	return nil
+}
+
+type MeasurementMode struct {
```
maruel: Don't forget to document all public symbols.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	MeasurementMode   byte
+	GenerateInterrupt bool
+	UseThreshold      bool
+}
+
+func (d *Dev) GetMeasurementMode() (mm MeasurementMode, err error) {
```
maruel: Don't use named return values on all functions. The only exception is when a return value is needed in a defered function.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	r := make([]byte, 1)
+	err = d.c.Tx([]byte{measurementModeReg}, r)
```
maruel: Inline here and everywhere applicable.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+func (d *Dev) ReadStatus() (byte, error) {
+	r := make([]byte, 1)
+	if err := d.c.Tx([]byte{statusReg}, r); err != nil {
+		return 0, err
+	}
+	return r[0], nil
+}
+
+// ReadRawData provides current and voltage on the sensor
+// current in uA
```
maruel: Weird comment alignment.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+// voltage from 0-1023, where 1023 = 1.65V
+func (d *Dev) ReadRawData() (current int, voltage int, err error) {
```
maruel: Use types from package https://periph.io/x/periph/conn/physic.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	if err = d.c.Tx([]byte{rawDataReg}, r); err != nil {
+		return 0, 0, err
+	}
+	current, voltage = valuesFromRawData(r)
+	return
+}
+
+// SetEnvironmentData allows to provide temperature and humidity so
+// sensor can compensate it's measurement
+func (d *Dev) SetEnvironmentData(temp, humidity float32) error {
```
maruel: Prefer implementing https://periph.io/x/periph/conn/physic#SenseEnv.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+		byte(rawTemp >> 8),
+		byte(rawTemp)}
+
+	if err := d.c.Tx(w, nil); err != nil {
+		return err
+	}
+	return nil
+}
+
+// GetBaseline provides current baseline used by measurement alogrithm
```
maruel: "algorithm"
but really, that doesn't tell the user much. :) I'd recommend to say "See SetBaseline for the choice of algorithms" or something like that.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+func (d *Dev) GetBaseline() ([]byte, error) {
+	r := make([]byte, 2)
+	if err := d.c.Tx([]byte{baselineReg}, r); err != nil {
+		return r, err
```
maruel: Don't return `r` here.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+	return r, nil
+}
+
+/*
```
maruel: Prefer // comments consistently.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+saved baseline value using the BASELINE register. The correct time to save the baseline
+will depend on the customer use-case and application.
+
+• For devices which are powered for >24 hours at a time:
+• During the first 500 hours – save the baseline every
+24-48 hours.
+• After the first 500 hours – save the baseline every 5-7 days.
+
+• For devices which are powered <24 hours at a time:
+• If the device is run in, save the baseline before power down
```
maruel: Please double check how it looks in godoc and tweak. You can run godoc locally with the command `godoc -http=:6060`

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	r := make([]byte, 2)
+	if err := d.c.Tx([]byte{baselineReg}, r); err != nil {
+		return r, err
+	}
+	return r, nil
+}
+
+/*
+SetBaseline sets current baseline for measurement algorithm.
+For mor detail check sensor's specification
```
maruel: more
period

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+
+Manual Baseline Correction
```
maruel: You can create a title by surrounding with empty lines.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+There is a mechanism within CCS811 to manually save and restore a previously
+saved baseline value using the BASELINE register. The correct time to save the baseline
+will depend on the customer use-case and application.
+
+• For devices which are powered for >24 hours at a time:
```
maruel: I wonder if this should be package documentation instead.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+
+• For devices which are powered <24 hours at a time:
+• If the device is run in, save the baseline before power down
+• If multiple operating modes are used, a separate baseline should be stored for each
+• The baseline should only be restored when the resistance is stable (typically 20-30 minutes)
+• If changing from a low to high power mode (without spending at least 10 minutes in idle),
+  the sensor resistance should be allowed to settle again before restoring the baseline
+
+Note(s):
+1) If a value is written to the BASELINE register while the sensor is stabilising, the output of the TVOC and eCO2 calculations may be higher than expected.
```
maruel: Try to wrap documentation at 80 cols, but it's not an hard request.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+// SensorValues represents data read from the sensor.
+// Data are populated based on NeededData parameter.
+type SensorValues struct {
+	ECO2           int
+	VOC            int
+	Status         byte
+	ErrorID        SensorErrorID
+	RawDataCurrent int // current in uA
```
maruel: Use types from physic.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	RawDataVoltage int // voltage from 0-1023, where 1023 = 1.65V
+}
+
+// Sense provides data from the sensor.
+// ReadData parameter specifies which data should be read.
+func (d *Dev) Sense(mode NeededData) (*SensorValues, error) {
```
maruel: make SensorValues a parameter instead of a return value. This way there's no heap allocation on the success case.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+
+// NeededData represents set of data read from the sensor
+type NeededData byte
+
+// What data should be read from the sensor
+const (
+	ReadCO2          NeededData = 2
+	ReadCO2VOC       NeededData = 4
+	ReadCO2VOCStatus NeededData = 5
+	ReadAll          NeededData = 8
```
maruel: Frankly, I'd always read all and skip this flag altogether. It's not worth bothering the user with this, and the difference of 2 bytes vs 8 bytes read after the setup time, the address and the command being written is fairly low.

Not only this simplifies the driver, but it also simplifies the life of the user.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+type FwVersions struct {
+	HWIdentifier       byte
+	HWVersion          byte
+	BootVersion        string
+	ApplicationVersion string
+}
+
+func (d *Dev) GetFirmwareData() (version *FwVersions, err error) {
+	version = &FwVersions{}
+	hwid := make([]byte, 1)
```
maruel: ~~You don't need the next 3 transactions; do one starting at 0x20 reading 5 bytes and use these.~~

(me reads the datasheet)

Wow this device is insane. Maybe even worth a comment specifically to note that while this device _looks_ like a memory mapped registers, it's not.

experimental/devices/ccs811/ccs811_test.go
```diff
@@ -0,0 +1,140 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
```
maruel: Update year.

experimental/devices/ccs811/ccs811_test.go
```diff
@@ -0,0 +1,140 @@
...
+package ccs811
+
+import (
+	"testing"
+
+	"periph.io/x/periph/conn/i2c/i2ctest"
+)
+
+func TestBasicInitialisationAndDataRead(t *testing.T) {
+	// FIXME: Try to include basic code coverage. You can use "replay" tests by
```
maruel: Remove the comment.

experimental/devices/ccs811/doc.go
```diff
@@ -0,0 +1,12 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
```
maruel: Same.

experimental/devices/ccs811/doc.go
```diff
@@ -0,0 +1,12 @@
...
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package CCS811 controls CCS811 Volatile Organic Compounds sensor via
+// I²C interface.
+//
+// Datasheet
+//
+// https://ams.com/ccs811
+
```
maruel: This won't show as package documentation unless you remove this line. See https://blog.golang.org/godoc-documenting-go-code for more background.

experimental/devices/ccs811/doc.go
```diff
@@ -0,0 +1,12 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package CCS811 controls CCS811 Volatile Organic Compounds sensor via
+// I²C interface.
+//
+// Datasheet
+//
+// https://ams.com/ccs811
```
maruel: That's the product page, the datasheet is at https://ams.com/documents/20143/36005/CCS811_DS000459_6-00.pdf. You can put both.

experimental/devices/ccs811/ccs811.go
DeziderMesko: You are right, 20ms is for power on. For APP_START its just 1ms. I'm going to remove it completely.

experimental/devices/ccs811/ccs811.go
DeziderMesko: It's related to internal algorihtm of changing voltage and current of sensor to VOC/eCO2. User doesn't have chance to influence it. I'll try to be more clear.

experimental/devices/ccs811/ccs811.go
```diff
@@ -0,0 +1,361 @@
...
+	ReadRegInvalid  SensorErrorID = 0x2
+	MeasModeInvalid SensorErrorID = 0x4
+	MaxResistance   SensorErrorID = 0x8
+	HeaterFault     SensorErrorID = 0x10
+	HeaterSupply    SensorErrorID = 0x20
+)
+
+// Opts holds the configuration options. The address must be 0x5A or 0x5B.
+// MeasurementMode:
+// -  Mode 0: Idle, low current mode.
```
maruel: These belongs to the actual constants along lines 19-23.

experimental/devices/ccs811/ccs811.go
DeziderMesko: I'm keeping both, sense is ReadAll, sensePartial is old version where user for some reason can read just something

experimental/devices/ccs811/ccs811.go
DeziderMesko: I'm not very good in sensors internals, so if you have some specific idea (paragraph) what to put there, let me know, or add it anytime later

experimental/devices/ccs811/ccs811.go
DeziderMesko: I'm not sure, this is excerpt from datasheet, I wouldn't go to package to look for this info :-) So for lazy programmers I would keep it there. Keen programmers will go to datasheet anyway.

experimental/cmd/ccs811/main.go
```diff
@@ -18,31 +19,62 @@ import (
...
+	baseline := flag.Bool("baseline", false, "command displays value used for correction of measurement")
+	sense := flag.Bool("sense", false, "command performs one time measurement")
+	readContinuously := flag.Bool("readcontinuously", false, "command performs continous measuremnt with interval of one second")
+	fwInfo := flag.Bool("fwinfo", false, "command displays different versions of hardware, boot and firmware")
+	appStart := flag.Bool("appstart", false, "command starts sensor's application - move it from boot to application mode")
+	printMeasureMode := flag.Bool("printmeasuremode", false, "command shows current measurement mode")
+	setMeasureMode := flag.Uint("setmeasuremode", 1, "command sets mode, valid values are 0-4, default is periodic 1s reading")
+	flag.Parse()
+
+	numberOfCommandFlags := 0
```
maruel: Oh I was fine to use flag.Args()[0] as the command, sorry I hadn't been clear.

maruel:
gohci

experimental/cmd/ccs811/main.go
DeziderMesko: Well in this particular case, flag package sucks. I can't creat mutually exclusive commands (and I dont want to have -cmd <list of 10 possible parameters>), and to have same command in read and write mode just adds complexity. So I've split it.

maruel: What I meant is to not specify them as flags, and instead look at the value of `flags.Args()[0]`. You can override the help page with `flag.Usage = func() {...}`.

maruel:
I immediately forcibly removed the commit, because you had committed a binary. Please redo the PR, you can reuse commit 523c0c4 if you can fetch it, and remove the binary from it.

DeziderMesko:
Damn, sorry. I'll check it after work.

PR #374, opened 2019-01-13
i2c: Add Stringer for i2c.Addr

NeuralSpaz:
Add Addr.String()
Add Flag example

i2c.Addr was missing String() to complete flag.Value interface.

### DIFF
--- a/conn/i2c/i2c.go
+++ b/conn/i2c/i2c.go
@@ -125,6 +125,11 @@ func (a *Addr) Set(s string) error {
 	return nil
 }
 
+// String returns the i2c.Addr as a string formated in hex.
+func (a Addr) String() string {
+	return "0x" + strconv.FormatInt(int64(a), 16)
+}
+
 var errI2CSetError = errors.New("invalid i2c address")
 
 var _ conn.Conn = &Dev{}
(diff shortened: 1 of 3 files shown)


conn/i2c/i2c.go
```diff
@@ -125,6 +125,11 @@ func (a *Addr) Set(s string) error {
 	return nil
 }
 
+// String returns s i2c.Addr as a string formated in hexadecimal.
```
maruel: returns

conn/i2c/i2c_test.go
```diff
@@ -128,3 +128,21 @@ func TestAddr_Set(t *testing.T) {
...
+		Addr Addr
+		want string
+	}{
+		{0x01, "0x1"},
+		{0x24, "0x24"},
+		{0x3ff, "0x3ff"},
+	}
+
+	for _, tt := range tests {
+		got := tt.Addr.String()
```
maruel: This can be inlined in the next line as `if got := tt.Addr.String(); got != tt.want {`

conn/i2c/i2c.go
maruel: a :)

NeuralSpaz: an :)

PR #376, opened 2019-01-13
travis.yml: use travis condition

maruel:
Experimenting, not sure it'll work. Even less as a PR.


maruel:
Didn't do what I expected.

PR #385, opened 2019-01-19
experimental: new device driver - CCS811

DeziderMesko:
Notes

Please prefix the issue title with the primary package affected. For example, if
this PR fixes periph.io/x/periph/host/sysfs, prefix the PR title with "sysfs:".

Please add new drivers under "experimental". Wonder what it takes to promote a
driver as "stable"? See https://periph.io/project/#driver-lifetime-management
A stable driver requires the smallest API surface, good unit test code coverage,
good documentation and a page in
https://github.com/periph/website/tree/master/site/content/device so it can show
up at https://periph.io/device.

Mention the issue number it fixes or add the details of the changes if it
doesn't have a specific issue.

Examples:

Fixes #12345
Helps with #12345 but doesn't not completely fix it.

### DIFF
--- a/experimental/cmd/ccs811/main.go
+++ b/experimental/cmd/ccs811/main.go
@@ -1,0 +1,202 @@
+// Copyright 2019 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package main
+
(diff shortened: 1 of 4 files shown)


maruel:
gohci

maruel:
Thanks!

PR #390, opened 2019-01-27
experimental: Adds device driver for Inky pHAT e-ink screens

hatstand:
Only supports pHAT for now, but wHAT support should be simple to add.


experimental/cmd/inky/main.go
```diff
@@ -0,0 +1,49 @@
+package main
```
maruel: I'd recommend to leave this file out for now, including the png.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
```
maruel: We're now in 2019. :)

experimental/devices/inky/lut.go
```diff
@@ -0,0 +1,222 @@
+package inky
```
maruel: Please include copyright

experimental/devices/inky/lut.go
```diff
@@ -0,0 +1,222 @@
...
+
+var (
+	blackLUT = []byte{
```
maruel: Use [...] form so it is an array instead of a slice. See https://blog.golang.org/go-slices-usage-and-internals for more info.

experimental/devices/inky/doc.go
```diff
@@ -0,0 +1,12 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
```
maruel: We're now in 2019. :)

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+
+	d.busy.In(gpio.PullUp, gpio.FallingEdge)
+	defer d.busy.In(gpio.PullUp, gpio.NoEdge)
+	d.sendCommand(0x12, nil) // Soft Reset
+	d.busy.WaitForEdge(-1)
+}
+
+func (d *Dev) sendCommand(command byte, data []byte) error {
+	d.dc.Out(spiCommand)
+	err := d.c.Tx([]byte{command}, nil)
```
maruel: You can inline this code; so it looks like
```
if err := d.c.Tx([]byte{command}, nil); err != nil {
```

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to send data for command %x to inky: %v", command, err)
+		}
+	}
+	return nil
+}
+
+func (d *Dev) sendData(data []byte) error {
+	if len(data) > 4096 {
+		log.Fatalf("Sending more data than chunk size")
```
maruel: Return an error instead.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	d.dc.Out(spiData)
```
maruel: Repeat of comments as above here and next line.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+		return fmt.Errorf("Partial update not supported")
+	}
+
+	if src.Bounds() != d.Bounds() {
+		return fmt.Errorf("Image must be the same size as bounds: %v", d.Bounds())
+	}
+
+	b := src.Bounds()
+	// Black/white pixels.
+	white := make([]bool, rows*cols)
```
maruel: Sadly, []bool in Go is not optimized, so it's using 1 byte per bit, so a 8x blow up. It's not a show stopper here but would be nice to fix later. There is already an example of bitmap image format at https://periph.io/x/periph/devices/ssd1306/image1bit but it's not the right format.

You're not required to do this but one possibility would be to recreate a 2 bit image format that would be able to represent the 3 colors. Technically you could pack it to 1.5 bits per pixel, but that'd be more work than worth it. You already started with ColorModel() above. :)

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+	// Red/Transparent pixels.
+	red := make([]bool, rows*cols)
+	for x := b.Min.X; x < b.Max.X; x++ {
+		for y := b.Min.Y; y < b.Max.Y; y++ {
+			i := x*cols + y
+			srcX := x
+			srcY := b.Max.Y - y - 1
+			r, g, b, _ := d.ColorModel().Convert(src.At(srcX, srcY)).RGBA()
+			if r == 0xffff && g == 0xffff && b == 0xffff {
```
maruel: Generally you'd use >= 0x8000 or something like that instead of exact match for pure white.

experimental/devices/inky/lut.go
```diff
@@ -0,0 +1,222 @@
...
+		0x48,
```
maruel: Can you wrap a bit to make the file a bit less long? 16 per line would be great.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+
+const (
+	// Constants for an Inky pHAT
+	cols = 104
+	rows = 212
+
+	speed      = 488 * physic.KiloHertz
+	spiBits    = 8
+	chunkSize  = 4096
+	spiCommand = gpio.Low
```
maruel: Personally I'd just inline these two.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+)
+
+const (
+	// Constants for an Inky pHAT
+	cols = 104
+	rows = 212
+
+	speed      = 488 * physic.KiloHertz
+	spiBits    = 8
+	chunkSize  = 4096
```
maruel: Remove, it's not used.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+	"periph.io/x/periph/conn/spi"
+)
+
+const (
+	// Constants for an Inky pHAT
+	cols = 104
+	rows = 212
+
+	speed      = 488 * physic.KiloHertz
+	spiBits    = 8
```
maruel: Please inline.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+	"periph.io/x/periph/conn/physic"
+	"periph.io/x/periph/conn/spi"
+)
+
+const (
+	// Constants for an Inky pHAT
+	cols = 104
+	rows = 212
+
+	speed      = 488 * physic.KiloHertz
```
maruel: Please inline. The speed is really odd though. :)

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+	spiBits    = 8
+	chunkSize  = 4096
+	spiCommand = gpio.Low
+	spiData    = gpio.High
+)
+
+type Color int
```
maruel: Please add a comment about the fact that it can be used to specify which model is being used, and also to set the border color.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+
+var borderColor = map[Color]byte{
+	Black:  0x00,
+	Red:    0x33,
+	Yellow: 0x33,
+	White:  0xff,
+}
+
+// NewpHAT opens a handle to an Inky pHAT.
+func NewpHAT(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinIn, color Color) (*Dev, error) {
```
maruel: Since it's very much not configurable, I wonder if the pins should be specified as arguments or not. I don't know. Normally devices are not using the registry but I can make an exception here since this device is really meant to be used with SPI0 and specific GPIOs.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+		}
+	}
+
+	bufA, _ := pack(white)
+	bufB, _ := pack(red)
+	return d.update(borderColor[d.border], bufA, bufB)
+}
+
+func (d *Dev) update(border byte, black []byte, red []byte) error {
+	d.reset()
```
maruel: Please check for errors everywhere, would it be for consistency. And yes reset() should return an error.

I understand the code below will become a mess though, in theory some of it could be moved into a table + a loop, where you create a table and then iterate over the commands in it, i.e.

```
init := []struct{
  cmd byte
  data[]byte
}{
  {0x74, []byte{0x54}},
  {0x7e, []byte{0x3b}},
  ...
}
```

That said I don't want you to spend too much time on this, so it's not a big deal.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,291 @@
...
+				white[i] = true
+			} else if r == 0xffff {
+				// Red pixels also need white behind them.
+				white[i] = true
+				red[i] = true
+			}
+		}
+	}
+
+	bufA, _ := pack(white)
```
maruel: Not required, but personally I'd recommend to let go the []bool and immediately work in packed data instead. It would look a bit like this:

```
index := x / 8 + y * b.Max.X
shift := 7 - uint(x)%8
ret[index] |= 1 << shift
```

but you only need to do these for the `true` assignments, removing the need for boolToByte.

experimental/devices/inky/inky.go
hatstand: Yeah, I couldn't decide here either. It would technically be possible to hook it up to different pins if you really wanted to but I'm not sure anyone ever would?

experimental/devices/inky/inky.go
hatstand: I thought about it but decided the complexity wasn't worth it for now - the memory blowup is only ~350KB and I think the minimum RAM you can get on a Pi is 512MB these days. Might be worth it for a wHAT though (~2MB).

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,327 @@
...
+func pack(bits []bool) ([]byte, error) {
+	if len(bits)%8 != 0 {
+		return nil, fmt.Errorf("len(bits) must be multiple of 8 but is %d", len(bits))
+	}
+
+	ret := make([]byte, len(bits)/8)
+	for i, b := range bits {
+		index := i / 8
+		shift := 7 - uint(i)%8
+		ret[index] |= (boolToByte(b) << shift)
```
maruel: boolToByte is still not needed, since you don't need to set it when it's false, so the code can be simplified to:

```
if b {
  ret[i/8] |= 1 << (7 - uint(i)%8)
}
```

It's both faster and less complex to read.

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,327 @@
...
+	d.busy.WaitForEdge(-1)
+
+	if err := d.sendCommand(0x10, []byte{0x01}); err != nil { // Enter deep sleep.
+		return err
+	}
+	return nil
+}
+
+func (d *Dev) reset() error {
+	d.r.Out(gpio.Low)
```
maruel: Please consistently check for errors.

experimental/devices/inky/inky.go
maruel: Let's say we keep it this way, what I could recommend it to add an `example_test.go` (see other packages for example) which could look like:

```
// Make sure periph is initialized.
if _, err := host.Init(); err != nil {
  log.Fatal(err)
}
s, err := spireg.Open("SPI0.0")
if err != nil {
  log.Fatal(err)
}
dc := gpioreg.ByName("123")
if dc == nil {
  log.Fatal("expecting to find DC pin")
}
(same for reset, busy)
h, err := inky.NewpHAT(s, dc, reset, busy, Yellow)
if err != nil {
  log.Fatal(err)
}
(draw something simple)
```

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,327 @@
...
+var borderColor = map[Color]byte{
+	Black:  0x00,
+	Red:    0x33,
+	Yellow: 0x33,
+	White:  0xff,
+}
+
+// NewpHAT opens a handle to an Inky pHAT.
+func NewpHAT(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinIn, color Color) (*Dev, error) {
+	if color != Black && color != Red && color != Yellow {
```
maruel: I'm thinking we may group all the pimonori drivers into a subpackage. I'll think about it.

Another thing is that generally I prefer a `New()` function. Here do you plan to add another New function? Otherwise I'd prefer to rename.

experimental/devices/inky/inky.go
hatstand: Is it worth checking errors on every gpio write?

experimental/devices/inky/inky.go
hatstand: I called it `NewpHAT()` to leave room for a `NewwHAT()` later.

maruel: It looks like from the python code that only effective difference is the dimensions, so maybe pass this as an Opts struct? e.g.

```
// Opts is the options to specify which device is being controlled, and
// its default settings.
type Opts struct {
  // Model being used.
  Model Model
  // Model color used.
  ModelColor Color.
  // Initial border color. Will be set on the first Draw().
  Border Color
}

// Model lists the supported e-ink display models.
type Model int

const (
  PHAT Model = iota
  WHAT
)

// New opens a handle to an Pimonori Inky pHAT or wHAT.
func New(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinIn, o *Opts) (*Dev, error) {
  ...
}
```

wdyt?

experimental/devices/inky/inky.go
maruel: That's a fair question. In theory I'd say no. In practice, I do because of these reasons:

- periph's code is meant to be as the reference for high quality code.
- People tend to copy paste code, and as such, the code should do the right thing.
- You can't know what failure mode will happen.

Here it's a special case where we know the possibility of the hardware setup is fairly constrained, but people are more creative than that.

Imagine someone hacks up a setup with a FT232H (connected via USB) on their Windows workstation and connections up the wires to drive an inky from there. It's kinda cool when you think about it. There's 6 data wires to connect plus power. In this setup, if the user unplug the FT232H by accident, the functions will start failing at runtime.

experimental/devices/inky/inky.go
hatstand: Yes, that would work too.

experimental/devices/inky/inky.go
hatstand: That's fair. What do you think about the case where you want to do something like this?
```go
pin.In(gpio.PullUp, gpio.FallingEdge)
defer pin.In(gpio.PullUp, gpio.NoEdge)
```
I guess it's a bit like doing `defer file.Close()` on a writable file; i.e., you probably shouldn't use `defer` and maybe use a named return for the `error`?

maruel:
gohci

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,352 @@
...
+		{0x22, []byte{0xc7}},
+	}
+
+	for _, c := range write {
+		if err := d.sendCommand(c.cmd, c.data); err != nil {
+			return err
+		}
+	}
+
+	d.busy.In(gpio.PullUp, gpio.FallingEdge)
```
maruel: I'll commit as soon as gohci test passes but please do the same verification for this one in a follow up.

That said in practice I'm not sure it's really necessary to reset the edge detection here, you could do something like

```
// Ensure any buffered edge is flushed.
d.busy.WaitForEdge(0)

(do initialization)

if !d.busy.WaitForEdge(10*time.Second) {
  return errors.New("failed to initialize")
}
```

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,352 @@
...
+	}()
+	if err := d.sendCommand(0x12, nil); err != nil { // Soft Reset
+		return fmt.Errorf("failed to reset inky: %v", err)
+	}
+	d.busy.WaitForEdge(-1)
+	return
+}
+
+func (d *Dev) sendCommand(command byte, data []byte) error {
+	d.dc.Out(gpio.Low)
```
maruel: Here too

experimental/devices/inky/inky.go
```diff
@@ -0,0 +1,352 @@
...
+}
+
+func (d *Dev) sendData(data []byte) error {
+	if len(data) > 4096 {
+		return fmt.Errorf("Sending more data than chunk size: %d > 4096", len(data))
+	}
+	if err := d.dc.Out(gpio.High); err != nil {
+		return err
+	}
+	if err := d.c.Tx(data, nil); err != nil {
```
maruel: I wonder if you have to sleep a few µs before the device acknowledge the d.dc level. Without a datasheet means <shrug>.

PR #392, opened 2019-01-28
physic: add examples how to convert to float64

maruel:
No functional change, just new examples.

### DIFF
--- a/conn/physic/example_test.go
+++ b/conn/physic/example_test.go
@@ -53,6 +53,25 @@ func ExampleAngle_flag() {
 	flag.Parse()
 }
 
+func ExampleAngle_float64() {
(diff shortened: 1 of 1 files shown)


conn/physic/example_test.go
```diff
@@ -279,6 +373,20 @@ func ExampleEnergy_flag() {
 	flag.Parse()
 }
 
+func ExampleEnergy_float64() {
+	// BTU is used in barbecue rating. At 35000 BTU barbecue is an average power.
```
NeuralSpaz: BTU is used in barbecue rating. It is a measure of the thermal energy content of the of fuel consumed by the barbecue at its maximum rate. At 35000 BTU barbecue is an average power.

PR #395, opened 2019-01-29
Add go.mod to enable builds outside GOPATH

hatstand:


maruel:
Thanks, I hadn't taken a look at go mod yet. We should do the same for https://github.com/periph/extra

PR #396, opened 2019-02-04
fs: try to fix a deadlock in the new code

maruel:
Remove the lock from wakeUpLoop(). I suspect it is what was deadlocking
on travis.

Remove the tear down code. It was hard to make tear down never deadlock in unit
tests, so leak the goroutine instead.

Properly release the lock on initialization failure.

Make follow up init() call to be lockless.

### DIFF
--- a/host/fs/fs_linux.go
+++ b/host/fs/fs_linux.go
@@ -134,28 +134,26 @@ func (e *event) wait(timeoutms int) (int, error) {
 // One OS thread is needed for all the events. This is more efficient on single
 // core system.
 type eventsListener struct {
-	wg sync.WaitGroup
-
-	mu      sync.Mutex
-	closing bool
 	// file descriptor of the epoll handle itself.
 	epollFd int
 	// pipes to wake up the loop()
-	r, w *os.File
-	// mapping of file descriptors to wait on with their corresponding channels.
-	fds    map[int32]chan<- time.Time
+	r, w   *os.File
 	wakeUp <-chan time.Time
+
+	// Mapping of file descriptors to wait on with their corresponding channels.
+	mu  sync.Mutex
+	fds map[int32]chan<- time.Time
 }
 
 // init must be called on a fresh instance.
 func (e *eventsListener) init() error {
-	e.mu.Lock()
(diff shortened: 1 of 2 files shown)


host/fs/fs_linux.go
```diff
@@ -164,64 +164,34 @@ func (e *eventsListener) init() error {
...
 	e.wakeUp = wakeUp
-	// The mutex is still held after this function exits, it's loop() that will
-	// release the mutex.
+	e.fds = map[int32]chan<- time.Time{}
+	e.fds[int32(e.r.Fd())] = wakeUp
+	// The mutexes are still held after this function exits, it's loop() that will
+	// release the mutexes.
 	//
 	// This forces loop() to be started before addFd() can be called by users.
 	e.wg.Add(1)
```
NeuralSpaz: nothing waits for this wait group.

host/fs/fs_linux.go
```diff
@@ -164,64 +164,34 @@ func (e *eventsListener) init() error {
...
-	e.r = nil
-	err := syscall.Close(e.epollFd)
-	e.epollFd = 0
-	e.mu.Unlock()
-	return err
-}
-
 // loop is the main event loop.
 func (e *eventsListener) loop() {
 	defer e.wg.Done()
```
NeuralSpaz: same here, nothing is waiting.

host/fs/fs_linux.go
```diff
@@ -136,26 +136,26 @@ func (e *event) wait(timeoutms int) (int, error) {
 type eventsListener struct {
 	wg sync.WaitGroup
```
NeuralSpaz: no longer used.

host/fs/fs_linux.go
maruel: Thanks for catching this, removed the WaitGroup.

host/fs/fs_linux.go
maruel: Removed.

host/fs/fs_linux.go
maruel: Removed.

PR #397, opened 2019-02-04
sysfs: Move new epoll code from fs into sysfs

maruel:
I don't think it'll be used elsewhere after all. If it ever become the
case, we can revisit. But for now, it's better to not export
unnecessarily.

No functional change, just a move of unused code.


maruel:
I'm not sure why github doesn't present the file copy, `git diff -C -C origin/master` does present the file copy properly.

PR #398, opened 2019-02-04
sysfs: switch GPIO to use context.Context based epoll wait

maruel:
- Fixes Pin.Halt(), In() and Out() to unblock any pending WaitForEdge().
- Enable all of gpiosmoketest.
- Handle spurious edges deterministically.

This fixes issue 323.

I want to do more extensive testing on it before committing. Having a burn in test would be nice, albeit it only need to not be worse than the current implementation, it doesn't necessarily need to be perfect.

Also, I find the code too complicated, so I need to figure out a way to simplify it.


maruel:
Do not review yet since it's broken.

djensenius:
Any update on the state of this PR? Looks like it could also fix #423 ?

maruel:
Hi @djensenius!

Thanks for the ping. Yes it's still very definitely on my plate even if I haven't touched this for several months. :/

maruel:
Will rework for the new version that is go module compatible.

PR #399, opened 2019-02-06
as7262 add internal context.context

NeuralSpaz:
replaces some fragile time based tests with a with cancel context. No external api changes.

### DIFF
--- a/experimental/devices/as7262/as7262.go
+++ b/experimental/devices/as7262/as7262.go
@@ -39,12 +39,21 @@ func New(bus i2c.Bus, opts *Opts) (*Dev, error) {
 		gain:      opts.Gain,
 		interrupt: opts.InterruptPin,
 		cancel:    func() {},
+		done: func() <-chan struct{} {
+			c := make(chan struct{})
+			go func() {
+				select {
+				case c <- struct{}{}:
+				}
+			}()
+			return c
+		},
 	}, nil
 }
 
 var sensorTimeout = 200 * time.Millisecond
 
-// WaitForSensor is overridden in tests.
+// waitForSensor is overridden in tests.
 var waitForSensor = time.After
 
 // Dev is a handle to the as7262 sensor.
@@ -56,9 +65,10 @@ type Dev struct {
 	mu   sync.Mutex
 	gain Gain
 
-	// cancelMu guards just cancelFunc
+	// cancelMu guards cancel and done.
 	cancelMu sync.Mutex
 	cancel   context.CancelFunc
+	done     func() <-chan struct{}
 }
 
 // Spectrum is the reading from the sensor including the actual sensor state for
@@ -132,6 +142,7 @@ func (d *Dev) Sense(ledDrive physic.ElectricCurrent, senseTime time.Duration) (S
 	ctx, cancel := context.WithCancel(context.Background())
 	d.cancelMu.Lock()
 	d.cancel = cancel
+	d.done = ctx.Done
 	d.cancelMu.Unlock()
 	defer d.cancel()
 
@@ -229,6 +240,12 @@ func (d *Dev) Halt() error {
 	d.cancelMu.Lock()
 	defer d.cancelMu.Unlock()
 	d.cancel()
+	select {
+	case <-d.done():
+		// Halted.
+	case <-time.After(time.Second):
+		return errors.New("halt timeout")
+	}
 	return nil
 }
 
--- a/experimental/devices/as7262/as7262.go
+++ b/experimental/devices/as7262/as7262.go
@@ -40,13 +40,7 @@ func New(bus i2c.Bus, opts *Opts) (*Dev, error) {
 		interrupt: opts.InterruptPin,
 		cancel:    func() {},
 		done: func() <-chan struct{} {
-			c := make(chan struct{})
-			go func() {
-				select {
-				case c <- struct{}{}:
-				}
-			}()
-			return c
+			return make(chan struct{})
 		},
 	}, nil
 }
@@ -146,7 +140,7 @@ func (d *Dev) Sense(ledDrive physic.ElectricCurrent, senseTime time.Duration) (S
 	d.cancelMu.Unlock()
 	defer d.cancel()
 
-	if err := d.writeVirtualRegister(ctx, intergrationReg, it); err != nil {
+	if err := d.writeVirtualRegister(ctx, integrationReg, it); err != nil {
 		return Spectrum{}, err
 	}
 	led, drive := calcLed(ledDrive)
@@ -244,7 +238,8 @@ func (d *Dev) Halt() error {
 	case <-d.done():
 		// Halted.
 	case <-time.After(time.Second):
-		return errors.New("halt timeout")
+		// Prevent Halt() from hanging forever if Sense() has not been called.
+		return errHaltTimeout
 	}
 	return nil
 }
@@ -552,6 +547,7 @@ var (
 	errPinTimeout     = errors.New("timeout waiting for interrupt signal on pin")
 	errHalted         = errors.New("received halt command")
 	errGainValue      = errors.New("invalid gain value")
+	errHaltTimeout    = errors.New("halt timeout, sensor may not have been started")
 )
 
 const (
@@ -561,7 +557,7 @@ const (
 	hardwareVersion      = 0x00
 	firmwareVersion      = 0x02
 	controlReg           = 0x04
-	intergrationReg      = 0x05
+	integrationReg       = 0x05
 	deviceTemperatureReg = 0x06
 	ledControlReg        = 0x07
 	// RawBase used as base for reading uint16 values, data must be sequentially.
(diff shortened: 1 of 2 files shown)


experimental/devices/as7262/as7262_test.go
```diff
@@ -16,14 +17,12 @@ import (
 	"periph.io/x/periph/conn/physic"
 )
 
+var defaultSensorTimeOut = time.Millisecond * 200
```
maruel: I think (?) you could do

var defaultSensorTimeOut = sensorTimeout

experimental/devices/as7262/as7262.go
```diff
@@ -37,25 +38,27 @@ func New(bus i2c.Bus, opts *Opts) (*Dev, error) {
...
 		interrupt: opts.InterruptPin,
-		done:      make(chan struct{}, 1),
-		timeout:   200 * time.Millisecond,
+		cancel:    func() {},
 	}, nil
 }
 
+var sensorTimeout = 200 * time.Millisecond
+
+// WaitForSensor is overridden in tests.
```
maruel: lower case w

experimental/devices/as7262/as7262_test.go
```diff
@@ -941,19 +922,19 @@ func TestIOError_Error(t *testing.T) {
...
-	d.Halt()
-	if _, err := d.Sense(physic.MilliAmpere*100, time.Millisecond*3); err == errHalted {
-		t.Errorf("got %v expected nil", errHalted)
-	}
-	d.Halt()
-	if err := d.Gain(G1x); err == errHalted {
-		t.Errorf("got %v expected nil", errHalted)
-	}
-}
+// func TestIntergration_AfterHalt(t *testing.T) {
```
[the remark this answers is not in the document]
NeuralSpaz: delete

experimental/devices/as7262/as7262.go
```diff
@@ -221,17 +224,11 @@ func (d *Dev) Sense(ledDrive physic.ElectricCurrent, senseTime time.Duration) (S
...
-	d.canceledMu.Lock()
-	defer d.canceledMu.Unlock()
-
-	if !d.canceled {
-		d.done <- struct{}{}
-		d.canceled = true
-	}
+	d.cancelMu.Lock()
+	defer d.cancelMu.Unlock()
+	d.cancel()
```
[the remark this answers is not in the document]
NeuralSpaz: The wait group proved to be a little cumbersome to manage. 
I'm trying something new, and that is to have a done channel, but never send on it, just close it after the ctx.Done() is received.. This only works if close(chan) op is goroutine safe?

experimental/devices/as7262/as7262.go
```diff
@@ -37,25 +38,27 @@ func New(bus i2c.Bus, opts *Opts) (*Dev, error) {
 		c:         &i2c.Dev{Bus: bus, Addr: 0x49},
 		gain:      opts.Gain,
 		interrupt: opts.InterruptPin,
-		done:      make(chan struct{}, 1),
-		timeout:   200 * time.Millisecond,
+		cancel:    func() {},
```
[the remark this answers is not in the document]
NeuralSpaz: much safe if it is called at the wrong time. Could also be used for a message if that happens to track problem.
```go
		cancel: func() {
			fmt.Println("cancel without context")
		},
```

experimental/devices/as7262/as7262.go
```diff
@@ -125,64 +128,64 @@ func (d *Dev) Sense(ledDrive physic.ElectricCurrent, senseTime time.Duration) (S
...
-
 	it, integration := calcSenseTime(senseTime)
-	if err := d.writeVirtualRegister(intergrationReg, it); err != nil {
+	ctx, cancel := context.WithCancel(context.Background())
+	d.cancelMu.Lock()
+	d.cancel = cancel
+	d.cancelMu.Unlock()
+	defer d.cancel()
+
+	if err := d.writeVirtualRegister(ctx, intergrationReg, it); err != nil {
```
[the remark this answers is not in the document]
NeuralSpaz: yes. :)

experimental/devices/as7262/as7262.go
```diff
@@ -125,64 +130,67 @@ func (d *Dev) Sense(ledDrive physic.ElectricCurrent, senseTime time.Duration) (S
...
-
 	it, integration := calcSenseTime(senseTime)
-	if err := d.writeVirtualRegister(intergrationReg, it); err != nil {
+	ctx, cancel := context.WithCancel(context.Background())
+	d.cancelMu.Lock()
+	d.cancel = cancel
+	d.done = make(chan struct{})
+	d.cancelMu.Unlock()
+	defer d.cancel()
+
```
maruel: I don't understand why you wouldn't `defer func() { close(d.done) }()`, then Halt() wouldn't need a timeout.

experimental/devices/as7262/as7262.go
```diff
@@ -37,25 +38,29 @@ func New(bus i2c.Bus, opts *Opts) (*Dev, error) {
 		c:         &i2c.Dev{Bus: bus, Addr: 0x49},
 		gain:      opts.Gain,
 		interrupt: opts.InterruptPin,
-		done:      make(chan struct{}, 1),
-		timeout:   200 * time.Millisecond,
+		cancel:    func() {},
+		done:      make(chan struct{}),
```
maruel: you should create the channel, and close it immediately.

```
d := make(chan struct{})
close(d)
...

done: d,
```

experimental/devices/as7262/as7262.go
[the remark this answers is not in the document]
maruel: It shouldn't matter, since it's Sense() (the sender) that closes it.

experimental/devices/as7262/as7262.go
NeuralSpaz: Thats solves the timeout problem too. : )

experimental/devices/as7262/as7262.go
NeuralSpaz: Good catch. I want to wait on the close, so Timeout was there for the case if Halt() was called before Sense() when the channel was still not closed after New(). Now that done is closed in New() that is not a problem.

experimental/devices/as7262/as7262.go
```diff
@@ -221,16 +233,18 @@ func (d *Dev) Sense(ledDrive physic.ElectricCurrent, senseTime time.Duration) (S
...
-	d.canceledMu.Lock()
-	defer d.canceledMu.Unlock()
-
-	if !d.canceled {
-		d.done <- struct{}{}
-		d.canceled = true
+	d.cancelMu.Lock()
+	defer d.cancelMu.Unlock()
+	d.cancel()
+	select {
```
maruel: There's no need for select now, just do:
```
_, _ = <- d.done
```

and that's it.

experimental/devices/as7262/as7262.go
```diff
@@ -284,13 +298,9 @@ func (d *Dev) Gain(gain Gain) error {
...
 	defer d.mu.Unlock()
 
-	done := make(chan struct{}, 1)
-	d.canceledMu.Lock()
-	d.done = done
-	d.canceled = false
-	d.canceledMu.Unlock()
-
-	if err := d.writeVirtualRegister(controlReg, uint8(gain)); err != nil {
+	ctx, cancel := context.WithTimeout(context.Background(), sensorTimeout)
```
maruel: In this case, I'd say to just use context.Background(), because you don't intent to cancel the context while the function is running.

experimental/devices/as7262/as7262.go
NeuralSpaz: That's much cleaner.

experimental/devices/as7262/as7262.go
NeuralSpaz: Very true, fixed.

experimental/devices/as7262/as7262.go
```diff
@@ -221,17 +233,14 @@ func (d *Dev) Sense(ledDrive physic.ElectricCurrent, senseTime time.Duration) (S
...
 func (d *Dev) Halt() error {
-	d.canceledMu.Lock()
-	defer d.canceledMu.Unlock()
-
-	if !d.canceled {
-		d.done <- struct{}{}
-		d.canceled = true
-	}
+	d.cancelMu.Lock()
+	defer d.cancelMu.Unlock()
```
maruel: There's a small race condition, you need to do `done := d.done` while holding d.cancelMu, so line 243 become `_, _ = <-done`, otherwise there's a tiny window where a concurrent Sense() call may replace d.done with a new channel. Not a big deal but I prefer this for complete correctness.

NeuralSpaz: I have the lock at both ends don't I? 
L136 and L238. Only other spot is in New()

maruel: Ah I misread, I guess my point is that you don't need to hold cancelMu, but in practice it doesn't matter.

PR #400, opened 2019-02-10
sysfs: fix a data race condition in events

maruel:
It's not used yet. In practice the previous code would have worked
mostly fine, but it's triggering data race check with running tests with
-race, so it's necessary to have this fixed.

### DIFF
--- a/host/sysfs/fs_linux.go
+++ b/host/sysfs/fs_linux.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


host/sysfs/fs_linux.go
```diff
@@ -82,26 +83,33 @@ func (e epollEvent) String() string {
...
 func (e *eventsListener) init() error {
-	// In theory this check should be done with a lock but in practice this is
-	// not necessary, as this value is only set with a lock.
-	if e.fds != nil {
+	if atomic.LoadInt32(&e.initialized) != 0 {
 		// Was already initialized.
 		return nil
 	}
 	e.mu.Lock()
+	if atomic.LoadInt32(&e.initialized) != 0 {
```
NeuralSpaz: This should also release the lock. If the there are concurrent calls of init  all piled up waiting on the lock, if they get to take the lock it will stop loop().

maruel: Oh gosh thanks!

maruel:
Dang the beaglebone died again.

PR #403, opened 2019-04-01
mcp9808: Fix conversions between temperature and bits

mtraver:
The datasheet (http://ww1.microchip.com/downloads/en/DeviceDoc/25095A.pdf)
says that negative temperatures are stored in two's complement. See
page 22 for alert registers and pages 24-25 for the ambient temperature
register.

However, the code currently assumes negative values are stored in a
sign-and-magnitude representation. Fix this to handle two's complement
values correctly.

For register bits to temperature conversion this implementation follows
the datasheet, but the code in the datasheet isn't quite right -- it
should be `temp - 256` instead of `256 - temp` -- but the fact that the
256 figure is in Microchip's code backs up the idea stated in the text
that negative values are in two's complement. Others seem to agree that
this is correct: https://electronics.stackexchange.com/a/244826

For temperature to bits conversion this implementation relies on the
fact that Go also uses two's complement. I've added a long comment in
the code about this.

### DIFF
--- a/experimental/devices/mcp9808/mcp9808.go
+++ b/experimental/devices/mcp9808/mcp9808.go
@@ -264,10 +264,10 @@ func (d *Dev) readTemperature() (physic.Temperature, uint8, error) {
 		return 0, 0, errReadTemperature
 	}
 	// Convert to physic.Temperature 0.0625°C per bit
-	t := physic.Temperature(tbits&0x0FFF) * 62500 * physic.MicroKelvin
+	t := physic.Temperature(tbits&0x0fff) * 62500 * physic.MicroKelvin
 	if tbits&0x1000 > 0 {
-		// Check for bit sign.
-		t = -t
+		// Check for sign bit.
+		t -= 256 * physic.Celsius
 	}
 	t += physic.ZeroCelsius
 	return t, uint8(tbits>>8) & 0xe0, nil
@@ -383,10 +383,10 @@ var (
 )
 
 func alertBitsToTemperature(b uint16) physic.Temperature {
-	b = (b >> 2) & 0x07FF
-	t := physic.Temperature(b&0x03FF) * 250 * physic.MilliKelvin
+	b = (b >> 2) & 0x07ff
+	t := physic.Temperature(b&0x03ff) * 250 * physic.MilliKelvin
 	if b&0x400 > 0 {
-		t = -t
+		t -= 256 * physic.Celsius
 	}
 	t += physic.ZeroCelsius
 	return t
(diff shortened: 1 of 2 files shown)


mtraver:
Tagging @NeuralSpaz, who added MCP9808 support.

I haven't actually put an MCP9808 in the freezer to test this, but the info in the datasheet points toward this implementation for negative temperatures.

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -264,10 +264,10 @@ func (d *Dev) readTemperature() (physic.Temperature, uint8, error) {
 		return 0, 0, errReadTemperature
 	}
 	// Convert to physic.Temperature 0.0625°C per bit
-	t := physic.Temperature(tbits&0x0FFF) * 62500 * physic.MicroKelvin
+	t := physic.Temperature(tbits&0x0fff) * 62500 * physic.MicroKelvin
```
maruel: I don't think I've been consistent on the casing of hex values, but I'd prefer the ones that do not have the value changed to not be edited, so the blame list is not affected.

experimental/devices/mcp9808/mcp9808.go
```diff
@@ -383,10 +383,10 @@ var (
 )
 
 func alertBitsToTemperature(b uint16) physic.Temperature {
-	b = (b >> 2) & 0x07FF
-	t := physic.Temperature(b&0x03FF) * 250 * physic.MilliKelvin
+	b = (b >> 2) & 0x07ff
```
maruel: I think (?) the shift is not really needed, because one could do:
```
t := physic.Temperature(b&0x0fff) * 62500 * physic.MicroKelvin
if b&0x1000 != 0 {
  t -= 256 * physic.Celsius
}
return t + physic.ZeroCelsius
```

I find it a bit easier to read too but I'm not 100% sure the above is correct.

maruel:
gohci

maruel:
> You can add yourself to AUTHORS and CONTRIBUTORS if you want.

Oh I confirmed that'd be only CONTRIBUTORS for you since you are already covered in AUTHORS.

experimental/devices/mcp9808/mcp9808.go
mtraver: Fair enough. Reverted.

experimental/devices/mcp9808/mcp9808.go
maruel: Any strong opinion about this?

NeuralSpaz:
Never gets that cold here so I've popped one in the freezer.

NeuralSpaz:
freezer test lgtm:
```
-4.062°C
-4.125°C
-4.062°C
-4.062°C
-4.125°C
-4.187°C
-3.875°C
-2.500°C
-1.562°C
-1.437°C
-750m°C
-500m°C
-187.500m°C
187.500m°C
625m°C
1.125°C
1.375°C
1.687°C
1.937°C
2.312°C
2.500°C
2.875°C
3.125°C
3.375°C
3.625°C
3.937°C
4.062°C
```

mtraver:
> freezer test lgtm:

Fantastic! Thanks for testing.

PR #406, opened 2019-04-08
onewire: experimental netlink-based bus master for Linux

### LINKED ISSUE #359: sysfs: support onewire (w1)
It looks like sysfs supports Onewire, https://www.kernel.org/doc/Documentation/w1/w1.generic, and https://www.kernel.org/doc/Documentation/w1/w1.netlink

mhabb:
This driver adds a gpio-based 1-wire master that uses the Linux netlink
connector API to communicate with the w1 kernel module.

Major caveat: the w1 netlink API does not support setting strong
pull-ups after write operations. Devices such as the DS18B20 must be
powered externally in order to work with this driver.

Fixes #359

### DIFF
--- a/experimental/host/netlink/doc.go
+++ b/experimental/host/netlink/doc.go
@@ -1,3 +1,7 @@
+// Copyright 2019 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
 // Package netlink implements host drivers based on the Linux netlink connector
 // interface. See
 // https://www.kernel.org/doc/Documentation/connector/connector.txt for details.
--- a/experimental/host/netlink/doc.go
+++ b/experimental/host/netlink/doc.go
@@ -1,0 +1,4 @@
+// Package netlink implements host drivers based on the Linux netlink connector
+// interface. See
+// https://www.kernel.org/doc/Documentation/connector/connector.txt for details.
+package netlink
--- a/experimental/host/netlink/netlink_linux.go
+++ b/experimental/host/netlink/netlink_linux.go
@@ -1,4 +1,4 @@
-// Copyright 2016 The Periph Authors. All rights reserved.
+// Copyright 2019 The Periph Authors. All rights reserved.
 // Use of this source code is governed under the Apache License, Version 2.0
 // that can be found in the LICENSE file.
 
--- a/experimental/host/netlink/netlink_linux.go
+++ b/experimental/host/netlink/netlink_linux.go
@@ -4,4 +4,51 @@
 
 package netlink
 
+import (
+	"fmt"
+	"syscall"
+)
+
 const isLinux = true
+
+// connSocket is a simple wrapper around a Linux netlink connector socket.
+type connSocket struct {
+	fd int
+}
+
+// newConnSocket returns a socket instance.
+func newConnSocket() (*connSocket, error) {
+	// Open netlink socket.
+	fd, err := syscall.Socket(syscall.AF_NETLINK, syscall.SOCK_DGRAM, syscall.NETLINK_CONNECTOR)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open netlink socket: %v", err)
+	}
+
+	if err := syscall.Bind(fd, &syscall.SockaddrNetlink{Family: syscall.AF_NETLINK}); err != nil {
+		return nil, fmt.Errorf("failed to bind netlink socket: %v", err)
+	}
+
+	return &connSocket{fd: fd}, nil
+}
+
+// send writes w to the socket.
+func (s *connSocket) send(w []byte) error {
+	return syscall.Sendto(s.fd, w, 0, &syscall.SockaddrNetlink{Family: syscall.AF_NETLINK})
+}
+
+// recv reads at most len(r) bytes from the socket into r. Returns the actually
+// read number of bytes.
+func (s *connSocket) recv(r []byte) (int, error) {
+	n, _, err := syscall.Recvfrom(s.fd, r, 0)
+	if err != nil {
+		return 0, err
+	}
+	return n, nil
+}
+
+// close closes the socket.
+func (s *connSocket) close() error {
+	fd := s.fd
+	s.fd = 0
+	return syscall.Close(fd)
+}
--- a/experimental/host/netlink/netlink_linux.go
+++ b/experimental/host/netlink/netlink_linux.go
@@ -1,0 +1,7 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package netlink
+
+const isLinux = true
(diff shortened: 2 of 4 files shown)


mhabb:
I signed it!

experimental/host/netlink/onewire.go
```diff
@@ -0,0 +1,552 @@
...
+		masterID: masterID,
+		s:        s,
+	}, nil
+}
+
+//
+
+const (
+	// Supported netlink message type.
+	NLMSG_DONE = uint16(0x3)
```
maruel: Did you mean to export these?

experimental/host/netlink/onewire.go
```diff
@@ -0,0 +1,552 @@
...
+		if err := onewirereg.Register(name, aliases, bus, openerOneWire(bus).Open); err != nil {
+			return true, err
+		}
+	}
+	return true, nil
+}
+
+//
+
+func putUint16(b []byte, v uint16) {
```
maruel: What about using binary.LittleEndian.PutUint16 ?
https://golang.org/pkg/encoding/binary/

maruel:
gohci

experimental/host/netlink/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package netlink implements host drivers based on the Linux netlink connector
```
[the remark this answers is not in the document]
mhabb: Done

experimental/host/netlink/onewire.go
mhabb: No, changed all names of constants.

experimental/host/netlink/onewire.go
mhabb: I was basically doing the same as in https://github.com/google/periph/blob/f4587c12845918ffaba8b966fd794753a4b84f5a/conn/onewire/onewire.go#L215. But binary.LittleEndian is of course fine.

experimental/host/netlink/onewire.go
```diff
@@ -0,0 +1,552 @@
...
+//
+// masterID is the bus number reported by the netlink W1_LIST_MASTERS command.
+// The resulting object is safe for concurrent use.
+//
+// NOTE: the Linux 1-wire netlink API does not support strong pull-ups after
+// write operations. Hence this driver does not support this feature either. The
+// pull-up argument passed to Tx() is ignored. Devices may need to be powered
+// externally to work with this driver.
+//
+// Do not use netlink.NewOneWire() directly as package netink is providing a
```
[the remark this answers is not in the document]
mhabb: Done

experimental/host/netlink/netlink_linux.go
```diff
@@ -0,0 +1,7 @@
+// Copyright 2016 The Periph Authors. All rights reserved.
```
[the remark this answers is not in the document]
mhabb: Done

experimental/host/netlink/onewire.go
```diff
@@ -0,0 +1,552 @@
...
+	"fmt"
+	"sync"
+	"syscall"
+
+	"periph.io/x/periph"
+	"periph.io/x/periph/conn/onewire"
+	"periph.io/x/periph/conn/onewire/onewirereg"
+)
+
+// NewOneWire opens a 1-wire bus via its netlink interface as described at
```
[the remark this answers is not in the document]
mhabb: Switched to New().

experimental/host/netlink/onewire.go
maruel: Ah, this particular case was because I worked on making sure that packages under conn/ do not depend on reflect, and "binary" does depend on reflect. But here you are already depending on fmt anyway.

experimental/host/netlink/onewire.go
```diff
@@ -0,0 +1,511 @@
...
+	o.mu.Lock()
+	defer o.mu.Unlock()
+
+	o.seq++
+
+	d, err := o.s.sendAndRecv(o.seq, m)
+	if err != nil {
+		return fmt.Errorf("netlink-onewire: transaction failed: %v", err)
+	}
+	copy(r, d)
```
maruel: I don't understand why this line.

experimental/host/netlink/onewire.go
```diff
@@ -0,0 +1,511 @@
...
+
+	return nil
+}
+
+// Search performs a device search operation on the 1-wire bus. The resulting
+// device addresses are returned. If alarmOnly is true, only devices in alarm
+// state are returned.
+func (o *OneWire) Search(alarmOnly bool) ([]onewire.Address, error) {
+	cmd := w1CmdSearch()
```
maruel: It's kind of sad to allocate on the heap to immediately then discard the object when alarmOnly is true.

Optional nit: I personally wish cmds was a `[]w1Cmd` to reduce heap pressure but that's a micro optimization, that doesn't need to be done here.

maruel:
gohci

maruel:
Looks like you'll have to move some of the syscall into netlink_linux.go because they are not defined on macOS and Windows. e.g. https://gist.github.com/gohci-bot/90884ea4bcd6503926f32fa86cad8653

experimental/host/netlink/onewire.go
mhabb: Copying to r is correct: d is the data read from the netlink socket (after removing all headers), r is the read buffer passed to Tx() that needs to be populated.

experimental/host/netlink/onewire.go
mhabb: Changed the way cmd is constructed. Leaving the type of cmds as is to avoid copying struct w1Cmd if that's ok with you.

mhabb:
gohci

experimental/host/netlink/onewire.go
```diff
@@ -247,35 +249,38 @@ func w1CmdWrite(d []byte) *w1Cmd { return &w1Cmd{typ: cmdWrite, payloadLen: len(
 
 //
 
-// socket is a netlink connector socket for reading and writing 1-wire messages.
-type socket struct {
-	fd int
+type socket interface {
```
mhabb: Note that this new interface is not needed right now, but will enable testing in a follow-up PR.

experimental/host/netlink/onewire.go
maruel: Ugh, brain fart.

maruel:
gohci

maruel:
Only project owners can trigger gohci because of lack of process containment, as by definition the worker needs access to physical pins.

maruel:
Sadly macOS still failed with `syscall.SizeofNlMsghdr`, fix this and then lgtm.

experimental/host/netlink/onewire.go
```diff
@@ -152,6 +151,11 @@ func newOneWire(masterID uint32) (*OneWire, error) {
 //
 
 const (
+	// Size of struct nlmsghdr.
+	sizeofNlMsghdr = 16
```
maruel: I confirmed the value is the same everywhere, but I'd prefer to have a comment here, because it's not obvious, so something like:

    // Size of struct nlmsghdr, the value of syscall.SizeofNlMsghdr.

maruel:
gohci

experimental/host/netlink/onewire.go
maruel: This can be done in a follow up.

maruel:
Thanks a lot!

PR #414, opened 2019-07-06
pca9685: cleanup

ether42:
The PCA9685's datasheet is available [here](https://cdn-shop.adafruit.com/datasheets/PCA9685.pdf).

First commit cleanups some stuff you commented on in https://github.com/google/periph/pull/296.

Second commit makes use of the auto-increment capability of the device to simplify contiguous writes. You've talked about `mmr`, do you think it's still required?

I tested it with a couple of servos. Do you have plans for the [servo module](https://github.com/google/periph/blob/master/experimental/devices/pca9685/servo.go)? I feel like one should be able to associate some `Servo`s to different `ServoGroup`s and the channel should be abstracted away (for example it could be possible to set the angle of all the `Servo`s in the `ServoGroup`).

### DIFF
--- a/experimental/devices/pca9685/pca9685.go
+++ b/experimental/devices/pca9685/pca9685.go
@@ -15,32 +15,31 @@ import (
 // I2CAddr i2c default address.
 const I2CAddr uint16 = 0x40
 
-// PCA9685 Commands
+// PCA9685 registers.
 const (
-	mode1      byte = 0x00
-	mode2      byte = 0x01
-	subAdr1    byte = 0x02
-	subAdr2    byte = 0x03
-	subAdr3    byte = 0x04
-	prescale   byte = 0xFE
-	led0OnL    byte = 0x06
-	led0OnH    byte = 0x07
-	led0OffL   byte = 0x08
-	led0OffH   byte = 0x09
-	allLedOnL  byte = 0xFA
-	allLedOnH  byte = 0xFB
-	allLedOffL byte = 0xFC
-	allLedOffH byte = 0xFD
-
-	// Bits
-	restart byte = 0x80
-	sleep   byte = 0x10
-	allCall byte = 0x01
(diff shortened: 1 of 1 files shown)


experimental/devices/pca9685/pca9685.go
```diff
@@ -89,7 +88,7 @@ func (d *Dev) init() error {
 	return d.SetPwmFreq(50 * physic.Hertz)
 }
 
-// SetPwmFreq set the pwm frequency
+// SetPwmFreq set the pwm frequency.
```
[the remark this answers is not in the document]
ether42: I know of this one but I was a bit reluctant to change the name of the functions. Experimental doesn't have any compatibility policy, so should we anyway?

ether42:
Squashed commits and added an helper function to avoid redundancy.

experimental/devices/pca9685/pca9685.go
```diff
@@ -111,40 +110,29 @@ func (d *Dev) SetPwmFreq(freqHz physic.Frequency) error {
...
 	time.Sleep(100 * time.Millisecond)
 
-	_, err := d.dev.Write([]byte{mode1, (byte)(oldmode | 0x80)})
+	_, err := d.dev.Write([]byte{mode1, oldmode | restart})
 	return err
 }
 
-// SetAllPwm set a pwm value for all outputs
+// setPWM writes a PWM value in a specific register.
+func (d *Dev) setPwm(register uint8, on, off gpio.Duty) error {
```
maruel: Update the function name too, not just the comment.

Let's punt on the public ones for now to not break clients yet, but the private one can be fixed right away.

ether42: That was an oversight on my part, sorry.

maruel:
gohci

maruel:
Thanks!

PR #417, opened 2019-08-06
environment: migrate interface from physic.SenseEnv

maruel:
- environment.SenseWeather replaces physic.SenseEnv in v4.
- Use Weather instead of Env. This is more precise.
- Use SenseWeather() instead of Sense(). The rationale for the rename is
  that we will add new SenseXX() functions, and a single device could
  sense multiple things (environment, orientation, etc) on the same Go
  object. Having all interfaces uses the generic verb 'Sense()' would
  cause conflict.
- Add examples but do not implement it yet.
- Make SenseWeatherContinous blocking (synchronous) and require a
  context. This is part of the larger context based operation in v4.

Package physic wasn't the right place to host this interface. Other
interfaces (like orientation) can be added to package environment.

No behavior change, only new interfaces.

### DIFF
--- a/conn/environment/doc.go
+++ b/conn/environment/doc.go
@@ -1,0 +1,6 @@
+// Copyright 2019 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package environment declares interfaces for environmental sensors.
+package environment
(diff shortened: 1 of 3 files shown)


maruel:
I'm not a fan of environment.SenseWeather.SenseWeather(). Better naming ideas are appreciated.

I did the implementation of all the current sensors, including SenseWeatherContinuous(), but I want to nail down the API first.

maruel:
This needs to be redone, I'll open a separate CL.

PR #430, opened 2019-12-23
inky: Add wHAT support

davidsansome:
The wHAT is a larger version of the pHAT: https://shop.pimoroni.com/products/inky-what


maruel:
Hi David, I'm travelling atm I'll review as soon as I can.

davidsansome:
No problem: there's no hurry at all!  Enjoy your holiday!

experimental/cmd/inky/inky.go
```diff
@@ -0,0 +1,103 @@
...
+	path        = flag.String("image", "", "Path to a png file to display on the inky")
+	dcPin       = flag.String("dc", "22", "Inky DC Pin")
+	resetPin    = flag.String("reset", "27", "Inky Reset Pin")
+	busyPin     = flag.String("busy", "17", "Inky Busy Pin")
+	model       = flag.String("model", "PHAT", "Inky model (PHAT or WHAT)")
+	modelColor  = flag.String("model-color", "red", "Inky model color (black, red or yellow)")
+	borderColor = flag.String("border-color", "black", "Border color (black, white, red or yellow)")
+)
+
+func parseModel(s string) inky.Model {
```
maruel: What do you think of this? While this is way more code, the runtime behavior is more interesting, as it prints a proper message instead of crashing.

```
type modelType inky.Model

func (m *modelType) String() string {
	switch *m {
	case inky.PHAT:
		return "PHAT"
	case inky.WHAT:
		return "WHAT"
	default:
		return "Unknown"
	}
}

func (m *modelType) Set(s string) error {
	switch s {
	case "PHAT":
		*m = (modelType)(inky.PHAT)
	case "WHAT":
		*m = (modelType)(inky.WHAT)
	default:
		return fmt.Errorf("Unknown model %q: expected either PHAT or WHAT", s)
	}
	return nil
}

var model modelType

func main() {
	flag.Var(&model, "model", "PHAT", "Inky model (PHAT or WHAT)")
	(...)
}
```

It's also something we'd encourage to copy paste.

As a matter of fact I'd prefer putting these two methods directly on inky.Model instead, similar to https://godoc.org/periph.io/x/periph/conn/i2c#Addr. Then only the main() code above is needed here.

Same for color.

experimental/devices/inky/inky.go
```diff
@@ -38,9 +32,9 @@ const (
 
 var borderColor = map[Color]byte{
 	Black:  0x00,
-	Red:    0x33,
+	Red:    0x73,
```
maruel: That's weird, what happened?

experimental/devices/inky/inky.go
```diff
@@ -83,6 +81,15 @@ func New(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinIn, o *Opts
 		border: o.BorderColor,
 	}
 
+	switch o.Model {
+	case PHAT:
+		d.bounds = image.Rect(0, 0, 104, 212)
+		d.flipVertically = true
+	case WHAT:
+		d.bounds = image.Rect(0, 0, 400, 300)
+		d.flipVertically = false
```
maruel: I'm fine with removing the line.

experimental/devices/inky/inky.go
```diff
@@ -100,13 +107,26 @@ type Dev struct {
 	color Color
 	// Modifiable color of border.
 	border Color
+
+	bounds         image.Rectangle
```
maruel: Since these are not mutable, they should go in the block above of immutable members.

experimental/devices/inky/inky.go
```diff
@@ -188,84 +211,58 @@ func (d *Dev) Draw(dstRect image.Rectangle, src image.Image, srcPtrs image.Point
...
 		{0x3c, []byte{0x00}},
 		{0x3c, []byte{byte(border)}}, // Border colour
 	}
-
-	for _, c := range init {
-		if err := d.sendCommand(c.cmd, c.data); err != nil {
-			return err
-		}
+	if d.color == Yellow {
+		cmds = append(cmds, cmdData{0x04, []byte{0x07, 0xac, 0x32}}) // Set voltage of VSH and VSL
```
maruel: That's interesting. Do you know why?

experimental/devices/inky/inky.go
```diff
@@ -188,84 +211,58 @@ func (d *Dev) Draw(dstRect image.Rectangle, src image.Image, srcPtrs image.Point
...
 	}
+	cmds = append(cmds, cmdData{0x32, modelLUT[d.color]}) // Set LUTs
```
maruel: Does this have to be done after the VSH/VSL or can it be included in the block lines 233-244? Same for the rest.

experimental/devices/inky/inky.go
```diff
@@ -292,6 +289,9 @@ func (d *Dev) update(border byte, black []byte, red []byte) (err error) {
 }
 
 func (d *Dev) reset() (err error) {
+	if err = d.busy.In(gpio.PullUp, gpio.FallingEdge); err != nil {
```
maruel: That was an error in the code?

experimental/devices/inky/inky.go
```diff
@@ -332,14 +329,19 @@ func (d *Dev) sendCommand(command byte, data []byte) error {
 }
 
 func (d *Dev) sendData(data []byte) error {
-	if len(data) > 4096 {
-		return fmt.Errorf("sending more data than chunk size: %d > 4096", len(data))
-	}
 	if err := d.dc.Out(gpio.High); err != nil {
```
maruel: I wonder what's the required timing in ms between setting the line and sending the actual commands and data. i.e. if a very small sleep should be added.

experimental/cmd/inky/inky.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+func main() {
+	flag.Parse()
+
+	// Open and decode the image.
+	f, err := os.Open(*path)
+	if err != nil {
+		log.Fatal(err)
```
maruel: I'm really a fan of having a `mainImpl() error` and use this instead of log.Fatal() since it's much more user friendly than a crash.

maruel:
gohci

experimental/devices/inky/inky.go
davidsansome: I found the datasheet: https://www.china-epaper.com/uploads/soft/DEPG0420R01V3.0.pdf - it has timings for the 3-wire SPI configuration but not the 4-wire configuration with D/C pin that we're using here :/

experimental/devices/inky/inky.go
davidsansome: Yep it seems to still work with all these moved above the VSH/VSL command.  I think it's just the 0x22 (image update) that needs to go at the end .

experimental/devices/inky/inky.go
davidsansome: This was a bug that was fixed upstream, in https://github.com/pimoroni/inky/commit/8165ca0f43c22f54fd6d018a816df4ab37fa92f0 for https://github.com/pimoroni/inky/issues/6

experimental/devices/inky/inky.go
davidsansome: Done!

experimental/devices/inky/inky.go
davidsansome: This one was changed upstream in https://github.com/pimoroni/inky/commit/4769114be193d44215ae64236e74f25f7b485c7a for https://github.com/pimoroni/inky/issues/12

experimental/cmd/inky/inky.go
davidsansome: Great ideas, done!

experimental/devices/inky/lut.go
```diff
@@ -5,27 +5,31 @@
...
+	yellowLUT = []byte{
 		0xfa, 0x94, 0x8c, 0xc0, 0xd0, 0x0, 0x0, 0xfa, 0x94, 0x2c, 0x80, 0xe0, 0x0, 0x0, 0xfa, 0x0,
 		0x0, 0x0, 0x0, 0x0, 0x0, 0xfa, 0x94, 0xf8, 0x80, 0x50, 0x0, 0xcc, 0xbf, 0x58, 0xfc, 0x80,
 		0xd0, 0x0, 0x11, 0x40, 0x10, 0x40, 0x10, 0x8, 0x8, 0x10, 0x4, 0x4, 0x10, 0x8, 0x8, 0x3,
 		0x8, 0x20, 0x8, 0x4, 0x0, 0x0, 0x10, 0x10, 0x8, 0x8, 0x0, 0x20, 0x0, 0x0, 0x0, 0x0,
 		0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
 	}
+
+	modelLUT = map[Color][]byte{
+		Black:  blackLUT,
```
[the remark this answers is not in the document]
davidsansome: Done.

experimental/devices/inky/lut.go
```diff
@@ -5,27 +5,31 @@
 package inky
 
 var (
-	blackLUT = [...]byte{
+	blackLUT = []byte{
```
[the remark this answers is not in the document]
davidsansome: I'm an idiot - I've turned these back to arrays and used [:] below.

experimental/cmd/inky/inky.go
davidsansome: Done!

experimental/devices/inky/inky.go
davidsansome: Done!

experimental/devices/inky/inky.go
davidsansome: Err I actually can't remember why I moved this :smile: 
I've moved it back - it seems to work either way.

experimental/cmd/inky/inky.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	log.Printf("Opening %s...", *spiPort)
+	b, err := spireg.Open(*spiPort)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("Opening pins...")
+	dc := gpioreg.ByName(*dcPin)
```
maruel: ByName() can return nil if the user specified an invalid pin, this would crash the app, it'd be better to print a error to the user.

experimental/devices/inky/inky.go
```diff
@@ -36,11 +30,36 @@ const (
...
+	case Black:
+		return "black"
+	case Red:
+		return "red"
+	case Yellow:
+		return "yellow"
+	case White:
+		return "white"
+	default:
+		return "Unknown"
```
maruel: I guess "unknown" to be consistent with the rest?

experimental/devices/inky/inky.go
```diff
@@ -36,11 +30,36 @@ const (
...
+	}
+}
+
+// Set sets the Color to a value represented by the string s. Set implements the flag.Value interface.
+func (c *Color) Set(s string) error {
+	switch s {
+	case "black":
+		*c = (Color)(Black)
```
maruel: I'm pretty sure `*c = Black` should work?
Same for lines 89, 91.

experimental/devices/inky/inky.go
```diff
@@ -63,7 +106,16 @@ type Opts struct {
 	BorderColor Color
 }
 
-// New opens a handle to an Inky pHAT.
+const spiChunkSize = 4096
```
maruel: I think this code was specifically designed to work around the default fairly small SPI driver buffer size on Raspbian.

It can now be discovered dynamically with MaxTxSize().
https://github.com/google/periph/blob/master/host/sysfs/spi.go#L141

It's useful to use this value because it uses the current value as exposed by the driver
https://github.com/google/periph/blob/master/host/sysfs/spi.go#L592

I see the documentation is currently lacking, it's only noted in the Conn comment
https://godoc.org/periph.io/x/periph/conn/spi#Conn

There's a formal interface https://godoc.org/periph.io/x/periph/conn#Limits but I think I'll put it in all busses for v4.

maruel:
gohci

davidsansome:
> A few nits but nothing blocking. I'll merge right after the gohci run.

Thanks!  I've fixed these in a followup: https://github.com/google/periph/pull/431

PR #431, opened 2020-01-03
inky: Minor cleanups

davidsansome:
* Use maxTxSize from conn.Limits
* Show an error instead of crashing on invalid pin names


maruel:
gohci

maruel:
Thanks!

PR #435, opened 2020-04-08
bcm283x: present: check device tree on non-raspbian systems

stapelberg:
This is required to use periph.io on https://gokrazy.org.

I don’t know whether the device-tree method also works on Raspbian. If so, would you like to use just the device-tree method everywhere, or do you want to go the more conservative fall-back approach I’m suggesting here?

Thanks,

### DIFF
--- a/host/bcm283x/gpio.go
+++ b/host/bcm283x/gpio.go
@@ -80,8 +80,11 @@ var (
 // Present returns true if running on a Broadcom bcm283x based CPU.
 func Present() bool {
 	if isArm {
-		hardware, ok := distro.CPUInfo()["Hardware"]
-		return ok && strings.HasPrefix(hardware, "BCM")
+		for _, line := range distro.DTCompatible() {
+			if strings.HasPrefix(line, "brcm,bcm") {
+				return true
+			}
+		}
 	}
 	return false
 }


host/bcm283x/gpio.go
```diff
@@ -80,8 +82,24 @@ var (
...
+		// This checks a magic string which the
+		// https://github.com/raspberrypi/linux kernel patched into
+		// /proc/cpuinfo.
 		hardware, ok := distro.CPUInfo()["Hardware"]
-		return ok && strings.HasPrefix(hardware, "BCM")
+		if ok && strings.HasPrefix(hardware, "BCM") {
+			return true
+		}
+		// For non-Raspbian systems (e.g. gokrazy, Fedora, Debian, …):
+		b, err := ioutil.ReadFile("/proc/device-tree/compatible")
```
maruel, line 93: Please use distro.DTCompatible() so it's only read once per process.

maruel:
I don't have a strong opinion, if checking /proc/device-tree/compatible works all the time and doesn't lead to too much false positives, I'm fine with only checking this. Not that the previous code was optimized for false positives either.

host/bcm283x/gpio.go
stapelberg: Done, thanks for the pointer!

stapelberg:
> I don't have a strong opinion, if checking /proc/device-tree/compatible works all the time and doesn't lead to too much false positives, I'm fine with only checking this. Not that the previous code was optimized for false positives either.

I just verified that this method works on the Raspberry Pi 4 and 3 when running Raspbian or gokrazy. Updated the PR to unconditionally check /proc/device-tree/compatible.

maruel:
gohci

PR #436, opened 2020-04-09
rpi: obtain revision from device tree (for non-raspbian systems)

stapelberg:
This is required to use periph.io on https://gokrazy.org.

This change was tested on:

Raspbian (2020-02-13-raspbian-buster-lite.img)
Raspberry Pi 4 Model B Rev 1.1
Raspberry Pi 3 Model B Rev 1.2

gokrazy (Linux 5.6.2)
Raspberry Pi 4 Model B Rev 1.1
Raspberry Pi 3 Model B Rev 1.2

### DIFF
--- a/host/distro/devtree.go
+++ b/host/distro/devtree.go
@@ -4,6 +4,11 @@
 
 package distro
 
+import (
+	"encoding/binary"
+	"io/ioutil"
+)
+
 // DTModel returns platform model info from the Linux device tree (/proc/device-tree/model), and
 // returns "unknown" on non-linux systems or if the file is missing.
 func DTModel() string {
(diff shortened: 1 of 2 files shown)


host/distro/devtree.go
```diff
@@ -35,11 +41,33 @@ func DTCompatible() []string {
...
+		return dtRevision
+	}
+	b, err := ioutil.ReadFile("/proc/device-tree/system/linux,revision")
+	if err != nil {
+		return ""
+	}
+	if len(b) < 4 {
+		return ""
+	}
+	dtRevision = fmt.Sprintf("%04x", binary.BigEndian.Uint32(b[:4]))
```
maruel, line 61: Do you think it's always going to be a uint32? If so maybe we could keep it as a uint32.

One unstated part of the API here is that the format will always be prefixed with at least 4 zeros, and it's now implicitly baked into the expectations, e.g.

```
switch distro.DTRevision() {
  case "0x0001":
  cast "0x123456":
}
```

so this will be a breaking change to change the format string later. That's not an issue if you make it a uint32 or uint64.

maruel:
One question about the API.

host/distro/devtree.go
stapelberg: Done.

host/distro/devtree.go
```diff
@@ -35,11 +40,32 @@ func DTCompatible() []string {
...
+func DTRevision() uint32 {
+	mu.Lock()
+	defer mu.Unlock()
+
+	if dtRevision != 0 {
+		return dtRevision
+	}
+	b, err := ioutil.ReadFile("/proc/device-tree/system/linux,revision")
+	if err != nil {
+		return 0
```
maruel, line 54: The drawback is that the failure mode will continuously read the file and fail. For example the file doesn't exist on my desktop, so it means DTRevision() will try to open the file every time the function is called.

The two other functions in this file don't have this problem, so I'd prefer to have this fixed.

stapelberg: Done.

host/distro/devtree.go
```diff
@@ -35,11 +40,34 @@ func DTCompatible() []string {
...
+// from the Linux device tree, or 0 if the file is missing or malformed.
+func DTRevision() uint32 {
+	mu.Lock()
+	defer mu.Unlock()
+
+	if dtRevisionRead {
+		return dtRevision
+	}
+	dtRevisionRead = true
+	b, err := ioutil.ReadFile("/proc/device-tree/system/linux,revision")
```
maruel, line 53: Thinking about it more, this is simpler. ReadFile() won't return b if an error occurred, so there's no point in checking the error.

```
if b, _ := ioutil.ReadFile("/proc/device-tree/system/linux,revision"); len(b) >= 4 {
  dtRevision = binary.BigEndian.Uint32(b[:4])
}
return dtRevision
```

stapelberg: Done.

maruel:
gohci

maruel:
Thanks!

PR #440, opened 2020-04-16
bcm283x: Fixes #226: unrolling DMA controlBlocks

### LINKED ISSUE #226: bcm283x: gpio.PWM() Interfering with gpio.Out()

bezineb5:
The DMA controlBlocks are unrolled and the parameter txLen is set to a fixed value.

### DIFF
--- a/host/bcm283x/dma.go
+++ b/host/bcm283x/dma.go
@@ -412,6 +412,8 @@ type controlBlock struct {
 	reserved [2]uint32 // 0x18+0x1C
 }
 
+const controlBlockSize uint32 = 32
+
 // initBlock initializes a controlBlock for any valid DMA operation.
 //
 // l is in bytes, not in words.
@@ -713,12 +715,12 @@ func dmaWritePWMFIFO() (*dmaChannel, *videocore.Mem, error) {
 	if drvDMA.dmaMemory == nil {
 		return nil, nil, errors.New("bcm283x-dma is not initialized; try running as root?")
 	}
-	cb, buf, err := allocateCB(32 + 4) // CB + data
+	cb, buf, err := allocateCB(int(controlBlockSize + 4)) // CB + data
 	if err != nil {
 		return nil, nil, err
 	}
 	u := buf.Uint32()
-	offsetBytes := uint32(32)
+	offsetBytes := controlBlockSize
 	u[offsetBytes/4] = 0x0
 	physBuf := uint32(buf.PhysAddr())
 	physBit := physBuf + offsetBytes
@@ -744,14 +746,12 @@ func startPWMbyDMA(p *Pin, rng, data uint32) (*dmaChannel, *videocore.Mem, error
 		return nil, nil, errors.New("bcm283x-dma is not initialized; try running as root?")
 	}
 
-	// Hardcoded len(controlBlock) == 32
-	cbBytes := uint32(32)
-	cb, buf, err := allocateCB(int(rng*cbBytes + 4)) // rng CBs + mask
+	cb, buf, err := allocateCB(int(rng*controlBlockSize + 4)) // rng CBs + mask
 	if err != nil {
 		return nil, nil, err
 	}
 	u := buf.Uint32()
-	offsetBytes := cbBytes * rng
+	offsetBytes := controlBlockSize * rng
 	u[offsetBytes/4] = uint32(1) << uint(p.number&31)
 	physBuf := uint32(buf.PhysAddr())
 	physBit := physBuf + offsetBytes
@@ -766,7 +766,7 @@ func startPWMbyDMA(p *Pin, rng, data uint32) (*dmaChannel, *videocore.Mem, error
 			_ = buf.Close()
 			return nil, nil, err
 		}
-		ptr += cbBytes
+		ptr += controlBlockSize
 		cb[i].nextCB = ptr
 	}
 	// Low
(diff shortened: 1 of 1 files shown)


host/bcm283x/dma.go
```diff
@@ -743,13 +743,15 @@ func startPWMbyDMA(p *Pin, rng, data uint32) (*dmaChannel, *videocore.Mem, error
 	if drvDMA.dmaMemory == nil {
 		return nil, nil, errors.New("bcm283x-dma is not initialized; try running as root?")
 	}
-	cb, buf, err := allocateCB(2*32 + 4) // 2 CBs + mask
+
+	// Hardcoded len(controlBlock) == 32
+	cbBytes := uint32(32)
```
maruel, line 748: can you make it a const? No idea why I hadn't done it before.

host/bcm283x/dma.go
```diff
@@ -758,20 +760,30 @@ func startPWMbyDMA(p *Pin, rng, data uint32) (*dmaChannel, *videocore.Mem, error
 		drvGPIO.gpioBaseAddr + 0x1C + 4*uint32(p.number/32), // set
 	}
 	// High
-	if err := cb[0].initBlock(physBit, dest[1], data*4, false, true, false, false, dmaPWM); err != nil {
-		_ = buf.Close()
-		return nil, nil, err
+	ptr := physBuf
+	for i := uint32(0); i < data; i++ {
```
maruel, line 764: My only worry is that this could be fairly large in egregious cases. Can you calculate what's the worst case?

For example if PWM(DutyHalf, 1Hz) is called, rng would be huge.

bezineb5: With:
drvDMA.pwmDMAFreq = 200 * physic.KiloHertz
freq = 1 * physic.Hertz

We have:
rng := uint64(drvDMA.pwmDMAFreq / freq)
rng = 200 000

And we allocate:
rng*controlBlockSize + 4
200000 * 32 + 4 = 6400004
Rounded to 4KB:
6402048 bytes => 6252 kiB or around 6MiB

So, to progress on that question:
* Is it too much or not (I have frankly no idea)? Maybe I can return an error if it's too much.
* Do you know what kind of device would require such low-frequency PWM signal? I know servos are in the range 50Hz - 200Hz, more than 100kHz for neopixel LEDs, but there's probably plenty of other devices I don't know.
* I like the cleanness of the solution of using only 2 control blocks, but I don't explain why it's unstable with multiple DMA channels. I unfortunately couldn't find much documentation on this. However, shouldn't the source buffer as large as this size?

host/bcm283x/dma.go
bezineb5: Done

host/bcm283x/dma.go
maruel: It's a significant amount, as it's allocated from the GPU memory. The default amount available is 64MiB as per https://www.raspberrypi.org/documentation/configuration/config-txt/memory.md.

1Hz is borderline egregious; 50Hz (like for servos) is 125Kb, which is tolerable. Can you just add a comment about the trade off being made here?

maruel: I'd say as documentation for startPWMbyDMA. Technically the CL is good as is, so I can commit and the comment can be added as a follow up, since you already waited a fair amount to get this in.

maruel:
gohci

host/bcm283x/dma.go
```diff
@@ -412,6 +412,8 @@ type controlBlock struct {
 	reserved [2]uint32 // 0x18+0x1C
 }
 
+const controlBlockSize uint32 = 32
```
maruel, line 415: I'd recommend removing the uint32 type, as then it'll be used as the correct type in context, which will remove the need for casting as you did at line 901.

Thanks for adding one, using 32 everywhere was a bit dumb.

maruel: well there's offsetBytes that would be incorrect. So you'd save two casts but add a new one. I'd still vote for removing since it's customary in Go.

PR #442, opened 2020-04-24
mfrc522: Add support for 7-bytes UUID

System-Glitch:
- Added support for 7-bytes UUID.
- Added a method to only read a card's UUID.
~~- Added a `frequency` parameter to the `NewSPI()` function to let users tweak SPI frequency. The default 10 MHz wasn't working really well for me, until I changed to 100 KHz. This change doesn't break backwards compatibility.~~


System-Glitch:
@googlebot I signed it!

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -123,6 +134,20 @@ func (r *Dev) SetAntennaGain(gain int) error {
 	return nil
 }
 
+// ReadUID  reads the card UID with IRQ event timeout.
```
maruel, line 137: `s/  / /` i.e. remove double-space

experimental/devices/mfrc522/read_uid_example_test.go
```diff
@@ -0,0 +1,88 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
```
maruel, line 1: This content should go in example_test.go.

experimental/devices/mfrc522/read_uid_example_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"log"
+	"time"
+
+	"periph.io/x/periph/conn/spi/spireg"
+	"periph.io/x/periph/experimental/devices/mfrc522"
+	"periph.io/x/periph/host"
+	"periph.io/x/periph/host/rpi"
+)
+
+func ReadUIDExample() {
```
maruel, line 18: To appear in godoc under the right method, you want `ExampleDev_ReadUID` IIRC.

experimental/devices/mfrc522/commands/low_level.go
```diff
@@ -35,14 +35,24 @@ type AuthStatus byte
...
+	return NewLowLevelSPIFrequency(spiPort, resetPin, irqPin, 10*physic.MegaHertz)
+}
+
+// NewLowLevelSPIFrequency creates and initializes the RFID card reader attached to SPI.
+//
+//  spiPort - the SPI device to use.
+//  resetPin - reset GPIO pin.
+//  irqPin - irq GPIO pin.
+//  frequency - the SPI frequency
+func NewLowLevelSPIFrequency(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn, frequency physic.Frequency) (*LowLevel, error) {
```
maruel, line 51: Technically, if the user wants a lower frequency that the maximum supported by the device, they should use instead spi.PortCloser.LimitSpeed(). I agree it's not obvious (thus I'm open to improved examples) but it's better than adding similar code to every single driver.

https://periph.io/x/periph/conn/spi#PortCloser

experimental/devices/mfrc522/mfrc522.go
```diff
@@ -71,6 +72,16 @@ func noop() {}
...
+	return NewSPIFrequency(spiPort, resetPin, irqPin, 10*physic.MegaHertz, configs...)
+}
+
+// NewSPI creates and initializes the RFID card reader attached to SPI.
+//
+//  spiPort     the SPI device to use.
+//  resetPin    reset GPIO pin.
+//  irqPin      irq GPIO pin.
+//  frequency   the SPI frequency
+func NewSPIFrequency(spiPort spi.Port, resetPin gpio.PinOut, irqPin gpio.PinIn, frequency physic.Frequency, configs ...configF) (*Dev, error) {
```
maruel, line 84: Same here.

experimental/devices/mfrc522/commands/low_level.go
System-Glitch:
```go
type LowLevel struct {
	resetPin    gpio.PinOut
	irqPin      gpio.PinIn
	spiDev      spi.Conn
	antennaGain int
	stop        chan struct{}
}
```
`spiDev` is unexported, so it's impossible to do that. Maybe we could export it and remove the `NewSPIFrequency` method?

maruel: Here's an example for a client: https://github.com/google/periph/blob/master/cmd/apa102/main.go#L154

experimental/devices/mfrc522/example_test.go
```diff
@@ -102,3 +102,75 @@ func Example() {
...
 	}
 }
+
+func ExampleDev_ReadUID() {
+	// Make sure periph is initialized.
+	if _, err := host.Init(); err != nil {
+		log.Fatal(err)
+	}
+
+	// Using SPI as an example. See package ./spi/spireg for more details.
```
maruel, line 112: ./spi/spireg here is a bit confusing.

experimental/devices/mfrc522/commands/low_level.go
System-Glitch: Thanks, I made it work using `LimitSpeed`. I didn't know about that method. I will remove the two new functions.

experimental/devices/mfrc522/example_test.go
```diff
@@ -23,7 +23,7 @@ func Example() {
 		log.Fatal(err)
 	}
 
-	// Using SPI as an example. See package ./spi/spireg for more details.
+	// Using SPI as an example. See package "periph.io/x/periph/conn/spi/spireg" for more details.
 	p, err := spireg.Open("")
 	if err != nil {
 		log.Fatal(err)
```
maruel, line 29: Would it help if you added a LimitSpeed() call here?

maruel:
gohci

experimental/devices/mfrc522/example_test.go
System-Glitch: Yes that's what I did in my local test but I don't think I should add it to he example because I think the frequency issue is related to the card I'm using. I have a regular Mifare Classic and it work perfectly with the default 10 Mhz frequency. However, with my Desfire ev1, it doesn't work without lowering the frequency to 100 Khz.

maruel: You could add something like this?

```
// If the device is misbehaving, you can try limiting the communication speed:
if err := p.LimitSpeed(1*physics.MegaHertz); err != nil {
  log.Fatal(err)
}
```

System-Glitch: Good idea, I will add that later.

PR #448, opened 2020-07-26
netlink/onewire: fix error handling when reading from the netlink socket

mhabb:
When responding to a sequence of 1-wire commands, the kernel driver
individually acks every single one. It is important to keep reading all
of these messages in case of an error in order to flush the netlink
socket for future commands.

### DIFF
--- a/experimental/host/netlink/onewire.go
+++ b/experimental/host/netlink/onewire.go
@@ -281,32 +281,30 @@ func (ws *w1Socket) sendAndRecv(seq uint32, m *w1Msg) ([]byte, error) {
 		return nil, fmt.Errorf("failed to send W1 message: %v", err)
 	}
 
-	var (
-		data []byte
-		errs []error
-	)
+	var data []byte
+	var err1 error
 
 	// Read responses and acks for all commands. It is important to keep
 	// reading in case of an error to flush the netlink socket for future
 	// commands.
-	for i, cmd := range m.cmds {
+	for _, cmd := range m.cmds {
 		if cmd.wantResponse {
 			d, err := ws.recvCmd(seq, seq+1, m.typ, cmd.typ)
-			if err != nil {
-				errs = append(errs, fmt.Errorf("payload %d: %v", i, err))
+			if err1 == nil {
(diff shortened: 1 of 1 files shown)


experimental/host/netlink/onewire.go
```diff
@@ -281,23 +281,34 @@ func (ws *w1Socket) sendAndRecv(seq uint32, m *w1Msg) ([]byte, error) {
 		return nil, fmt.Errorf("failed to send W1 message: %v", err)
 	}
 
-	var data []byte
-	for _, cmd := range m.cmds {
+	var (
```
maruel, line 284: It's shorter to do:

```
var data []byte
var errs []error
```

experimental/host/netlink/onewire.go
```diff
@@ -281,23 +281,34 @@ func (ws *w1Socket) sendAndRecv(seq uint32, m *w1Msg) ([]byte, error) {
...
 		// Every command is ack'ed with a separate message, including commands
-		// for which a response has already been returned.
+		// for which a response has already been received.
 		if _, err := ws.recvCmd(seq, 0, m.typ, cmd.typ); err != nil {
-			return nil, fmt.Errorf("failed to receive command ack message: %v", err)
+			errs = append(errs, fmt.Errorf("ack %d: %v", i, err))
 		}
 	}
 
+	if len(errs) > 0 {
```
maruel, line 308: Once an error was received, I don't think the next errors are useful, and in practice may be overwhelming. I think just saving the first error is likely going to be enough.

You can do by changing line 294 to:
```
d, err1 :=  ws.recvCmd(seq, seq+1, m.typ, cmd.typ)
if err == nil {
  err = err1
}
```

same for line 303~304.

experimental/host/netlink/onewire.go
mhabb: Done.

experimental/host/netlink/onewire.go
mhabb: Done.

maruel:
gohci

PR #452, opened 2020-11-28
package mcp23xxx - Fixing according datasheet

honzakadlec:
Notes

Please prefix the issue title with the primary package affected. For example, if
this PR fixes periph.io/x/periph/host/sysfs, prefix the PR title with "sysfs:".

Please add new drivers under "experimental". Wonder what it takes to promote a
driver as "stable"? See https://periph.io/project/#driver-lifetime-management
A stable driver requires the smallest API surface, good unit test code coverage,
good documentation and a page in
https://github.com/periph/website/tree/master/site/content/device so it can show
up at https://periph.io/device.

Mention the issue number it fixes or add the details of the changes if it
doesn't have a specific issue.

Examples:

Fixes #12345
Helps with #12345 but doesn't not completely fix it.

### DIFF
--- a/experimental/devices/mcp23xxx/mcp23xxx.go
+++ b/experimental/devices/mcp23xxx/mcp23xxx.go
@@ -152,7 +152,7 @@ func mcp23x178ports(devicename string, ra registerAccess) []port {
 
 		// interrupt handling registers
 		gpinten:          ra.define(0x05),
-		intcon:           ra.define(0x0B),
+		intcon:           ra.define(0x09),
 		intf:             ra.define(0x0F),
 		intcap:           ra.define(0x11),
 		supportInterrupt: true,


honzakadlec:
@googlebot I signed it!

maruel:
Hi! Thanks for the change. It seems the commit is associated with the user "jan". Can you confirm? That's why the CLA check fail. I'd recommend a git commit --amend then git push -f.

Agreed on the stable migration, I plan to redo all that for next version but sadly have made no progress in the past months.

Do you have a link to a datasheet? There's none in the current package documentation.

maruel:
Err ignore the comment about stable. :)

maruel:
I looked at https://ww1.microchip.com/downloads/en/DeviceDoc/20001952C.pdf but failed to quickly spot the value, which page should I look at?

balazsgrill:
> I looked at https://ww1.microchip.com/downloads/en/DeviceDoc/20001952C.pdf but failed to quickly spot the value, which page should I look at?

@maruel page 12, Table 3-1, and page 17 table 3-5 (bank select is zero on reset). This is an error in my original contribution, thanks @honzakadlec for catching it!

maruel:
Pushed 19c7e6b260b10719669122fa72fcbd1194e2f0b4 to replace this, since the CLA check was failing. Thanks!

PR #454, opened 2020-12-28
host/pine64: improve detection

flokli:
Read the model name from /proc/device-tree/model (via distro.DTModel())
instead of probing for /boot/pine64.dtb to be present.

There's various distros out there that don't have a /boot/pine64.dtb,
and /proc/device-tree/model is a much nicer way to do this.

### DIFF
--- a/host/pine64/pine64.go
+++ b/host/pine64/pine64.go
@@ -6,12 +6,13 @@ package pine64
 
 import (
 	"errors"
-	"os"
+	"strings"
 
 	"periph.io/x/periph"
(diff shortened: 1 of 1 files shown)


flokli:
CI failure seems unrelated:

```
FAIL	periph.io/x/periph/experimental/devices/tlv493d	0.146s
```

maruel:
Hi! Thanks for the contribution! Sadly my pine64 is dead and this code predated when distro was added. 
I'll look at the CI failures separately.

flokli:
Thanks!

PR #455, opened 2020-12-29
nrzled: Don't set NoCS

flokli:
Some SPI drivers don't support setting SPI_NO_CS, and fail entirely
configuring the port in that case.

In the case of nrzled, on the host side, we shouldn't really need to
configure if there's SPI CS happening or not.

If the nrzleds are just wired to the MOSI port directly, that's totally
fine - and if there's some proper CS going on, actually not setting
SPI_NO_CS would suddenly allow selecting different nrzled stripes
sequentially.

### DIFF
--- a/experimental/devices/nrzled/nrzled.go
+++ b/experimental/devices/nrzled/nrzled.go
@@ -90,7 +90,7 @@ func NewSPI(p spi.Port, opts *Opts) (*Dev, error) {
 			return nil, errors.New("spi port buffer is too short for the specified number of pixels")
 		}
 	}
-	c, err := p.Connect(spiFreq, spi.Mode3|spi.NoCS, 8)
+	c, err := p.Connect(spiFreq, spi.Mode3, 8)
 	if err != nil {
 		return nil, err
 	}


maruel:
Can you expand on which host have trouble with NoCS? I'd prefer to disable the flag in the sysfs driver instead on known bad configuration than not using it anywhere.

flokli:
That was on a PINE64, spi0 (`allwinner,sun8i-h3-spi`), 5.10.2 mainline kernel.

`nrzled.NewSPI` unconditionally sets the `SPI_NO_CS` flag, which caused a 

```
spidev setup: unsupported mode bits 40
```

When trying to run the `nrzled` command.

Why do we want to set NoCS inside periph? If the light strip doesn't "listen" on the state of a potential CS pin, there's no harm in still setting it?

I could imagine a usecase where multiple WS2812 strips are connected to the same SPI master, there's multiple `/dev/spidev0.${chipselect}` nodes, periph maintains multiple `spi.Port` instances, and with some simple chip selection circuit, both share the same MOSI port. You can't write concurrently, true, but that might be fine.

maruel:
Ok I'd still prefer for the sysfs driver to be updated to retry without the flag when this happens.

Separately, would you mind joining the slack channel? I'm currently in the process of migrating the project to github.com/periph/(conn|cmd|devices|host) and in the meantime it means I have to cherry pick each change manually. I'll try to have it done in the coming days.

flokli:
> Ok I'd still prefer for the sysfs driver to be updated to retry without the flag when this happens.

Hmm… unsure. We shouldn't silently ignore flags passed by users when configuring the SPI port.

So what's the reasoning behind nrzled setting NoCS in first place? Why would it be fine to now ignore it in some lower layer, but not fine to not set it?

maruel:
Because many other device drivers could want to say "I don't need a CS line" and a (non-sysfs) implementation could say "cool then you don't have to configure one" and save a pin.

What you are asking for is essentially to have no device driver use this flag ever. It's a bug in the sysfs driver implementation in pine64, so I'd prefer this to be worked around there specifically.

flokli:
> Because many other device drivers could want to say "I don't need a CS line" and a (non-sysfs) implementation could say "cool then you don't have to configure one" and save a pin.

If there's devices out there that dynamically allocate pins, and you can reuse others setting NoCS that makes sense. Thanks! It should probably be added in a comment, though ;-)

> What you are asking for is essentially to have no device driver use this flag ever. It's a bug in the sysfs driver implementation in pine64, so I'd prefer this to be worked around there specifically.

No, that's not what I meant, I'm only worried about silently "dropping" a flag in the sysfs spi implementation that was set by the user while initializing the port, especially if it's named the same way as an option where the kernel *does* fail if it's not able to set.

Are you aware of use cases where people would want this to fail if it couldn't be set? Should the periph flag probably be named differently? (`NeedCS` maybe?)

On the other hand, `nrzled.NewSPI(p spi.Port, opts *Opts)` also has no idea about the cabling on specific hardware, so configuring it there (with linux-specific magic values for modes!) probably also shouldn't happen.

What about giving `host/sysfs.NewSPI` (and other implementations) a slightly different list of arguments, where users could specify they don't care about the chip select pin being set (in an option that isn't named exactly like the kernel option)?

`NewSPI` is usually called by users (or code more familiar with the hardware), so might know about the exact cabling, and nicely encapsulates that detail away from libraries, who really just want to take a `conn/spi.Port`, configure frequency and write data to it.

maruel:
I'll close here as all the PRs should now be done on github.com/periph.

flokli:
For posterity - this ended up in https://github.com/periph/devices/pull/4/files.

PR #456, opened 2020-12-29
nrzled/rasterSPI: fix ordering

flokli:
This is supposed to be sent out in GRB order (and the comment below also
below claims it), but the code (and corresponding unit tests) did output
RGB order.


maruel:
Thanks!

PR #457, opened 2021-01-07
added support for EPD2in13v2

nullsumme:
Added support for the EPD2in13v2 display.

LUT values taken from:
https://github.com/waveshare/e-Paper/blob/master/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd2in13_V2.py


nullsumme:
@googlebot I signed it!

maruel:
Hi! Thank you so much for the change. I'm trying to only scope improvements to the new version of the code that is Go modules compatible.

The announcement is https://periph.io/news/2020/a_new_start/
This would go in https://periph.io/news/2020/a_new_start/

As a nice side effect, the "experimental" bit is removed.

nullsumme:
Ah, great news. So I will just make a new pull request to periph/devices, right?

maruel:
yep!

